PR #10, opened 2015-02-27
Ruby: Updated for Ruby 1.9+

elektronaut:
Hey! 

This patch makes mal run on modern rubies. I've tested it on 2.1 and 2.2, but it should work on anything from 1.9 on. It also fixes a crash if the history file doesn't exist.

Ruby 1.8 will no longer work. Hopefully that won't be a problem, 1.8.7 is long retired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ cd racket
 ./stepX_YYY.rb
 ```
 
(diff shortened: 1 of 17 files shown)


kanaka:
Merged. Thanks @elektronaut

PR #24, opened 2015-03-05
Adds livescript implementation

alexkalderimis:
This is an implementation in [LiveScript](http://livescript.net/), a compile-to-js language.


kanaka:
You need to update the top level README (the count and section explaining how to use it). Then I'll merge it in.

BTW, it would be helpful for people learning LiveScript and comparing it to other mal implementations if the code in logic.ls was actually in the steps themselves. I know it's repeated code, but it would show more clearly how each step incrementally builds up the language. That's one of the goals of mal is to have parallel languages with a similar structure and flow that can be easily compared. I won't reject it for that, but I would like to see that if you're willing.

alexkalderimis:
Sure, I'll totally do those things

alexkalderimis:
@kanaka - Can you review this again - it is now more conformant, passes up to test-suite 9 and has an incremental development in each step, as requested.

kanaka:
@alexkalderimis cool! It's looking very close. Just a couple more things:

- you should have a stepA_mal.ls that sets the *host-language* variable and prints a language string "Mal [livescript]" before the prompt. That's also where interop and other misc stuff would go.

- get stepA_mal.ls to self-host. In other words, you should be able to run the mal implementation of mal, using your livescript implementation of mal like this:
```
node_modules/LiveScript/bin/lsc stepA_mal.ls ../mal/stepA_mal.mal
```

More generally, all the self-hosting tests need to pass:
```
make test MAL_IMPL=livescript test^mal
```

I think the main thing you are missing is a working readline call. You can copy the js/node_readline.js if you like (it uses system FFI to invoke readline). Or you can figure out a way to use node repl/readline to do it (but the readline core function is expecting to be synchronous so it will be tricky).

At some point I'll have time to finish up step9 and stepA in the guide and describe the above details clear because right now you sort of have to infer it from other things.

kanaka:
@alexkalderimis I was curious, so I did a couple of small changes just to get readline and initial self-hosting to work.

With the changes below, `node_modules/LiveScript/bin/lsc stepA_mal.ls ../mal/step1_read_print.mal` and `node_modules/LiveScript/bin/lsc stepA_mal.ls ../mal/step2_eval.mal` work but running it on step3 gives this (when I do something simple like `(+ 2 3)`):
```
mal-user> (+ 2 3)
Uncaught exception: Error calling function: Error calling function: Cannot call nil
```

Anyways, here is the patch to get that far.

```patch
diff --git a/livescript/core.ls b/livescript/core.ls
index 941aa60..97215e0 100644
--- a/livescript/core.ls
+++ b/livescript/core.ls
@@ -12,6 +12,8 @@ require! {
         Builtin, Lambda, MalList, MalVec, MalMap, UserError, JSObject
     }
 }
+readlineSync = require('readline-sync')
+node_readline = require('./node_readline')
 
 export ns = {}
 
@@ -28,6 +30,8 @@ ns['with-meta'] = new Builtin ([val, metadata]) ->
 
 ns['atom'] = new Builtin ([value]) -> atom value
 
+ns['atom?'] = new Builtin ([x]) -> bool x.type is \ATOM
+
 ns.list = new Builtin (elems) -> new MalList elems
 
 ns['list?'] = new Builtin ([x]) -> bool x.type is \LIST
@@ -264,3 +268,6 @@ ns['-readline'] = new Builtin ([prompt]) -> new Promise (resolve, reject) ->
         resolve string result
         rl.close()
     rl.on \close, -> (resolve NIL)
+
+ns['readline'] = new Builtin ([prompt]:args) ->
+    string node_readline.readline prompt.value
diff --git a/livescript/package.json b/livescript/package.json
index 2d522aa..7dae9a7 100644
--- a/livescript/package.json
+++ b/livescript/package.json
@@ -8,6 +8,7 @@
     "es6-promise": "^2.0.1",
     "prelude-ls": "^1.1.1",
     "q-io": "^1.11.6",
+    "ffi": "1.2.x",
     "readline-sync": "^0.7.5"
   },
   "config": {
diff --git a/tests/step9_try.mal b/tests/step9_try.mal
index 01fce99..2c4ac6e 100644
--- a/tests/step9_try.mal
+++ b/tests/step9_try.mal
@@ -318,6 +318,9 @@
 (def! a (atom 2))
 ;=>(atom 2)
 
+(atom? a)
+;=>true
+
 ;;;(type a)
 ;;;;=>"atom"
```

alexkalderimis:
Thanks for the attention @kanaka! I haven't had much time to look at this the last couple of days, but I will attend to this very shortly.

kanaka:
@alexkalderimis so, a bit more debugging reveals this:

```
(get (assoc (assoc {"a" 1 } "b" 2) "c" 3) "a") ; broken
(get (assoc (assoc {"a" 1 } "b" 2) "c" 3) "b") ; works
(get (assoc {"a" 1 } "b" 2 "c" 3) "a") ; works
```

I've added a step9 test to upstream to catch that case. It seems like there is something wrong with the implementation of assoc. I don't see anything obvious from a brief look at core.ls and types.ls but thought that might help in debuggin.

kanaka:
@alexkalderimis no pressure, but just wanted to let you know that https://github.com/timculver/mal is closing in fast with a C++ implementation ;-)

alexkalderimis:
That's very helpful. I will take a look at this asap (on a business trip at
the moment, so it will not be immediate I am afraid)

kanaka:
@alexkalderimis wow, it's been three weeks already (time flies). You still going to be able to get back to this? Anything I can help with?

kanaka:
@alexkalderimis I'm going to go ahead and close this pull since LiveScript was finished in a separate pull: https://github.com/kanaka/mal/pull/262 BTW, this doesn't preclude your implementation from getting into the tree if it's uniquely different in some way (you would need to finish it up and move it to a unique directory). Or perhaps a new implementation in a new language.

PR #26, opened 2015-03-05
Update Rust code to the current Rust nightly

alexcrichton:
Lots of changes! Hopefully everything is still semantically equivalent (tests
still pass)

This is an alternative to #23

Note that I forked rust-pcre to https://github.com/alexcrichton/rust-pcre and updated it as well. My current compiler version is `rustc 1.0.0-nightly (fed12499e 2015-03-03) (built 2015-03-04)`

PR #28, opened 2015-03-06
Replace `pcre` with Rust-implemented regex crate (based on RE2).

BurntSushi:
I fixed the bug in Rust's `regex` library and things seem to be working great now. Even through `regex` is based on RE2, I didn't have to change your regex at all. :-)

I ran this to test:

```bash
make perf^rust
make MAL_IMPL=rust test^mal
```

The first passes and the second passes all tests except for 1, but it looks like the same failing test from #26.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -8,6 +8,3 @@ authors = [ "Your name <you@example.com>" ]
 time = "0.1"
 regex = "0.1"
 libc = "0.1"
-
-[dependencies.pcre]
-git = "https://github.com/alexcrichton/rust-pcre"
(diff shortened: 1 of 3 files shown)


kanaka:
@BurntSushi I merged it in and tested it but I was getting reader errors. I reverted the change for now since it wasn't working for me. I might have some more time tomorrow to take a look.

kanaka:
@BurntSushi hmm, okay, a fresh checkout of your repo worked. I think mine was checking out a stale version of regex (0.1.17). Probably I forgot to run cargo update. Let me try again.

kanaka:
@BurntSushi yep, it was just me forgetting to run cargo update and getting a stale build of regex. Silly me. I reverted the revert (first time I've done that in git I think) and pushed. We should be all good. 

Thanks for that!

BurntSushi:
Aha! I completely forgot to remind you to run `cargo update`. It has tripped me up more than once, so don't feel bad! :-) Thanks!

PR #31, opened 2015-03-06
Compatibility with Python 3

microamp:
Two things:
* a syntax error in 'step0_repl.py' (see below) fixed
```bash
$ python --version
Python 3.4.3
$ cd python/
$ python step0_repl.py
  File "step0_repl.py", line 14
    exec compile(ast, '', 'single') in globals()
               ^
SyntaxError: invalid syntax
```
* the test script, 'runtest.py', now made compatible with Python 3 (initially getting errors about ``print`` missing parentheses, and ``bytes``/``str`` decoding/encoding)

Now, I can run Mal tests in both Python 2 and 3. Also tested with Ruby and Clojure.

Thanks for reviewing.

### DIFF
--- a/python/step0_repl.py
+++ b/python/step0_repl.py
@@ -11,7 +11,7 @@ def EVAL(ast, env):
     try:
         return eval(ast)
     except SyntaxError:
-        exec compile(ast, '', 'single') in globals()
+        exec(compile(ast, '', 'single'), globals())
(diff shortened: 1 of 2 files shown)


kanaka:
Thanks! I occasionally had been testing the python implementation with python3 but not runtest.py

PR #34, opened 2015-03-08
guide: incorrect filename fixed

microamp:
another minor typo in the guide; `step2_env.qx` should've been `step3_env.qx`


kanaka:
Yep, good catch.

PR #39, opened 2015-03-11
perl: Support newer Perl versions (e.g. v5.20.0)

vi:
There were error messages:

keys on reference is experimental at perl/printer.pm line 25.
Compilation failed in require at perl/step1_read_print.pl line 10.
BEGIN failed--compilation aborted at perl/step1_read_print.pl line 10.

---

Error: Modification of a read-only value attempted at perl/types.pm line 111, <FIN> line 1.

### DIFF
--- a/perl/printer.pm
+++ b/perl/printer.pm
@@ -22,7 +22,7 @@ sub _pr_str {
         }
         when(/^HashMap/) {
             my @elems = ();
-            foreach my $key (keys $obj->{val}) {
+            foreach my $key (keys %{$obj->{val}}) {
                 push(@elems, _pr_str(String->new($key), $_r));
                 push(@elems, _pr_str($obj->{val}->{$key}, $_r));
             }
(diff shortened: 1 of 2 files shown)


kanaka:
@vi I pushed a Dockerfile to tests/docker/Dockerfile. It installs all the tools/languages in an Ubuntu Utopic image. I discovered a few issues in a couple of languages so I made those fixes and that included getting the perl implementation working with perl 5.20. I just pushed that (there were a couple of other things I had to change in addition to what you listed here).

Note about the Dockerfile. All the tools together result in an image that is over 3GB virtual size and takes me 80 minutes to rebuild from scratch. So if you're just wanting to use it to test a couple of languages, I suggest pruning it down.

vi:
Yes, the current master version works in my NixOS environment too.

kanaka:
@vi I just update the README to provide information about building/running the docker test environment.

PR #41, opened 2015-03-14
Add outer tests to step3_env.mal

epylar:
I noticed that when I implemented step 3 with an environment that didn't have an "outer" reference, all the tests passed.  Here are some simple tests that check if outer is functioning.


kanaka:
You're right! Thanks! Merged.

PR #42, opened 2015-03-15
Fix extension for Racket in README.

stamourv:


kanaka:
I guess you can tell which section of the README I copy and pasted for that one :-) Thanks for the catch. Merged.

PR #43, opened 2015-03-16
Step 9 apply tests test user fns as well as core

sdt:
I had a bug in my apply core function. It worked for core functions, but
not for user-defined functions.

This bug escaped direct testing, but caused problems in self-host
testing.

Here's a couple of simple test cases that catch those errors.


kanaka:
Yes, those are important tests, thanks.

PR #44, opened 2015-03-16
Remove extra ) from step9 test snippet

sdt:
My reader complains if there is extra input after the top-level
read_form, so this test was inadvertently failing for me.


kanaka:
Heh, that's actually the second case of an extraneous trailing paren. Apparently somebody needs to make a mal syntax and indenting support for vim and emacs :-)

PR #45, opened 2015-03-17
Test case & fix for typo in Perl _clone implementation

sdt:
I stumbled over this typo while looking through the Perl implementation.

Thought I might as well add a test case to catch similar errors.


kanaka:
Yes, I'm surprised that I didn't have something that caught that already. Thanks!

PR #46, opened 2015-03-20
C: fix compilation errors and warnings for the official toolchain of OSX...

neutralevil:
....


kanaka:
@neutralevil thanks!

I made a small cleanup to remove the glib includes in step0 and step1 which I don't think are necessary. Can you verify that change (I don't have a OS X system easily accessible).

neutralevil:
@kanaka I've verified the updated code, and the glib includes are unnecessary. I added glib.h to eliminate a warning complaining that implicit declaration of function 'g_print' is invalid in C99, as C99 is the default c standard for clang.

kanaka:
@neutralevil ah yes, you're right, g_print is referenced in there. Part of the reason I assumed it was unnecessary was because you only added it to step0 and step1. The other steps are pulling in glib.h implicitly via another include. So technically, the include should be in all the step files. If you make that change to all the steps and send me another PR, I'll merge it.

PR #47, opened 2015-03-22
Step 9 test cases for apply on vectors

sdt:
While it still needs a little work, my C++ implementation now passes all[1] of the self-hosting tests (!)

The only bug I had between passing stepA, and then passing the self-hosting tests was that my apply implementation insisted on a list, and needed to be relaxed to accept any sequence.

Here's a couple of (apply f vector) tests which catch that bug.

[1] All but one. `(read-string ";; comment")` in the step 6 test gives me a strange python IO error from runtest.py. I need to look further into that one - it works fine if I type it in manually.

### DIFF
--- a/tests/step9_try.mal
+++ b/tests/step9_try.mal
@@ -56,16 +56,24 @@
 ;=>5
 (apply + 4 (list 5))
 ;=>9
+(apply + 4 [5])
+;=>9
 (apply prn (list 1 2 "3" (list)))
 ; 1 2 "3" ()
 (apply prn 1 2 (list "3" (list)))
 ; 1 2 "3" ()
+(apply prn 1 2 ["3" 4])
+; 1 2 "3" 4
 ;=>nil
 
 ;; Testing apply function with user functions
 (apply (fn* (a b) (+ a b)) (list 2 3))
 ;=>5
 (apply (fn* (a b) (+ a b)) 4 (list 5))
+;=>9
+(apply (fn* (a b) (+ a b)) [2 3])
+;=>5
+(apply (fn* (a b) (+ a b)) 4 [5])
 ;=>9
 
 ;; Testing map function


kanaka:
Sweet! Sounds like you're basically ready to merge. I think the only things remaining are:
- make sure the IO error you're seeing from read-string isn't a big deal. read-string of a non-token (comment or empty line) has slightly different behavior across some of the implementations (some return nil and some raise a blank line exception). What is the IO error that you're seeing
- Move sdt-cpp to cpp and create a branch without sdt-perl (and update the top level Makefile)
- Send me a PR. :-)

Yeah, those are good additional tests. They caught an error in the Racket implementation. I merged and then did a little re-ordering to move those tests down into the optional tests section (vectors are optional types).

sdt:
That'd be great, but there's a couple of other issues I need to take care of before it's ready to merge.

* This doesn't compile on Linux. I use std::regex in the reader, which compiles fine on OS X, but maybe needs a more recent clang/llvm than most Linux distros seem to ship with. A working docker/fig environment should be easy enough to knock up though.
* I think some of the earlier steps got broken by changes in the later ones.
* Sometime around step 7 I went on an ill-conceived constness-crusade, completely forgetting about reset! and swap!. I've now got this mutable object buried down in my malAtom class, thumping away like a telltale heart. I need to take a step back and think about this a little more.
* Various implementation & stylistic quibbles (I seem to switch between camel & snake cases without noticing)

Regarding a rebase, would one commit per step be an appropriate granularity? (maybe even stepN, stepN-deferrables)

I deliberately did this implementation depth-first, which means I'm way off the script in terms of what got implemented when.

kanaka:
Okay, cleaning up all of that first would be good. And a commit per step or step+deferrables is fine.

PR #49, opened 2015-03-24
Initial Swift implementation.

keith-rollin:


kanaka:
Thanks @keith-rollin! I updated the language count in the README to 28 and noted that XCode/OS X is required. Only thing left to do is tweet. Do you have a twitter account? If so, please tweet about it and I'll retweet. If not then I'll just tweet it. Now I look forward to Apple releasing Swift for clang/LLVM generally so that I can try it. Probably won't happen anytime soon :-)

keith-rollin:
> I updated the language count in the README to 28 and noted that XCode/OS X is required.

Ah, thanks for addressing those oversights.

> If not then I'll just tweet it.

Here you are: https://twitter.com/keithrollin/status/580452821105029120

> Now I look forward to Apple releasing Swift for clang/LLVM generally so that I can try it.

Yeah, that would be nice. I know that they want to do it, but I haven't heard that there's a schedule for it.

PR #50, opened 2015-03-27
C++ implementation

sdt:
It took a little longer than expected, but I finally got this into some sort of reasonable shape.

Points of note regarding this implementation:
* All mal values and environments are reference counted and automatically garbage collected.
* All special forms and core functions are fully type checked (this was an unintended side-effect of the implementation).
* One of the goals of this implementation was to avoid switching on the types, instead using the virtual methods of the class hierarchy. I wasn't able to realise this 100%, but came fairly close. I'm not entirely sure if this makes the implementation more or less understandable.
* After completing my original perl implementation, I wanted to see if I understood it well enough to build this one from memory. This implementation was written by referring only to the red text in the guide.md diagrams (and the excellent tests).
* Compilation notes are in README.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ stepA = stepA_mal
 
 EXCLUDE_TESTS += test^bash^step5 # no stack exhaustion or completion
 EXCLUDE_TESTS += test^c^step5    # segfault
+EXCLUDE_TESTS += test^cpp^step5  # completes at 10,000
 EXCLUDE_TESTS += test^cs^step5   # fatal stack overflow fault
 EXCLUDE_TESTS += test^haskell^step5 # test completes
 EXCLUDE_TESTS += test^make^step5 # no TCO capability/step
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ PYTHON = python
 # Settings
 #
 
-IMPLS = bash c clojure coffee cs forth go haskell java js lua make mal \
+IMPLS = bash c clojure coffee cpp cs forth go haskell java js lua make mal \
 	ocaml matlab miniMAL nim perl php ps python r racket ruby rust \
 	scala swift vb
 
@@ -54,6 +54,7 @@ bash_STEP_TO_PROG =    bash/$($(1)).sh
 c_STEP_TO_PROG =       c/$($(1))
 clojure_STEP_TO_PROG = clojure/src/$($(1)).clj
 coffee_STEP_TO_PROG =  coffee/$($(1)).coffee
+cpp_STEP_TO_PROG =     cpp/$($(1))
 cs_STEP_TO_PROG =      cs/$($(1)).exe
 forth_STEP_TO_PROG =   forth/$($(1)).fs
 go_STEP_TO_PROG =      go/$($(1))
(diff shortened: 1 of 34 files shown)


kanaka:
@sdt it's looking good. I checked it out to try it before merge and ran into some test failures with the reader:

- Two spaces at the end of a line causing an exception:
```
user> 7  
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_S_create
Aborted (core dumped)
```

I did a little debugging and it is the call to `m_iter->str(1);` within peek that is causing the issue. I put some debug in there and notice that even though m_iter->size() is 2, m_iter->str(1) still crashes. I thought m_iter->size() should be the number of match groups (submatches), but maybe not.

- Some tokens are not properly split:
```
./step1_read_print
user> @abc
@abc
user> @ abc
(deref abc)
user> ~abc
~abc
user> ~ abc
(unquote abc)
user> ~@abc
~@abc
user> ~@ abc
(splice-unquote abc)
```

Also, by moving LDFLAGS, I was able to build with gcc-4.9:
```
--- a/cpp/Makefile
+++ b/cpp/Makefile
@@ -35,7 +35,7 @@ all: $(TARGETS)
        $(CXX) $(CXXFLAGS) -MM *.cpp > .deps
 
 $(TARGETS): %: %.o libmal.a
-       $(LD) $(LDFLAGS) $^ -o $@
+       $(LD) $^ -o $@ $(LDFLAGS)
 
 libmal.a: $(LIBOBJS)
        $(AR) rcs $@ $^
```

The errors I mentioned above happen regardless of whether I build with clang-3.5 or gcc-4.9 and happen inside and out of the docker container.

sdt:
Ah crap.

Yep - I can repro both the double-space and the token-splitting issues on Linux, but both of these work fine on Yosemite with clang-3.6. I've got a bad feeling that std::regex is still a bit buggy with clang-3.5.

I've also got a 15.05/clang-3.6 image here. I'll have a look and see how that fares.

I may need to go back to the drawing board for the reader.

sdt:
Same problems with 15.04/clang-3.6 unfortunately.

sdt:
I've fixed a few things and this now passes all the tests locally and in the docker container (directly and in self-hosting mode).

* Rewrote the tokeniser to manually iterate over the input using regex_search, rather than automatically with regex_iterator. This seems to fix all the reader issues.
* Removed the reader macro which converted { X } to ( hash-map X ) - this fixes step 2 tests.
* Switched to g++-4.9 as per your suggestion.
* Added required packages into the Dockerfile (I had done this, but it must've fallen through the cracks while rebasing)

kanaka:
Excellent. Builds without warnings and all tests (regular and self-host) pass. Merged.

I pushed a couple of small updates:
- bumped language count to 29 on the main README and added small section explaining build and run process (and linking to language specific README).
- added a stats and stats-lisp target to the cpp Makefile

Thanks for your work on this. This knocks another language off typical top 10 language lists. The only other one that is consistently in the top 10 that mal doesn't have yet is Objective-C. Care to tackle that one next? ;-) Since you've now done the C++ implementation and know both OS X and Linux, you might be the right person to tackle it. No pressure :-)

One of the challenges there is that Linux Objective-C compiler and libraries are a bit dated (basically GNUstep), but with some effort you could probably make an Objective-C version that works in both. Here is a link that might get you started if you're at all interested: http://blog.tlensing.org/2013/02/24/objective-c-on-linux-setting-up-gnustep-clang-llvm-objective-c-2-0-blocks-runtime-gcd-on-ubuntu-12-04/

Anyways, nice work. You're implementation is certainly better quality than my C implementation and is a great addition to the set of implementations.

The only thing left for you to do if you have a twitter account is to tweet about it so that I can retweet (if you don't have an account I can just tweet it directly).

kanaka:
@sdt also, a few other implementors and myself hang out in #mal on freenode if you're interested.

sdt:
That's great news. Very pleased to have contributed to such an excellent project!

Sorry, I don't have a twitter account - do you mind taking care of that?

As far as Objective-C goes, I've never written a line (and have no intention of doing so!)

I've been meaning to pop in on #mal - I've got a whole bunch of dumb questions for ya.

Cheers!

kanaka:
I tweeted it here: https://twitter.com/bus_kanaka/status/582249915264319488

PR #57, opened 2015-04-04
Erlang implementation of Make-a-Lisp

nlfiedler:
Howdy,

This uses Erlang/OTP R17 features (namely maps) so my Dockerfile additions diverge from the norm and build Erlang from source, rather than installing the older package. Also had to install rebar for building (it makes generating the escript executables trivial).

All tests pass including self-hosting.

If I could do it over again, I would have used records for all of the MAL values (e.g. #list{...} instead of {list, ...}). Much too difficult to fix it now though, and since it is finished, nothing would be gained by changing it now.

Thanks so much for this project. It has been terrific fun.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ cs/*.dll
 cs/*.mdb
 clojure/target
 clojure/.lein-repl-history
+erlang/ebin
+erlang/.rebar
+erlang/src/*.beam
 go/step*
 go/mal
 java/target/
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ PYTHON = python
 # Settings
 #
 
-IMPLS = bash c clojure coffee cpp cs factor forth go haskell java \
+IMPLS = bash c clojure coffee cpp cs erlang factor forth go haskell java \
 	julia js lua make mal ocaml matlab miniMAL nim perl php ps \
 	python r racket ruby rust scala swift vb
 
@@ -57,6 +57,7 @@ clojure_STEP_TO_PROG = clojure/src/$($(1)).clj
 coffee_STEP_TO_PROG =  coffee/$($(1)).coffee
 cpp_STEP_TO_PROG =     cpp/$($(1))
 cs_STEP_TO_PROG =      cs/$($(1)).exe
+erlang_STEP_TO_PROG =  erlang/$($(1))
 factor_STEP_TO_PROG =  factor/src/$($(1))/$($(1)).factor
 forth_STEP_TO_PROG =   forth/$($(1)).fs
 go_STEP_TO_PROG =      go/$($(1))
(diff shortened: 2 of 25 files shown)


b0oh:
Hi, I am making mal also using erlang — https://github.com/b0oh/mal/tree/erlang, but little bit different, I don't use a state for environments, I am passing it explicitly thru arguments.
Do we need a few realisations of mal on erlang?

nlfiedler:
Hi Dmitry

Yes, I had tried to pass the environment around as well, but ran into problems in step 5. While in step 4 I was able to pass the tests using an ugly “fallback” environment when invoking (recursive) closures, in step 5 I found I could not pass the tests using that approach. I don’t remember the exact problem, but the only way I could see getting around it was to use a mutable “object”, which in Erlang means a process.

Note that the critical problem was self-referential functions: how does the following closure know it’s own name?

(def! sumdown (fn* (N) (if (> N 0) (+ N (sumdown  (- N 1))) 0)))
      ^        ^
      |        |
      |        +-- closure is defined here, before sumdown is known
      |
      +-- the name of the closure is in this outer environment

Anyway, for supporting the atom type, it will certainly be the case that you need an unchanging reference to an “object” whose internal state changes. Only way to do that in Erlang is with a process. Kinda expensive, but hopefully we don’t see too many atoms in MAL, and I think that is likely the case.

The one thing I would have changed in my implementation would be to use records for the various MAL data types, like list, vector, and map. Passing around {list, List} was fine until I had to add the meta data. It was painful going back through all of the steps and fixing all of the places where a MAL value was constructed, or received in a function clause pattern match, etc.

Have fun with it, I certainly did.

n

b0oh:
Yes, I am trying to solve the same problem now :). Exactly with sumdown test.

b0oh:
@kanaka, what about alternate solution, do you interested in?

kanaka:
@nlfiedler merged! I expanded the README section a bit to add you name as the implementor and also to expand the instructions a bit.

@b0oh since Erlang has builtin support for TCO, you should be able to implement mal by passing the environment in and out (non-mutating) and still have TCO by leveraging Erlang's TCO. And yes, if you take a different approach but still follow the basic structure of the other implementations (e.g. the organization or steps, core, env, etc), then I'm happy to take alternate implementations. Mal is all about learning so alternate implementations that take a different approach fit into that goal.

kanaka:
@nlfiedler only thing left if for you to tweet about it and I'll retweet. :-)

kanaka:
@nlfiedler @b0oh oh, and don't forget to join #mal (freenode) for random chat about mal, Lisps, languages, etc.

nlfiedler:
@kanaka Terrific, thanks for merging. Here's my tweet: https://twitter.com/nlfiedler/status/585100306435964929

@b0oh I would very much encourage you to continue with your implementation. It would be especially neat to see if you can devise another way to deal with the mutable environment issue. Either way, I'm sure I will learn something from studying your work.

b0oh:
@nlfiedler, cool then, I'll try :).

PR #61, opened 2015-04-08
Optimize, refactor, and convert to Swift 1.2

keith-rollin:


kanaka:
I assume self-hosted tests all still pass?

keith-rollin:
Yes. I use that as a performance test, so I run it quite a bit. It still runs (and a lot more quickly than with Swift 1.1, I note).

kanaka:
Excellent! Merged.

PR #62, opened 2015-04-09
Require correct node_readline file

notduncansmith:
The previous version was requiring the `js` directory's version of `node_readline`, which crashes if `npm install` has not been run in that directory.

This commit uses the file located in the `coffee` directory. Running `make test^coffee` passes.


kanaka:
Oops. That was temporary but I never fixed it. Thanks for the catch.

notduncansmith:
No problem, thanks for making this awesome project!

PR #70, opened 2015-06-02
Mal implementation by Crystal

rhysd:
I implemented all steps of Mal by [Crystal](https://github.com/manastech/crystal).

I checked that my implementation [passed all test cases](https://gist.github.com/rhysd/ec6c27bb54ceeccad406) and self-hosted Mal implementation also [passed](https://gist.github.com/rhysd/f146c255626f4078eda3) in my OS X environment.


kanaka:
@rhysd something seems to be off with printing:

    $ crystal run ./step1_read_print.cr
    user> 123
    #<Mal::Type:0x2094f90>

    $ crystal run ./stepA_mal.cr 
    #<Mal::Type:0x17c4640>#<Mal::Type:0x17cf140>#<Mal::Type:0x17d4000>#<Mal::Type:0x17da240>#<Mal::Type:0x17dbc00>Mal [crystal]
    user> 123
    #<Mal::Type:0x17db0c0>#<Mal::Type:0x17e2fc0>

I tested with crystal-0.7.2 on Ubuntu 14.04.1

rhysd:
Really?
No such a thing is output in my environment (OS X, Crystal 0.7.1).  I'll check it with Crystal 0.7.2.

rhysd:
Oh, I could reproduce it with Crystal 0.7.2... I think it is compiler's bug because Crystal is yet on pre-alpha stage.

rhysd:
Can you use Crystal 0.7.1 temporarily?  I'll report the bug to Crystal repository.

rhysd:
Sorry, previous comments were my mistake.  It seems that I didn't notice the bug because of compiler's cache.  I'll fix it.

rhysd:
It seems that `print()` is added in Crystal 0.7.2 and Mal's `print` is not called.  I'll use `Mal` module to avoid that.

rhysd:
@kanaka 

I fixed the bug and confirmed step0 ~ stepA and self-hosted mal with Crystal 0.7.2.

rhysd:
Thank you!
I really enjoyed implementing mal.

kanaka:
@rhysd You're welcome! Do you have any feedback on the guide? Especially places where you got stuck that the guide could have helped address?

I did a couple of minor cleanups to the Makefile. I also pushed an update to the README to bump the language count (35 now!) and to list you as the creator of this implementation.

The only thing left is for you to tweet about it and I'll retweet :-)

rhysd:
@kanaka 

Sorry, I failed to catch your reply.

> I did a couple of minor cleanups to the Makefile. I also pushed an update to the README to bump the language count (35 now!) and to list you as the creator of this implementation.

I see. Thank you for your work.


> Do you have any feedback on the guide?

I think it is better to add more tests for object deepcopy.

Many languages (including Crystal) provide reference semantics, which employs shallow copy to copy array, dictionary and so on.  It leads to shallow copy of mal's object and enables to change a value in a environment from outside the environment.
I met a bug that breaks environment in self-hosted mal and I worked a bit hard to fix it.  Specifically, I forgot deepcopy for dictionary [here](https://github.com/kanaka/mal/blob/663fd47dac1450da5911511c2cd6d6a777c14cdf/crystal/stepA_mal.cr#L43-L45).  No test for previous steps couldn't find the shallow copy bug.

kanaka:
@rhysd that's a good point. Do you have some example tests that would have caught the problem in your case?

PR #73, opened 2015-07-06
Mal implemented in F#.

pstephens:
All the tests pass.

Also the self host tests pass except for the conj test for StepA.


kanaka:
@pstephens I've tested it and it looks good. The only thing I noticed is that when loading a file, the return values are printed. I think this is because each line of the file is being run through the REP function rather than calling load-file with the filename passed on the command line. For example, here is what is happening:

```
$ mono stepA_mal.exe ../tests/incC.mal 
{"a" 1}
"incC.mal finished"
nil
"incC.mal return string"
```

And here is what the behavior should be:

```
$ mono stepA_mal.exe ../tests/incC.mal 
"incC.mal finished"
```

I noticed this when running self-hosted tests where there is a lot of output before the prompt is presented because the return value of all statements are being printed. I went ahead and merged because that's not a critical issue, but it does limit the ability of mal to run mal scripts and get only the output you want output.

pstephens:
Just saw your comment. Will see if I can make the required changes to remove load time print statements.

PR #76, opened 2015-07-21
perl: add meta support for builtin function

kariya-mitsuru:


kanaka:
Tested and merged. Thanks!

PR #77, opened 2015-07-21
C#: fix arguments handling

kariya-mitsuru:
The arguments handling was incorrect like below.

- step6 to step9  
    `*ARGV*` was right but the filename of `load-file` was wrong.
- stepA  
    The filename of `load-file` was right but `*ARGV*` includes the
    filename.


kanaka:
Merged. Thanks for all the cleanup work you are doing!

PR #81, opened 2015-09-09
Add Elixir implementation

ekmartin:
Finally :tada:! All tests should pass, both regular and self-hosting.
I've updated the Dockerfile and README.

Really cool project and an awesome learning experience.


kanaka:
@ekmartin excellent!

If you get a chance, there is a couple of minor changes I would like to see:

- I think it would be better to use the "elixir" command rather than "iex" in both the Makefile and README. That way you don't get the extraneous Elixir REPL header.

- The header/info that is printed by stepA ("Mal [Elixir]") shouldn't be printed when there are command line arguments. In other words, the header/info is really just when it's run in regular REPL mode. For example, this:

```
$ elixir -S mix stepA_mal ../mal/stepA_mal.mal 
Mal [Elixir]
Mal [Elixir-mal]
mal-user> 
```

Should really be this:

```
$ elixir -S mix stepA_mal ../mal/stepA_mal.mal 
Mal [Elixir-mal]
mal-user> 
```

And finally, the only other thing is for you to tweet about the implementation and I'll retweet it. :-)

ekmartin:
Great! 

* The reason I'm running it with `iex -S` is to be able to use `iex`'s readline functionality, as there's no readline wrapper available for Elixir. If you want to I'll change it to just `mix stepA_mal` which compiles if necessary and runs the REPL, albeit without history/parenthesis matching and so on.

* Fixed [here](https://github.com/ekmartin/mal/commit/230dc1aa8de218ff0aae9696f7b7ee31bd640bbe), I'll make a pull request when we've resolved the previous point.

And [here's](https://twitter.com/ekmartinb/status/641945921962512384) my tweet :)

PR #82, opened 2015-09-10
Only print Elixir header info in REPL mode

ekmartin:
As for the `iex -S mix stepA_mal` vs `mix stepA_mal` issue:

> The reason I'm running it with `iex -S` is to be able to use `iex`'s readline functionality, as there's no readline wrapper available for Elixir. If you want to I'll change it to just `mix stepA_mal` which compiles if necessary and runs the REPL, albeit without history/parenthesis matching and so on.

What do you think?

### DIFF
--- a/elixir/lib/mix/tasks/step6_file.ex
+++ b/elixir/lib/mix/tasks/step6_file.ex
@@ -6,12 +6,10 @@ defmodule Mix.Tasks.Step6File do
     env = Mal.Env.new()
     Mal.Env.merge(env, Mal.Core.namespace)
     bootstrap(args, env)
-    load_file(args, env)
     loop(env)
   end
 
-  defp load_file([], _env), do: nil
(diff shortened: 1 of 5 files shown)


kanaka:
That makes sense. As an alternative, let's do this: change the Makefile to use elixir (since readline/REPL functionality isn't relevant there), and expand the instructions in the README to look something like this:

```
cd elixir
elixir -S mix stepX_YYY
    # OR
iex -S mix stepX_YYY  # to get readline/lineedit functionality
```

Does that sound reasonable?

ekmartin:
It does, I'll send a pull request.
Any reason you want to use `elixir -S mix stepX_YYY` and not just `mix stepX_YYY`? If I'm not mistaken they do pretty much the same, neither of them have the extraneous header either.

kanaka:
mix is fine. I'm not familiar with elixir usage so I was mostly just guessing.

PR #85, opened 2015-09-16
Changed nim/nimcache to nim/nimcache* in .gitignore

enitihas:
A seperate nimcache folder was being generated for each step with the
name of the step appended to nimcache, like nimcache-step0_repl and so
was not being ignored by git.
This change will prevent git from tracking
any folder with name starting with nimcache in the nim directory.


kanaka:
thanks!

PR #92, opened 2015-10-20
Add Vimscript implementation

dubek:
This pull request implements a full mal implementation in Vimscript, the built-in scripting language of the Vim editor. This passes all tests and all the self-hosted implementation tests on my Linux machine with Vim 7.4.

Readline is currently implemented as a short Vim "plugin" shared-object written in C (27 lines) which exposes the readline(3) function to Vimscript code. The implementation will run "make libvimreadline.so" if that file doesn't exist. Currently there's no history support. I'd prefer a pure Vim solution if possible but I couldn't think of one; I tried using Vim's internal `input()` function but it doesn't return any special value on EOF, which makes it a bit annoying.

The implementation is slow, so I extended the tests timeouts for vimscript in `Makefile`.

I had a lot of fun (and learned a lot). @kanaka - thanks for building such a great project.

Comments are of course welcome!


kanaka:
Awesome! 

Everything looks good but I had to make a small change to get it to load the dynamic lib:

```
diff --git a/vimscript/run_vimscript.sh b/vimscript/run_vimscript.sh
index 44dcf93..a0d948a 100755
--- a/vimscript/run_vimscript.sh
+++ b/vimscript/run_vimscript.sh
@@ -8,6 +8,8 @@
 # See: http://vim.wikia.com/wiki/Vim_as_a_system_interpreter_for_vimscript
 
 vimscriptfile="$1"
+rundir=`dirname $0`
+export LD_LIBRARY_PATH=`readlink -f $rundir`
 shift
 if [ x$DEBUG = x ] ; then
   exec 2> /dev/null
```

If that seems reasonable to you I'll merge it in with those changes.

dubek:
@kanaka thanks for the fix!

kanaka:
@dubek okay, it's merged. I also made a small tweak to the Dockerfile so that it will use the same layers as other implementations that use g++/gcc. I built/pushed the docker image and have added vimscript to the travis list. It's running right now (on all implementatiosn): https://travis-ci.org/kanaka/mal/builds/86485499

The only thing left is to announce/tweet it. If you would like to do it, drop a link to your tweet here and I'll retweet. If not (for whatever reason), I can just tweet it to announce it.

Also, there is a #mal freenode channel you're welcome to hang-out at. It's pretty quiet and number of people has dropped off. It might pick up after I talk about mal at midwest.io in a few weeks.

PR #93, opened 2015-10-24
Update step5_tco.mal

PhilipCollison:
test should succeed on correct implementation

### DIFF
--- a/tests/step5_tco.mal
+++ b/tests/step5_tco.mal
@@ -22,9 +22,8 @@ res2
 ;;; no try* yet, so test completion of side-effects
 (def! res1 nil)
 ;=>nil
-;;; For implementations without their own TCO this should fail and
-;;; leave res1 unchanged
+;;; For implementations with their own TCO this should succeed
 (def! res1 (sum-to 10000))
 res1
-;=>nil
+;=>50005000
 


PhilipCollison:
Hmm... it appears that this change should also modify the EXCLUDE_TESTS in MAKFILE to be exactly those tests that fail test5 (although I'm a bit curious about why some of those languages would fail...).

At this point I could either make the required changes in MAKEFILE and submit another pull request or you could close this request and I could drop the matter. What are you thoughts?

PhilipCollison:
Ah-ha!

After looking through the ruby example (and from my own experiences resolving this issue) it would appear that the TCO issues come from not evaluating argument before they're passed into the closure.

As an example from my own code,

    } else if(auto closure = cast(MalClosure)list[0]) {
        MalVal[] exprs = [];
        ast = closure.ast;
        env = new Env(closure.env, closure.binds, exprs);
        ...

gave me the same error that ruby currently has while

    } else if(auto closure = cast(MalClosure)list[0]) {
        MalVal[] exprs = [];
        foreach(v; list[1..$]) {
            exprs ~= EVAL(v, env);
        }
        ast = closure.ast;
        env = new Env(closure.env, closure.binds, exprs);
        ...

allows me to evaluate 50005000 successfully.

Conceptually arguments to functions should always be values as opposed to asts, otherwise you get  an inability to evaluate recusive functions as the arguments are never evaluated.

kanaka:
The main purpose of the second section of the step5 tests is to be a control for the first part. In other words, target languages without automatic TCO should still succeed for the first part but fail for the second part. A second purpose of the second section of step5 is to make sure that implementations without automatic TCO in the target language throw a stack overflow without crashing.

I added an issue to split the non-TCO test case into implementation specific tests directories: https://github.com/kanaka/mal/issues/94

PR #95, opened 2015-10-27
Kotlin

ivern:
Theres some cleanup I want to do (mainly tidying up the steps and cleaning up the types hierarchy), but I'm setting up the PR now so I can get any additional needed feedback.

Some comments:
* The latest version of kotlin available on Homebrew won't do.  Use this instead, or newer: https://github.com/JetBrains/kotlin/releases/tag/build-1.0.0-beta-1038.  Unzipping kotlin-compiler and setting KOTLIN_HOME and PATH should be sufficient.
* You're going to get a warning about an unnecessary cast when compiling.  I don't have a way around this, as there appears to be a bug in the Kotlin compiler--if I remove the "unnecessary" cast, then I get a type error.  I don't think this is a major problem, as it's just a compile time warning with no runtime effect.
* As far as I can tell, all required and optional tests pass in both direct and self-hosted mode.  I added a few new tests to catch some of the stuff that tripped me (mostly, my terrible implementation of swap!)
* I didn't bother implementing readline, as I used rlwrap for development and it gives me everything a readline integration would.  `rlwrap -H ~/.mal-history java -jar stepA_mal.jar` does the trick...available in pretty much any decent Linux package manager and Homebrew.

Any suggestions are welcome.  Thanks for making this awesome project!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,4 +64,6 @@ elixir/_build
 elixir/deps
 elixir/erl_crash.dump
 elixir/*.ez
-
+kotlin/*.jar
+kotlin/.idea
+kotlin/*.iml
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ mal_TEST_OPTS = --start-timeout 60 --test-timeout 120
 #
 
 IMPLS = awk bash c clojure coffee cpp crystal cs erlang elixir es6 factor forth fsharp go groovy \
-	haskell java julia js lua make mal ocaml matlab miniMAL nim \
+	haskell java julia js kotlin lua make mal ocaml matlab miniMAL nim \
 	perl php ps python r racket rpython ruby rust scala swift vb vimscript guile
 
 step0 = step0_repl
@@ -83,6 +83,7 @@ java_STEP_TO_PROG =    java/src/main/java/mal/$($(1)).java
 haskell_STEP_TO_PROG = haskell/$($(1))
 julia_STEP_TO_PROG =   julia/$($(1)).jl
 js_STEP_TO_PROG =      js/$($(1)).js
+kotlin_STEP_TO_PROG =  kotlin/$($(1)).jar
 lua_STEP_TO_PROG =     lua/$($(1)).lua
 make_STEP_TO_PROG =    make/$($(1)).mk
 mal_STEP_TO_PROG =     mal/$($(1)).mal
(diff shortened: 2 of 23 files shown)


kanaka:
Apart from the cleanup you already mentioned, especially the difference between step5-stepA from the first steps, I think it looks good (Kotlin is new to me so I can't really determine if it is good Kotlin code).

A couple of other minor things:

* could you make a Dockerfile for Kotlin (probably based on the Java one). I'll use that to activate Kotlin in travis.
* bump the count in the README.

kanaka:
BTW, if you don't have experience with docker or a place to run it to test it, don't worry about it, I can do that. Doesn't sound like the kotlin install process is too complicated so I should be able to figure it out.

kanaka:
Oh, one more thing. This is still a TODO in the guide, but if you have a chance could you also implement a time-ms function that returns the current milliseconds since epoch. I use that to be able to do a very rough benchmarking between implementations.

ivern:
The big change from step4 to step5 was that TCO made my step4 approach to eval cumbersome as I started needing to replace the values of ast and env.  There may be a way to make eval look way cleaner with TCO, but I'm not seeing it right now (I'm fairly new to Kotlin as well.)

I'll make the other changes now.  I think I can figure out Docker, I've it installed but haven't done a whole lot with it yet.  Gimme a few.

ivern:
Just pushed a docker file, which seems to work.  Running `docker run -v $HOME/projects/mal/:/mal -w /mal -t mal-kotlin make test^kotlin` after creating a mal-kotlin image from it runs all tests and whatnot.

ivern:
Hrm, I'm missing something.  Running make -C kotlin from Docker works, but trying to run the tests directly fails to compile them.  So right now it's a two step process of making, then running.  Outside docker, running the tests builds them as expected.

kanaka:
@ivern It actually worked fine for me. After building the docker image, I ran this:

```
docker run -it -u $(id -u) -v `pwd`:/mal kanaka/mal-test-kotlin make test^kotlin
```

Which did both the build and ran the tests successfully. I use $(id -u) since my mal directory is a NFS mount and doesn't allow root access. I would expect permission errors rather than failure to build.

kanaka:
@ivern I've gone ahead and merged. Also, I was able to run all tests including self-hosted tests using the container.

I pushed a minor change to the Dockerfile so that layers from the Java Dockerfile are re-used from the Java Dockerfile (given that the openjdk layer is fairly large and takes a while to build.

Feel free to keep cleaning up things until you're happy with it and I'll continue merging things in.

At some point, please announce your implementation by tweeting it and I'll retweet it (or I can just do it if you would prefer). Also, in case you aren't aware yet, there is also #mal on freenode IRC. It's pretty quite, but a few people involved in mal hang out there.

Thanks for the implementation. Yours is 42 which is an awesome number to have :-) Also, it's just in time for the presentation at midwest.io that I'll be doing about mal.

kanaka:
@ivern do you want to tweet to announce your implementation?

kanaka:
@ivern I'm getting antsy to announce this :-). You want to tweet it (and I'll retweet) or should I?

PR #96, opened 2015-10-27
Kotlin step cleanup

ivern:
Sorry, juuust missed your merge.  Here's some step cleanup.  I'm done for the night, I think things are in decent shape now.

### DIFF
--- a/kotlin/src/mal/core.kt
+++ b/kotlin/src/mal/core.kt
@@ -3,83 +3,81 @@ package mal
 import java.io.File
 
 val ns = hashMapOf(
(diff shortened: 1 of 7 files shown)


ivern:
Just added some additional cleanup--removed a bunch of boilerplate in core.kt, it's more readable this way I think.

All tests still pass and whatnot.

kanaka:
Okay, merged. That's a good test addition and yes, the vimscript atom? wasn't converting 1/0 to True/False values (another indication the tests were missing that). I pushed out a fix for that.

Issue #58, opened 2015-04-06
TCO test not general enough


lispm:
step5_tco.mal

TCO means 'Tail Call Optimization'. Not 'tail recursion'. TCO means that ALL tail calls are optimized, not only calls to a recursive function.

The test only tests the simple case of a self-recursive tail call. A test for the more general cases of TCO is missing.

```
; recursive, simple case

(define (foo a)
   (foo ...))  ; <- tail call

; mutually recursive. Will run in an endless loop, will not run out of stack space.

(define (bar a)
   (baz a))  ; <- tail call

(define (baz a)
  (bar a))  ; <- tail call

; general

(define (example-1 a)
  (+    ;  <-  tail call
     (* a a)
     (* a a a)))

; let's say you have a fantastillion functions and each calls the next one in a tail call
; -> will not run out of stack space 
```

kanaka:
#97 resolves this.

PR #97, opened 2015-10-27
Added TCO test for mutually recursive functions

ivern:
I believe this addresses https://github.com/kanaka/mal/issues/58

### DIFF
--- a/tests/step5_tco.mal
+++ b/tests/step5_tco.mal
@@ -28,3 +28,11 @@ res2
 res1
 ;=>nil
 
+
+;; Test mutually recursive tail-call functions
+
+(def! foo (fn* (n) (if (= n 0) 0 (bar (- n 1)))))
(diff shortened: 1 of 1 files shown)


kanaka:
Looks like several implementations fail that. I think the reason is because 1000000 is just too long for some of the implementations before the test times out. I did some quick experimentation and I think all those impls should pass with 10000 (which is still enough to blow the stack of languages that don't have TCO builtin).

Also, please move the mutually recursive tests above the ones that are expected to fail (below the regular recursive tests). Once that change is made and it passes the tests I'll merge it in. Thanks for doing that.

ivern:
Done, let's see what Travis says this time...

kanaka:
Right, regular TCO tests, mutually recursive, then the tests expected to fail.

kanaka:
Everything except rpython finished (takes 20 minutes but it passed last time). I'll also made a small tweak to the order to get the non-TCO function definition grouped with the non-TCO call.

PR #98, opened 2015-10-28
Factor updates

mrjbq7:
Some improvements:

* Works on soon-to-be-released 0.98 (triple quoted strings removed)
* 50% more iterations per second on perf3.mal
* Removed 'src' directory, use 'factor' as the root
* Using 'mal' namespace for core, reader, printer, malenv, types
* Fixed issue with "empty" readline input (should be ignored and wasn't)
* Tried to simplify some of the code
* If it matters, using a ``malkeyword`` class instead of funky unicode prefix

Before:

```
$ cd factor
$ make stats
      25     105     607 src/types/types.factor
      65     402    2237 src/reader/reader.factor
      35     257    1461 src/printer/printer.factor
      42     204    1042 src/malenv/malenv.factor
      76     697    3987 src/core/core.factor
     155    1007    5416 src/stepA_mal/stepA_mal.factor
     398    2672   14750 total
   81   149  2815 [comments/blanks]

$ make perf
Performance test for factor:
Running: factor ../factor/src/stepA_mal/stepA_mal.factor ../tests/perf1.mal
"Elapsed time: 2 msecs"
7
Running: factor ../factor/src/stepA_mal/stepA_mal.factor ../tests/perf2.mal
"Elapsed time: 4 msecs"
233
Running: factor ../factor/src/stepA_mal/stepA_mal.factor ../tests/perf3.mal
iters/s: 9204+2/3
nil
```

After:

```
$ cd factor
$ make stats
      34     146     787 mal/types/types.factor
      65     345    2077 mal/reader/reader.factor
      31     184    1050 mal/printer/printer.factor
      32     143     767 mal/env/env.factor
      74     641    3061 mal/core/core.factor
     153     913    4979 stepA_mal/stepA_mal.factor
     389    2372   12721 total
   76   144  2509 [comments/blanks]

$ make perf
Performance test for factor:
Running: factor ../factor/stepA_mal/stepA_mal.factor ../tests/perf1.mal
"Elapsed time: 1 msecs"
7
Running: factor ../factor/stepA_mal/stepA_mal.factor ../tests/perf2.mal
"Elapsed time: 1 msecs"
233
Running: factor ../factor/stepA_mal/stepA_mal.factor ../tests/perf3.mal
iters/s: 13757
nil
```

I've run a bunch of tests manually, but if someone wants to show me how to run a "validate this mal implementation" test, that would be great and I could do that too.


kanaka:
The regular tests need to pass:

```
make test^factor
```

Also, the self-hosted tests using factor also need to pass:

```
make MAL_IMPL=factor test^mal
```

mrjbq7:
Oh hm, looks like the original implementation had some test failures (and mine seems to have inherited it).  Will look into it.

kanaka:
Well, the existing failures are soft failures in stepA. Those are optional failures that aren't needed even for self-hosting. The new failures are start even in step1 (tests don't even start to run). See the Travis output: https://travis-ci.org/kanaka/mal/jobs/88013009

mrjbq7:
Those failures are configuration:

How does Travis configure the "factor" command? 

I changed it so the factor root is "mal/factor" not "mal/factor/src".

kanaka:
Travis uses a docker image and runs the tests like this:

```
docker run -it -u $(id -u) -v `pwd`:/mal kanaka/mal-test-factor make TEST_OPTS="--soft --log-file ../factor.out" test^factor
```

mrjbq7:
Okay, it looks like ``FACTOR_ROOTS`` was defined in the ``Makefile``.  Since I flattened the directory structure by removing the ``src`` directory, that broke the Travis tests.

Should be fixed now?

The tests that are broken seem to be ``(meta)`` and ``(with-meta)``...  I'm going to remove support completely for those rather than have a broken implementation I think.

mrjbq7:
Okay, I partially fixed ``with-meta``, now I see only a few failures with the meta stuff:

```
FAILURES: 8
make: *** [test^factor^stepA] Error 2
```

And I added some unit tests, nothing was broken that I tested but still good practice...

mrjbq7:
Should be good-to-go if the Travis tests pass.

kanaka:
I've gone ahead and merged it. Thanks for the updates!

The metadata tests aren't a big issue. Only function clojures/types must support metadata for the self-hosted tests to run. At some point, the latter tests for stepA I'm going to move into an extras test step. Basically, functionality that is useful/common for a Lisp to have, but not really part of the core mal implementation. For now though they are in the stepA tests and if you run the tests with --soft then they don't cause an error.

BTW, what is the rationale for factor not having "." as a root by default?  It seems like anything that is not a toy program will have multiple modules as part of the program and therefore will end up doing something like FACTOR_ROOTS=".".

For mal, is there any way we can do that in the step itself without needing external support (config files or environment variables)? I tried adding "." add-vocab-root to the top of the step file, but it appears that USING is a macro/pragma that is processed prior to any code running. Are there other options for how we can do this? The environment variable makes it more complicated to dockerize the tests (and is just sort of tacky when describing to the end-user how to run it. At the very least, it would be nice if the factor command line had a flag for adding roots.

mrjbq7:
Both of those are good ideas. Since Factor is more often used as long-lived image, particularly since loading code in "script" mode has a longer startup delay due to compiling into the image new libraries.

We have used a `.factor-roots` file to configure as well as the environment variable.

However, I can see that is awkward for this use case, so maybe a command line argument is a good idea too. 

I will check with @erg. 

Thanks for the feedback!

I'm going to blog about `mal` when I get a chance on my blog:

http://re-factor.blogspot.com

mrjbq7:
Oh, and factor supports adding roots at runtime. 

This would work at the top of the step files if you want that behavior.  The `<<` and `>>` make this code run essentially immediately before the rest of the code runs. 

```
<< USE: vocabs.loader "." add-vocab-root >>
```

kanaka:
@mrjbq7 thanks for filing that ticket. A command line parameter would make me happy even if "." doesn't get added as default. I look forward to your blog post (and let me know here when it lands I'll tweet/retweet it).

PR #101, opened 2015-11-04
Kotlin refactoring (more idiomatic)

ivern:
I've made a couple refactorings to make dispatching more idiomatic for Kotlin.


ivern:
The Travis failures seem to be related to the Erlang and VB implementations rather than these changes.

kanaka:
Merged. Erlang is currently broken with regards to string escaping/unescaping (I expanded the tests in that area). @nlfiedler should have a fix soon.

PR #102, opened 2015-11-06
Fix reading/printing of escaped characters

### LINKED ISSUE #100: Fix string escaping/unescaping in Erlang implementation
@nlfiedler I recently added some tests that do better coverage of string escaping and unescaping. In particular, this allows ANSI escape codes to be used in output to support. For example, being able to run the mal/clojurewest2014.mal presentation correctly.

I was able to fix the escaping/unescaping for all the other languages, but I'm not familiar enough with Erlang to be able to fix it easily. I think the escaping in printer.erl is okay, but the unescaping in reader.erl is missing the "\n" case. At least, that appears to be the issue. Do you think you could take a look at it?

You can reproduce the failures with:

```
make test^erlang^step4
```

nlfiedler:
Properly read \\n and convert to a newline, and when printing with
readability enabled, convert the newline back to \\n.

Add support for Erlang R18, which was nothing more than tweaking the
requirements in the rebar.config file.

Fixes #100

make test^erlang passes

### DIFF
--- a/erlang/rebar.config
+++ b/erlang/rebar.config
@@ -2,7 +2,7 @@
 %% rebar configuration file (https://github.com/rebar/rebar)
 %%
 
-{require_otp_vsn, "17"}.
+{require_otp_vsn, "17|18"}.
 
 {erl_opts, [debug_info, fail_on_warning]}.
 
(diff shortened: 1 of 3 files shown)


kanaka:
Great. Thanks @nlfiedler!

PR #104, opened 2015-11-09
tests: step9: add hash-map equality tests

dubek:
No one was testing that the `=` function works correctly on hash-maps. I introduce this test in step9 together with `hash-map`, `map?` etc.

### DIFF
--- a/tests/step9_try.mal
+++ b/tests/step9_try.mal
@@ -261,4 +261,14 @@
 ;=>true
 
 
+;; ------- Optional Functionality --------------
+;; ------- (Not needed for self-hosting) -------
+;>>> soft=True
 
+;; Testing equality of hash-maps
+(= {:a 11 :b 22} (hash-map :b 22 :a 11))
+;=>true
+(= {:a 11 :b 22} (hash-map :a 11))
+;=>false
+(= {:a 11 :b 22} (list :a 11 :b 22))
+;=>false


dubek:
After watching the travis results, I understand that this reveals a bug in many implementations... Or maybe it's not a bug (=undefined behaviour) and I missed something?

kanaka:
@dubek I don't think I've called it out anywhere as required functionality and obviously without test coverage a good number of the implementations don't support it. My suggestion is to add the following header to step9 tests before the hash-map equality tests:

```
;; ------- Optional Functionality --------------
;; ------- (Not needed for self-hosting) -------
;>>> soft=True
```

That's basically a configuration option to the runtest.py program to indicate that the following tests are truly optional and failures should just logged but non-fatal (I just pushed a change to make runtest allow soft tests by default). That way those tests will be captured (as something desirable) but they won't cause the tests to be fatal errors (unless --hard is passed to runtest). Sound reasonable?

dubek:
I agree. I pushed a version with the `soft=True` per your suggestion.

Note that in Python (which I happened to check) this doesn't return the wrong value - it actually hits an underlying exception:

```
Testing equality of hash-maps
TEST: (= {:a 11 :b 22} (hash-map :b 22 :a 11)) -> ['',true] -> SOFT FAIL (line 270):
    Expected : '(= {:a 11 :b 22} (hash-map :b 22 :a 11))\r\ntrue'
    Got      : '(= {:a 11 :b 22} (hash-map :b 22 :a 11))\r\nTraceback (most recent call last):\r\n  File "../python/step9_try.py", line 166, in <module>\r\n    print(REP(line))\r\n  File "../python/step9_try.py", line 143, in REP\r\n    return PRINT(EVAL(READ(str), repl_env))\r\n  File "../python/step9_try.py", line 134, in EVAL\r\n    return f(*el[1:])\r\n  File "/home/dmurik/world/mal/python/mal_types.py", line 39, in _equal_Q\r\n    if not equal_Q(a[akeys[i]], b[bkeys[i]]): return False\r\nNameError: global name \'equal_Q\' is not defined\r\n'
```

not too pretty...


How do you feel about adding this as a "requirement" in the process? And then slowly fixing the existing implementation one by one?

dubek:
Oh and how was the talk at midwest.io ?

kanaka:
Yep, I saw that problem in python after you mentioned it. It was code rot (anything not tested is broken as the axiom goes).

I'm definitely all for fixing the implementations one by one to add this. I'm a bit ambivalent about making it a strict requirement. For the most part my philosophy has been that any of the functionality that's not needed for self-hosting is marked soft/optional (although the line is definitely fuzzy). I think having it complain without failing will probably be enough motivation for the majority of people to add the functionality just to make their implementation completely clean in testing. However, I want to keep the mandatory things to a minimum if possible. There are some thing that are sort of on the fuzzy line. For example quasiquoting and macros are not strictly needed for self-hosting, but the mal implementation would be a LOT uglier without the use of "cond" and "or" macros. Plus full macros are still one of the things that continue to distinguish Lisp from most other languages. So I think it makes sense to have those as hard requirements. So I'm 100% on board with adding the tests as soft requirements and gradually bringing all the implementations up to snuff (or encouraging the original implementors to do so).

BTW, another thing that falls into this category of really nice to have but not mandatory for self-hosting is being able to treat strings as sequences (so that the sequence functions like first, rest, nth, etc do the right thing with strings in all implementations). There aren't any dedicated strings operations, but if sequence operations supported strings as a sequence of characters, then you wouldn't need another set of functions for strings (and that's how they work in some other Lisps).

I think the midest.io talk went well. Thanks for asking! Several people indicated they enjoyed it and got a lot out of it. Everything was recorded so at some point midwest.io will be posting the videos so you'll be able to see for yourself at that point.

PR #108, opened 2015-11-10
tcl: add --raw flag to allow running without Readline

dubek:
The Readline library for Tcl (tclreadline) enables history
substitutions, which means that lines that begin with `^` or `!` are
preprocessed by readline.  This interferes with some step1 tests of the
`^` form (--> with-meta).

Now the Tcl implementation supports a `--raw` command line flag which
disables the readline library and opts for simple input from stdin. The
tests (Makefile) are run with `--raw`.


kanaka:
Travis is down, but I verified it locally. Looks good to go.

PR #111, opened 2015-11-12
go: fix hash equality

dubek:


kanaka:
Looks like travis is back online, and the Go test passed (including the hash-map equality soft tests).

Issue #112, opened 2015-11-12
Pure mal equal implementation


dubek:
After implementing a few of those hash-equality fixes, I have an idea: how about we implement `scalar-equal?` with a native implementation, and implement the generic `=` in mal itself (like we implement `not`). The mal implementation will use `sequential?`, `hash_map?`, `contains?`, `get`, `first`, `rest` (and `scalar-equal?`)  to recursively check whether the two data structures are equal.

Of course changing it now is a bit late, because `=` is already implemented natively in all languages (and few still got hash-equality wrong).

Here is the "pure" mal implementation of `=` (called here `new-equal?`):

```
(def! scalar-equal? =)

;; need the `and` macro
(load-file "../core.mal")

(def! sequential-equal?
  (fn* [a b]
    (if (scalar-equal? (count a) (count b))
      (if (scalar-equal? 0 (count a))
        true
        (if (new-equal? (first a) (first b))
          (sequential-equal? (rest a) (rest b))
          false))
    false)))

(def! hash-vals-equal?
  (fn* [a b map-keys]
    (if (scalar-equal? 0 (count map-keys))
      true
      (let* [key (first map-keys)]
        (if (and (contains? a key) (contains? b key))
          (if (new-equal? (get a key) (get b key))
            (hash-vals-equal? a b (rest map-keys))
            false)
          false)))))

(def! hash-map-equal?
  (fn* [a b]
    (if (scalar-equal? (count (keys a)) (count (keys b)))
      (hash-vals-equal? a b (keys a))
      false)))

;; This implements = in pure mal (using only scalar-equal? as native impl)
(def! new-equal?
  (fn* [a b]
    (cond
      (and (sequential? a) (sequential? b)) (sequential-equal? a b)
      (and (map? a) (map? b)) (hash-map-equal? a b)
      true (scalar-equal? a b))))
```

kanaka:
[...]

PR #114, opened 2015-11-17
add examples/eqaulity.mal

dubek:
(as discussed in #112)

Implementations that don't have proper native comparison of hash-maps and sequences can load examples/equality.mal for a pure-mal implementation.

The file automatically detects if the native `=` works OK for hashes and sequences. If yes, it does nothing; if not, it redefines `=` as `mal-equal?`, which is the pure-mal implementation.

Example (perl mal as of Nov 2015 doesn't support hash-map equality):

```
$ perl stepA_mal.pl
Mal [perl]

user> (= {:a 11 :b 22} (hash-map :b 22 :a 11))
Error: TODO: Hash map comparison

user> (= {:a 11 :b 22} (hash-map :a 11))
Error: TODO: Hash map comparison

user> (load-file "../examples/equality.mal")
equality.mal: Replaced = with pure mal implementation
nil

user> (= {:a 11 :b 22} (hash-map :b 22 :a 11))
true

user> (= {:a 11 :b 22} (hash-map :a 11))
false
```

### DIFF
--- a/examples/equality.mal
+++ b/examples/equality.mal
@@ -1,0 +1,78 @@
+;;
+;; equality.mal
+;;
+;; This file checks whether the `=` function correctly implements equality of
+;; hash-maps and sequences (lists and vectors).  If not, it redefines the `=`
+;; function with a pure mal (recursive) implementation that only relies on the
+;; native original `=` function for comparing scalars (integers, booleans,
+;; symbols, strings).
+;;
+
+;; Save the original (native) `=` as scalar-equal?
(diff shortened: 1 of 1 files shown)


kanaka:
Awesome, thanks. I also fixed several implementations native equality last night (bash, c, coffee, cs, es6, java, js, julia, make, php. I think the only nine remaining to do are: forth, guile, kotlin, ocaml, miniMAL, perl, ps, r, and racket.

PR #115, opened 2015-11-18
ps: Fix hash equal

dubek:
* extract `_keys` and `_hash_map_get` to `types.ps`
* correct sequential types equals - only compare if *both* arguments are sequential
* implement equality for two hash-maps

I'd appreciate a review because it's almost the first postscript code I wrote.

Also added a few (optional) test cases to step9.

### DIFF
--- a/ps/core.ps
+++ b/ps/core.ps
@@ -42,13 +42,8 @@ end } def
     dup null eq { %if hash_map is a nil
         pop pop null
     }{ %else hash_map is not a nil
-        /data get % stack: args dict
-        exch 1 _nth % stack: dict key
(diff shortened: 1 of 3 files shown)


kanaka:
Looks good to me. Apparently github won't diff .ps files (thinks they are binary) so I had to pull it down to compare.

Yeah, stack based/concatenative languages are really interesting. If you have any further interest in stack/concatenative languages, I'd suggest  Factor. It happens to be the fastest dynamically typed mal implementation and it's also one of the smaller implementations. And it's a modern language with modern types, libraries, etc unlike Postscript which is (for good reason) largely unchanged over time :-)

PR #117, opened 2015-11-18
Upgraded Kotlin to a newer beta, solved new deprecation warnings

ivern:
I was also able to remove some unnecessary casts (but not all) since the relevant bug in kotlinc appears to be fixed.


ivern:
Looks like the Travis build server was shut down in the middle of the Kotlin test, and now it's not completing--perhaps the test runner needs a timeout value?

kanaka:
Travis seems to be having some issues right now. Things are taking a lot longer to run. And the kotlin docker image especially is taking forever to download (even though it's not that much bigger than many other docker images). Anyways, the most recent run at least passed the kotlin test (after 13 minutes!). So I consider this good to go.

PR #121, opened 2015-11-23
lua: fix time-ms to return actual millisecond resolution timestamp

dubek:
We install the luasocket rock to use its `socket.gettime()` function which returns sub-second resolution timestamps.


kanaka:
That's a great workaround for getting milliseconds. Thanks!

Issue #116, opened 2015-11-18
Implement native hash-map compare across all implementations


kanaka:
This is to track the remaining implementations that need native hash-map equality.

Here is current list of implementations which still need it native hash-map:
- [x] forth
- [x] guile
- [x] kotlin
- [x] matlab
- [x] miniMAL
- [x] ocaml
- [x] perl
- [x] r
- [x] racket
- [x] vb

@dubek, since we both seem to be working through them, let's claim them when we start working on them so that we don't duplicate work. So to start off, I'll do miniMAL.

dubek:
I'm out of town till next week so ain't gonna advance in the need few days.
Leave some for me!

dubek:
I'm claiming ocaml, and afterwards I'll try racket.

dubek:
ocaml fix in #118

dubek:
racket fix in #119 .

While coding this, I noticed another problem with implementations that rely on "native" `=` or `equal?` operator - they fail when you have nested vector/list comparisons. For example, in Racket, before the fix:

```
user> (= [1 2] (list 1 2))
true
user> (= [[1 2]] [(list 1 2)])
false
```

The first one is good (and tested in tests of step4); the second one is incorrect (and now fixed in #119 ).

So besides broken hash-map equality we also have this case. Not sure if you want to add this test case as mandatory (next to the other cases of vector/list equality in step4) or soft-fail.

kanaka:
It should eventually be with the other vector/list tests, but to avoid breaking Travis until they are fixed, let's make it soft for now until we have most fixed.

kanaka:
The embedded sequence implementations is tracked here: https://github.com/kanaka/mal/issues/123

kanaka:
Just pushed a fix for R.

dubek:
Guile fix in #124.

kanaka:
matlab pushed.

ivern:
@kanaka want me to fix the Kotlin implementation?

kanaka:
@ivern that would be great.

ivern:
I've fixed my pull request to contain a single commit.

ivern:
Kotlin should be updated to fixed now that it's merged.

kanaka:
I merged before the travis was finish with the PR, and unfortunately, travis doesn't pull the right thing so tests start to fail after the merge. So I'll wait for the travis builds for the merge itself to finish and verify and then close this if everything checks out.

[...]

PR #122, opened 2015-11-30
tests: add soft tests for nested vector/list equality in step 4

dubek:
(as discussed in #116)

For example, these nested list/vector tests currently soft-fail in the guile implementation.


kanaka:
Merged. Not sure what's going on with the nim impl failing in Travis.

Anyways, I filed a bug to track the broken implementations: https://github.com/kanaka/mal/issues/123

Issue #123, opened 2015-11-30
Fix nested equality tests and then make them mandatory


kanaka:
The following implementations fail the soft tests for nested equality. Sequence nested tests in step4 and hash-map nested steps in step9:

- [x] bash (hash-map in hash-map)
- [x] elixir (seq in seq)
- [x] erlang (seq in seq)
- [x] factor (seq in seq)
- [x] forth (no native hash-map equality, see #116)
- [x] guile
- [x] haskell (seq in hash-map)
- [x] julia
- [x] kotlin (no native hash-map equality, see #116)
- [x] php (seq and hash-map in hash-map)
- [x] scala

Once those are all fixed we should make those tests mandatory.

dubek:
Guile fix in #124.

kanaka:
I'll work on julia

kanaka:
I'll take scala too.

kanaka:
Okay, julia and scala fixes pushed. Also, I added some nested hash-map tests in addition to the nested sequences tests.

kanaka:
I updated the list with all the implementation that fail the new nested equality tests (including forth and kotlin which are missing hash-map equality entirely).

kanaka:
Fix bash and PHP in: https://github.com/kanaka/mal/commit/33d33bb31bc3015548a3ad0d00fc33daf1cf827a (silly bugs really)

kanaka:
Haskell was an easy fix too.

ivern:
Kotlin should be updated to fixed now that it's merged.

kanaka:
@jordanlewis @nlfiedler @ekmartin factor, erlang, and elixir and the only languages that still need this fix. Could you look at your respective implementation and fix the equality test when a sequence is embedded in another sequence? In other words, make these forms return true?

```
(= [(list)] (list []))
(= [1 2 (list 3 4 [5 6])] (list 1 2 [3 4 (list 5 6)]))
```

Thanks!

jordanlewis:
Sure I will take a look at this.

ekmartin:
Sure, I'll take a look.

kanaka:
Just factor left now.

kanaka:
@jordanlewis don't worry about factor since @dubek just fixed it.

kanaka:
Final thing is making the tests not soft tests. I just pushed that and will close this once travis gives the green light: https://travis-ci.org/kanaka/mal/builds/106320895

jordanlewis:
[...]

PR #124, opened 2015-11-30
guile: fix equality of hash-maps and lists/vectors

dubek:
(As part of #116 and #123)

Probably could use some better scheme styling. For example, maybe `_equal-lists?` and `_equal-hash-tables?` should be defined *inside* `_equal?` ? I'm not sure what's the best style here. I'll be glad to fix with some directions.

### DIFF
--- a/guile/core.scm
+++ b/guile/core.scm
@@ -29,7 +29,22 @@
 
 ;; Well, strange spec...
 (define (_equal? o1 o2)
-  (equal? (->list o1) (->list o2)))
+  (define (equal-lists? lst1 lst2)
+    (and (= (length lst1) (length lst2))
+         (for-all _equal? lst1 lst2)))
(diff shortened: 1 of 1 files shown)


kanaka:
Hmm, that is a bit weird looking. I'm not a guile expert at all (I do Clojure for my day job), but I think syntax-wise it's a clash between the leading underscore (I used for indicating a private function) and dashes which are a common way to separate words in variable and function names. Perhaps defining them as locals within the larger function would allow things to be cleaned up a bit (especially since they aren't likely to be use separately). Want to try that?

dubek:
You're right. Fixed. Also using `for-all` when comparing lists instead of my home-grown recursive impl.

kanaka:
Yep, looks a lot better. Merged.

PR #125, opened 2015-12-02
D implementation

dubek:
D (http://dlang.org) implementation - passes all tests and self-hosting tests. It has readline and history support via libreadline.

Tests of step5 are excluded because D is "small" enough to run 10,000 non-tail calls without crashing. At 1,000,000 it crashes fatally.  D has GC, so running an endless tail recursion has constant memory usage (doesn't crash).

During self-hosting I found a bug in my implementation (handling of hash-map with nil values), so I added the relevant test cases to step9 tests (commit 6496cd1d1732efafa1b63ebc2c3cdc3e3e48c347).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ bash/mal.sh
 c/mal
 coffee/mal.coffee
 crystal/mal
+d/mal
 haskell/mal
 js/mal.js
 js/web/mal.js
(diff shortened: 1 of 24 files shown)


kanaka:
I pushed a build of the docker image and activated travis tests which has passed: https://travis-ci.org/kanaka/mal/jobs/94509776

Issue #113, opened 2015-11-13
Testing commands fail in zsh


sleep:
Love this project, finding it super pedagogical!

The guide mentions the shell command

`make test^quux^stepX`

to run tests.

This works fine in bash. It does not work in zsh, however. After an entire night and morning of thinking it was something wrong in my implementation code that prevented me from running the tests, I found out that zsh treats unquoted carets as globbing characters.

Perhaps you should change the guide to say

`make "test^quux^stepX"`

for sake of universality?

Or make a note that the args should be quoted, for zsh users?
Or make a note that the command should be prefixed with `noglob`, i.e.
`noglob make test^quux^stepX`, for zsh users?

kanaka:
@sleep really sorry for the amount of time you spent figuring it out :-( although I'm glad you figured it out. I'm fine with either a note in the guide before the first test example or with quoting all the calls if that seems better. Would you like to send a PR with this change?

Issue #30, opened 2015-03-06
Not able to run tests.


moniya:
Just cloned the mal repo and I'm having problems running the tests. I use `make test^ruby^step0` to run the ruby implementation, but it's the bash implementation that gets run, and it even breaks, same for other languages I tried.

kanaka:
@moniya there are no tests for step0 (I may add some at some point just for completeness). Can you post the output that you get from running `make test^ruby^step1`? What OS are you on?

moniya:
I'm OSX Lion (10.7.5). I experience the same issue running other steps, too. Here's the output for `make test^ruby^step`:

```
Testing test^bash^step1, step file: bash/step1_read_print.sh, test file: tests/step1_read_print.mal
Running: ../runtest.py  ../tests/step1_read_print.mal -- bash ../bash/step1_read_print.sh
Started with:
../bash/types.sh: line 8: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
../bash/types.sh: line 8: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]

Testing read of nil/true/false
TEST: nil -> ['','nil'] -> SUCCESS
TEST: true -> ['','true'] -> SUCCESS
TEST: false -> ['','false'] -> SUCCESS
Testing read of numbers
TEST: 1 -> ['','1'] -> SUCCESS
TEST: 7 -> ['','7'] -> SUCCESS
TEST:   7    -> ['','7'] -> SUCCESS
Testing read of symbols
TEST: + -> ['','+'] -> SUCCESS
TEST: abc -> ['','abc'] -> SUCCESS
TEST:    abc    -> ['','abc'] -> SUCCESS
TEST: abc5 -> ['','abc5'] -> SUCCESS
TEST: abc-def -> ['','abc-def'] -> SUCCESS
Testing read of strings
TEST: "abc" -> ['','"abc"']
```
Completes with the following after a long pause:

```
TEST: "abc" -> ['','"abc"'] -> FAIL (line 34):
    Expected : '"abc"\r\n"abc"'
    Got      : None
TEST:    "abc"    -> ['','"abc"']Traceback (most recent call last):
  File "../runtest.py", line 173, in <module>
    r.writeline(form)
  File "../runtest.py", line 84, in writeline
    self.stdin.write(str + "\n")
IOError: [Errno 5] Input/output error
Error in atexit._run_exitfuncs:
[...]

PR #128, opened 2015-12-09
quoted all make args in documentation for zsh compatibility

sleep:
Addresses #113 and incidentally,  #30 

I thought that the `make foo^bar` pattern appeared in too many files in this project to warrant putting a disclaimer in each file for zsh users.


kanaka:
@sleep thanks! merged.

PR #129, opened 2015-12-10
examples: add memoize.mal as a usage example for atoms

dubek:
Adapted from the Clojure example of atoms: http://clojure.org/atoms

Benchmark:

```
$ cd examples
$ ../nim/mal memoize.mal
fib N= 32 : without memoization:
"Elapsed time: 14402 msecs"
fib N= 32 : with memoization:
"Elapsed time: 1 msecs"
```

### DIFF
--- a/examples/memoize.mal
+++ b/examples/memoize.mal
@@ -1,0 +1,53 @@
+;;
+;; memoize.mal
+;;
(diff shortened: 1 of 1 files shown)


kanaka:
Very nice :-)

PR #131, opened 2015-12-14
Makefile: add REPL invocation rules

dubek:
From README.md:

----

### Starting the REPL

* To start the REPL of an implementation in a specific step:

```
make "repl^IMPL^stepX"

# e.g
make "repl^ruby^step3"
make "repl^ps^step4"
```

* If you omit the step, then `stepA` is used:

```
make "repl^IMPL"

# e.g
make "repl^ruby"
make "repl^ps"
```

* To start the REPL of the self-hosted implementation, specify `mal` as the
  REPL implementation and use the `MAL_IMPL` make variable to change the
  underlying host language (default is JavaScript):
```
make MAL_IMPL=IMPL "repl^mal^stepX"

# e.g.
make "repl^mal^step2"   # js is default
make MAL_IMPL=ruby "repl^mal^step2"
make MAL_IMPL=python "repl^mal"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,6 +184,11 @@ DOCKER_BUILD = $(foreach impl,$(DO_IMPLS),docker-build^$(impl))
 
 IMPL_PERF = $(filter-out $(EXCLUDE_PERFS),$(foreach impl,$(DO_IMPLS),perf^$(impl)))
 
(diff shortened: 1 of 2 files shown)


kanaka:
Good idea! Merged.

PR #132, opened 2015-12-15
Native swap! implementation for cpp

sdt:
Here's the native swap! implementation, along with a minor env-passing cleanup.

The extra args are passed directly, there's no varargs handling like apply does. I can add that if necessary.

See #130


kanaka:
Thanks!

PR #133, opened 2015-12-15
runtest.py: Repeat failed test cases at the bottom

dubek:
Gather and re-print failed test cases at the end of the test suite. This saves the trouble of scanning up the 157 test cases and finding the two failures.

Example output:

```
...
...
Testing vector parameter lists
TEST: ( (fn* [] 4) ) -> ['',4] -> SUCCESS
TEST: ( (fn* [f x] (f x)) (fn* [a] (+ 1 a)) 7) -> ['',8] -> SUCCESS

-------- Soft tests --------
(when most implementations will implement these tests correctly this section
will be modified to hard failures)
Nested vector/list equality
TEST: (= [(list)] (list [])) -> ['',true] -> SUCCESS
TEST: (= [1 2 (list 3 4 [5 6])] (list 1 2 [3 4 (list 5 6)])) -> ['',true] -> SUCCESS

FAILURES:
FAILED TEST (line 314): (str (list 1 2 "abc" "\"") "def") -> ['',"(1 2 \"abc\" \"\\\"\")def"]:
    Expected : '(str (list 1 2 "abc" "\\"") "def")\r\n"(1 2 \\"abc\\" \\"\\\\\\"\\")def"'
    Got      : '(str (list 1 2 "abc" "\\"") "def")\r\n"(1 2 abc \\")def"'
FAILED TEST (line 317): (str [1 2 "abc" "\""] "def") -> ['',"[1 2 \"abc\" \"\\\"\"]def"]:
    Expected : '(str [1 2 "abc" "\\""] "def")\r\n"[1 2 \\"abc\\" \\"\\\\\\"\\"]def"'
    Got      : '(str [1 2 "abc" "\\""] "def")\r\n"[1 2 abc \\"]def"'

TEST RESULTS (for ../tests/step4_if_fn_do.mal):
    0: soft failing tests
    2: failing tests
  155: passing tests
  157: total tests

make: *** [test^ruby^step4] Error 1
```

### DIFF
--- a/runtest.py
+++ b/runtest.py
@@ -231,6 +231,7 @@ test_cnt = 0
 pass_cnt = 0
 fail_cnt = 0
 soft_fail_cnt = 0
+failures = []
 
 while t.next():
     log("TEST: %s -> [%s,%s]" % (t.form, repr(t.out), t.ret), end='')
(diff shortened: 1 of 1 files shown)


kanaka:
Great idea. I was feeling that pain too, but ignored it for some reason :-)

kanaka:
I think one possible issue with this, is when somebody is starting to work on a step for the first time, they are basically going to see every failure for every test twice. Maybe we should limit the the output at the end to the first five failures or something like that.

PR #134, opened 2015-12-15
tests: Add str and pr-str tests in step4 and step9

dubek:
Specifically test serializing nil, empty collections and `str` of
collections.  The Mal implementation is *not* identical to Clojure for
these cases.

I know the clojure mal impl is broken. Will wait for travis results to see what's the situation in other impls (c, ruby, python and racket have passed OK).

This will probably need some fixing because when serializing hash-map (step9) the order of keys might change. But I want to verify that no commas are inserted (current clojure impl does it).

### DIFF
--- a/clojure/src/core.clj
+++ b/clojure/src/core.clj
@@ -1,5 +1,6 @@
 (ns core
-  (:require [readline]))
+  (:require [readline]
+            [printer]))
 
 ;; Errors/exceptions
 (defn mal_throw [obj]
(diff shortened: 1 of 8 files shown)


dubek:
Here are the problematic ones:

1. clojure (nil, quoting, commas)
2. erlang (doesn't properly quote `"` in hash-map values)
3. forth (adds commas when serializing hash-maps)
4. guile (`str` quotes hash values)
5. ocaml (adds commas when serializing hash-maps)
6. python - problem in string equality - see #135

kanaka:
For the Clojure implementation I realized there is a much simpler solution which is basically just to coerce (str) using the *print-readably* binding (which IMO should have been the default for str anyways):

```patch
diff --git a/clojure/src/core.clj b/clojure/src/core.clj
index 6351137..3d31968 100644
--- a/clojure/src/core.clj
+++ b/clojure/src/core.clj
@@ -27,7 +27,7 @@
    ['keyword? keyword?]
 
    ['pr-str pr-str]
-   ['str str]
+   ['str #(binding [*print-readably* false] (apply str %&))]
    ['prn prn]
    ['println println]
    ['readline readline/readline]
```

The simple change above fixes everything except the empty list case ```(str (list))``` but that's actually a bug in Clojure that is fixed in 1.8.0-RC4: https://github.com/clojure/clojure/commit/828d82fb5bae9a985f32279b819e22ae436022da

I'm going to push the Clojure fix in a few minutes. If you'll drop the Clojure commit from your PR then I'll merge in the rest.

kanaka:
I went ahead and manually merged your commits except for the Clojure one (which required a couple more tweaks to get step9 tests to pass too).

dubek:
Thanks, I'm AFK for a couple of days.

PR #139, opened 2015-12-29
Updated mal.html

wendelb:
Due to the mixed content blocker in current versions of Firefox & Chrome the demo will not work on GitHub Pages. Loading jQuery via https will solve that issue.


kanaka:
thanks!

PR #143, opened 2016-01-21
Add gensym and clean `or` macro

dubek:
(Following discussion in #126)

This PR introduces `gensym` in the optional stage of stepA, implemented in Mal itself (part of the `rep` calls at the beginning of the program), and a clean `or` macro which doesn't shadow caller bindings.

There are separate commits for changes in tests, process, guide, core.mal, and then one commit which modifies the actual stepA code of 12 implementations (all pass the new tests of stepA).

If this looks good, I'll continue implementing the change on the rest of the implementations.


kanaka:
Overall looks good so far. I would probably leave the core.mal loading and "->" test in step8 to keep them with the macro step and then just drop the "and" tests. I had been thinking for a while to reduce the "and" tests. Slow implementations are especially slow with macros and the "and" macro is so similar to "or" macro that the tests for it are fairly redundant. If anything, one or two tests for "->>" in step8 would be more useful since that is a more complex macro.

dubek:
OK, sounds good, I'll restore core.mal back to step8 and replace `and` tests by several `->>` tests. Probably later tonight.

dubek:
As far as I can see all implementations pass the soft test at the end of test A which tests `gensym` and the clean `or` macro.

kanaka:
Nice work.

PR #144, opened 2016-01-22
Hashmap equality for kotlin

ivern:
Addresses #116, #123


ivern:
The failure seems to be rpython-related, FYI.

kanaka:
@ivern do you mind squashing that into a single commit with the three line change and without the four merges?

kanaka:
@ivern yeah, rpython builds are somewhat unreliable so they fail part of the time. At some point I'll update the version and see if that fixes the temperamental builds.

PR #145, opened 2016-01-22
Implemented hashmap equality (addresses #116, #123)

ivern:


ivern:
Re-created pull request with a single commit.

PR #154, opened 2016-01-27
python: Fix equality of unicode/normal strings in Python 2.x

dubek:
Python 2.6.6 and 2.7.10 fail the self-hosting tests of step4:

    make MAL_IMPL=python test^mal^step4

Maybe we should add the self-hosting tests to Travis-CI? Also I'm not sure if Travis runs Python 2 or 3.

I reduced one of the problems to the following:

```
Mal [python]
user> "abc"
"abc"
user> (str 'abc)
"abc"
user> (= "abc" (str 'abc))
false
```

The reason is that `"abc"` is stored as python type 'unicode' whereas the result of the `str` invocation is stored as python type 'str'. The _equal_Q function sees the different types and decides the objects are not equal - this is now fixed.

I added this as a test case in step7 (in which we introduce `quote`).

### DIFF
--- a/python/core.py
+++ b/python/core.py
@@ -23,8 +23,7 @@ def prn(*args):
     return None
 
 def println(*args):
-    line = " ".join(map(lambda exp: printer._pr_str(exp, False), args))
-    print(line.replace('\\n', '\n'))
+    print(" ".join(map(lambda exp: printer._pr_str(exp, False), args)))
     return None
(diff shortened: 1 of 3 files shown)


dubek:
The second commit fixes the newline problem that also failed self-hosting of step4. Now self-hosting passes on Python 2.

PR #156, opened 2016-01-28
Explain each regexp composing the main regexp used in the tokenizer

omarrayward:


kanaka:
@omarrayward looks good. Can you change it slightly so that those regex explanation bullets are one deeper (indented)?

omarrayward:
@kanaka Yep.
Regex explanations are now indented

PR #157, opened 2016-01-28
Pass extra function arguments (after &) as a Mal List

dubek:
When a function receives extra args (`& args`) these were stored as a
ruby/python array/tuple and not wrapped as a proper Mal List data type.
This works OK except when self-hosting *twice* (duck typing can only get
you that far :-).

Now these work OK:

```
python ../python/stepA_mal.py ../mal/stepA_mal.mal ../mal/stepA_mal.mal
ruby ../ruby/stepA_mal.rb ../mal/stepA_mal.mal ../mal/stepA_mal.mal
```

Tests were added in step4 to verify that the extra args parameter indeed
holds a Mal list.

- [x] clojure
- [x] forth
- [x] haxe (unrelated breakage, now fixed)
- [x] julia
- [x] python
- [x] ruby

### DIFF
--- a/clojure/src/env.clj
+++ b/clojure/src/env.clj
@@ -11,12 +11,12 @@
       (cond
         (= nil b)
         env
-        
+
         (= '& (first b))
         (assoc env (nth b 1) e)
-  
+
         :else
-        (recur (assoc env (first b) (first e)) (next b) (next e))))))
+        (recur (assoc env (first b) (first e)) (next b) (rest e))))))
 
 (defn env-find [env k]
   (cond
--- a/clojure/src/step4_if_fn_do.clj
+++ b/clojure/src/step4_if_fn_do.clj
@@ -56,7 +56,7 @@
 
         'fn*
         (fn [& args]
-          (EVAL a2 (env/env env a1 args)))
+          (EVAL a2 (env/env env a1 (or args '()))))
 
         ;; apply
         (let [el (eval-ast ast env)
--- a/clojure/src/step5_tco.clj
+++ b/clojure/src/step5_tco.clj
@@ -60,7 +60,7 @@
           'fn*
           (with-meta
             (fn [& args]
-              (EVAL a2 (env/env env a1 args)))
+              (EVAL a2 (env/env env a1 (or args '()))))
             {:expression a2
              :environment env
              :parameters a1})
--- a/clojure/src/step6_file.clj
+++ b/clojure/src/step6_file.clj
@@ -60,7 +60,7 @@
           'fn*
           (with-meta
             (fn [& args]
-              (EVAL a2 (env/env env a1 args)))
+              (EVAL a2 (env/env env a1 (or args '()))))
             {:expression a2
              :environment env
              :parameters a1})
--- a/clojure/src/step7_quote.clj
+++ b/clojure/src/step7_quote.clj
@@ -83,7 +83,7 @@
           'fn*
           (with-meta
             (fn [& args]
-              (EVAL a2 (env/env env a1 args)))
+              (EVAL a2 (env/env env a1 (or args '()))))
             {:expression a2
              :environment env
              :parameters a1})
--- a/clojure/src/step8_macros.clj
+++ b/clojure/src/step8_macros.clj
@@ -109,7 +109,7 @@
               'fn*
               (with-meta
                 (fn [& args]
-                  (EVAL a2 (env/env env a1 args)))
+                  (EVAL a2 (env/env env a1 (or args '()))))
                 {:expression a2
                  :environment env
                  :parameters a1})
--- a/clojure/src/step9_try.clj
+++ b/clojure/src/step9_try.clj
@@ -123,7 +123,7 @@
               'fn*
               (with-meta
                 (fn [& args]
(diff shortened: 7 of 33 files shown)


dubek:
Need to fix: clojure, forth, haxe, julia. Maybe tomorrow.

dubek:
BTW - another solution to the problem at hand is to make sure that impls of `count`, `first`, `rest` properly accept `nil` as an argument, like Clojure.

This will eliminate the need to force the type of `xs` in `( (fn* [& xs] (count xs) )` as a Mal List. Impementations can choose it to be an empty list or a nil, as long as it responds the same to `count`, `first`, `rest`.

If we take this approach we need to add these tests to step8:

```
(first nil)
;=>nil
(rest nil)
;=>()
```

`(count nil)` is already tested in step4.

@kanaka - Which approach would you prefer?

kanaka:
Hmm, both of the tests pass even without the change (although you're right that it does allow the double self-hosted mode to work). Also, the List classes in python and ruby return the same type when they are sliced or when drop is called on them. So I think there is some other issue that this change happens to work around.

You are right that the variable arguments should always be a mal list and not the native type (if they are different). Also, first and rest should support being called against nil as well.

dubek:
Before the change, the tests pass in python/ruby but fail in mal with the ruby/python impl (I think):

    make MAL_IMPL=python test^mal^step4

And regarding your last point: if there are no extra args, should the "xs" value be `nil` or `()` ? Consider this:

```
( (fn* [& xs] xs) )
```

In Closure xs is nil if I'm not mistaken. I think that an empty list is more consistent (meaning, the extra arg is *always* a Mal List, empty or not).

dubek:
Sorry, excuse me for the confusion. The added test cases in step4 don't pass on python and ruby before this change:

```
FAILURES:
FAILED TEST (line 170): ( (fn* (& more) (list? more)) 1 2 3) -> ['',true]:
    Expected : '( (fn* (& more) (list? more)) 1 2 3)\r\ntrue'
    Got      : '( (fn* (& more) (list? more)) 1 2 3)\r\nfalse'
FAILED TEST (line 176): ( (fn* (& more) (list? more)) ) -> ['',true]:
    Expected : '( (fn* (& more) (list? more)) )\r\ntrue'
    Got      : '( (fn* (& more) (list? more)) )\r\nfalse'
FAILED TEST (line 182): ( (fn* (a & more) (list? more)) 1) -> ['',true]:
    Expected : '( (fn* (a & more) (list? more)) 1)\r\ntrue'
    Got      : '( (fn* (a & more) (list? more)) 1)\r\nfalse'
```

So if we decide that we want to force a proper Mal List as the extra "args" argument (the one after `&`), these tests are valid and should be added.

I have a fix for Clojure and looking at Forth.

I'll open a separate PR for `(first nil)` and `(rest nil)` - about half the impls fail this.

kanaka:
Ah, yes, I was running python stepA. So to force that same code path regardless of whether you have TCO, a better test would be to use apply:

```
(apply (fn* (& more) (list? more)) (list 2 3 4))
```

Also, I think a better fix is to fix the usage of env rather than forcing env to do extra work when it doesn't need to (i.e. post-TCO for mal defined functions). So something like this for python and ruby (except for every step in ruby):

```patch
--- a/python/mal_types.py
+++ b/python/mal_types.py
@@ -81,7 +81,7 @@ def _keyword_Q(exp):
 # Functions
 def _function(Eval, Env, ast, env, params):
     def fn(*args):
-        return Eval(ast, Env(env, params, args))
+        return Eval(ast, Env(env, params, List(args)))
     fn.__meta__ = None
     fn.__ast__ = ast
     fn.__gen_env__ = lambda args: Env(env, params, args)
diff --git a/ruby/step4_if_fn_do.rb b/ruby/step4_if_fn_do.rb
index 151ecf6..85e80ef 100644
--- a/ruby/step4_if_fn_do.rb
+++ b/ruby/step4_if_fn_do.rb
@@ -59,7 +59,7 @@ def EVAL(ast, env)
         end
     when :"fn*"
         return lambda {|*args|
-            EVAL(a2, Env.new(env, a1, args))
+            EVAL(a2, Env.new(env, a1, List.new(args)))
         }
     else
         el = eval_ast(ast, env)
```

dubek:
OK, good point. I'll revise.

kanaka:
For the tests, it's probably best to leave the tests as you have them in step4 and then add apply versions in step9 (where apply becomes available).

kanaka:
Doh, Haxe was my fault. I forgot to submit the env/Env.hx when I added it.

dubek:
Julia fixed.

Haxe passes this issue, but Travis fails on performance testing:
```
Uncaught exception - 'time-ms' not found
Makefile:312: recipe for target 'perf^haxe' failed
make: *** [perf^haxe] Error 1
```

Forth - TBD.

kanaka:
I fixed the Haxe perf section too. I need to get better about doing full tests before pushing apparently :-) Maybe I'll try and get [travis-build](https://github.com/travis-ci/travis-build) running locally.

dubek:
Fixed Forth. All green now!

kanaka:
Excellent!

PR #160, opened 2016-01-31
Remove incorrect return in test case

omarrayward:
@kanaka Is this test correct?
`(def! a (atom 2))` shouldn't output `(atom 2)`, right?

### DIFF
--- a/tests/step6_file.mal
+++ b/tests/step6_file.mal
@@ -50,7 +50,6 @@
 (def! inc3 (fn* (a) (+ 3 a)))
 
 (def! a (atom 2))
-;=>(atom 2)
 
 (atom? a)
 ;=>true


dubek:
`def!` always returns the assigned value. So `(def! x (+ 3 5))` should return `8`. In this case, the value is an atom holding `2` which is printed as `(atom 2)`.

omarrayward:
@dubek So `(+ 3 (def! x 2))` should return `5`?

kanaka:
@omarrayward yes, `(+ 3 (def! x 2))` will return 5. Unlike many other Lisps, the def/defun forms operate on the current environment rather than a namespace. In most cases, `def!` works like you would expect if you come from another Lisp. In most Lisps, defun/def is generally only designed to be used from the top level, but in mal, because `def!` is defined not in terms of namespaces but in terms of mutating Lisp environments, the following:

```
(def! x 3)
(let* (y 4) (do (def! x y) (prn x)))
(prn x)
```

will print 4 and then 3 because `let*` creates a new environment and `def!` inside the `let*` modifies that environment and not the top level environment. This was not a mistake. In fact, I consider the way that namespaces work and are implemented in some other Lisps as one of my personal pet peeves. At some point mal may support namespaces, but my plan is that they would operate more like Lisp environments that and named/aliased I.e. `my.namespace/x` would refer to the `x` that exists in the environment named/aliased to `my.namespace`.

omarrayward:
@kanaka Thanks for the clarification

PR #161, opened 2016-02-01
guide: Document time-ms and conj

dubek:


kanaka:
A couple of small adjustments to that before merging:
- milliseconds since epoch (Jan 1, 1970) is preferable but milliseconds since any arbitrary point (e.g. program start) is also acceptable if epoch milliseconds are not available in a given implementation.
- conj takes a variable number of arguments which are either appended to the end of a vector or prepended to the beginning of a list (i.e. reverse order).

dubek:
Fixed.

kanaka:
Thanks

PR #162, opened 2016-02-01
erlang: Fix nested list/vector/map equality

dubek:
As described in #123 .

I also found another test case not covered, so I added it to the soft-tests in step9. We might see other impls soft-fail there if the map values weren't compared properly.

### DIFF
--- a/erlang/src/core.erl
+++ b/erlang/src/core.erl
@@ -84,13 +84,34 @@ equal_q(Args) ->
         [{string, S}, {string, T}] -> S == T;
         [{keyword, K}, {keyword, J}] -> K == J;
         [{symbol, S}, {symbol, T}] -> S == T;
-        [{list, L1, _M1}, {list, L2, _M2}] -> L1 == L2;
-        [{vector, L1, _M1}, {vector, L2, _M2}] -> L1 == L2;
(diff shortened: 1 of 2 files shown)


kanaka:
Good catch.

BTW, I've been using the following shell command (using the travis ruby gem command) to pull down logs locally so that I can grep/view them more easily:

```
BUILD=256; for x in $(seq 1 45); do echo ${BUILD}/${x}; mkdir -p logs/${BUILD}; travis logs ${BUILD}.${x} > logs/${BUILD}/${x}; done
```

Sometimes the logs take a while to be available for the command line tool (even after they complete and are available in the web interface). It's doesn't always happen, but when it does it's usually 5-10 minutes after completion before they are available from the command line.

Looks like factor is the only other implementation that fails these tests (and the nested tests).

dubek:
Actually elixir soft-fails this new step9 test. @ekmartin - want to try and tackle it following your #155 fix?

ekmartin:
#163, @dubek. Might have to refactor my equals implementation some day, there might be more cases like these that'll cause it to fail in regards of vectors/lists.

dubek:
Thanks @ekmartin .

I have a fix for Factor which passes step4 but still soft-fails on the new step9 test, so I'll work on that later.

PR #178, opened 2016-02-26
Fix negative numbers handling

dubek:
Negative numbers parsing and/or printing is broken in a few impls. I added simple tests in step1 and step2.

Here's a fix for ocaml, ps and haskell. I'll wait for travis for list of all broken impls.

* [x] bash
* [ ] erlang
* [x] forth
* [ ] fsharp
* [x] haskell
* [x] haxe
* [x] make
* [x] ocaml
* [x] ps


kanaka:
The make implementation is the reason that I punted on negative numbers. Make macros don't have any numeric operations, so everything is done with string operations (https://github.com/kanaka/mal/blob/master/make/numbers.mk). But it looks like now is the time ... so I'll take make.

Also, in languages without regex tokenization (make, ps, etc), you need a look-ahead in the parser (to be able to distinguish symbols that start with "-" from numbers).

kanaka:
Hmm, I was going to send a pull-request against your fix-negative branch (since I branched it locally), but I can't find how to do it in that direction. I.e. a pull request against an existing pull request. Have you done that before?

kanaka:
Cool. I think I figured it out. I had to use hub (https://github.com/github/hub), but I was able to create a pull request from my fix-request branch to your fix-request branch/PR by doing this:

```bash
hub pull-request -b dubek:fix-negative
```

dubek:
There was some mess because I force pushed to my branch (I thought no one was looking :-) and you based the PR on the original content. I ended up cherry-picking your `make` fix (that fix looks exhausting...) and push it into my branch; we'll see travis-ci soon.

I'm taking erlang and fsharp next.

kanaka:
I went ahead and cancelled all the other travis builds except the most recent: https://travis-ci.org/kanaka/mal/builds/112055853 Also, travis seems to be going super slow today. I wonder if the throttle heavy users? I'm sure there probably aren't many other projects that are quite so heavy.

kanaka:
I'll take bash

kanaka:
I'll track down the ps issue too.

kanaka:
I'll take haxe too (so I've got bash, haxe, and ps).

kanaka:
I took a different approach for ps that allocates exactly the correct amount of memory for the string conversion (including 0).

dubek:
Thanks for clearing up the postscript mess.

I fixed erlang too, so once travis is clear I think we're done with this one.

PR #183, opened 2016-03-06
Io implementation

dubek:
Full Mal implementation in [Io language](http://iolanguage.org/).

It all seems to work including self-hosting, but it's very very slow, even compared to slow implementations like vimscript. Specifically code involving macros is slow, with running (or false false false false true) taking several good seconds. I still want to look into that to see if there's some trivial error in that path in the code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ matrix:
     - {env: IMPL=guile,     services: [docker]}
     - {env: IMPL=haskell,   services: [docker]}
     - {env: IMPL=haxe,      services: [docker]}
+    - {env: IMPL=io,        services: [docker]}
     - {env: IMPL=java,      services: [docker]}
     - {env: IMPL=js,        services: [docker]}
     - {env: IMPL=julia,     services: [docker]}
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ mal_TEST_OPTS = --start-timeout 60 --test-timeout 120
 #
 
 IMPLS = awk bash c d clojure coffee cpp crystal cs erlang elisp elixir es6 \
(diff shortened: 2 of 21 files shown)


dubek:
OK indeed I have some working optimizations, so please don't merge yet. I found that `(x type == "ClassName")` is faster than `(x isKindOf(ClassName))` (I assume it's because isKindOf looks in the whole ancestry of prototypes to find a match).

dubek:
I improved `isMacroCall` and `==` performace in 1e85c48 . It feels better now but Io is still slow, probably due to the fact that *every* operation is "sending a message" (= dynamic call / virtual call), including things like if, while, try, method etc. See the [this comment](http://smalltalk.gnu.org/blog/zuluuuuuu/steve-dekortes-io-programming-language#comment-325).

Maybe I can still try to improve if I can find a way to reduce prototype chain depth. So look for places where I clone objects and try another copying approach. This should reduce the length of the prototypes ancestors chain and therefore reduce duration of method lookup.

kanaka:
BTW, I built and pushed a docker image for io based on what you have. That way PRs should be able to be tested via travis.

dubek:
I rebased this on top of current master and now it passes the travis build with `IMPL=io`.

Performance is still bad: perf1 and perf2 run much faster than bash, but perf3 is slow (13 iters/sec in Io compared with 300 iters/sec in bash). I already noticed that macro expansion is slow (like in `(or false nil false nil false nil false nil something)`) but couldn't find the reason.

kanaka:
I went ahead and merged it. Feel free to keep working on performance improvements if you want and I'll merge them in. It is unfortunate that it's slow slow, since it does have an appealing syntax. While performant implementations are nice, it's low priority for mal implementations. Personally, I would rather see progress on malc :-)

PR #184, opened 2016-03-09
mal: Add time-ms to core; add tests for time-ms

dubek:


dubek:
I just figured that the new test case might fail on impls that don't have millisecond resolution for time-ms. Let's wait for the travis build.

PR #186, opened 2016-03-15
Update ffi dep in js implementation

epylar:
The js dep won't compile cleanly without this bump because the old ffi version won't build properly 
(using latest Arch Linux as of Mar 14 2016, probably failing on other distros as well).  

ffi 2.0.0 builds cleanly.


kanaka:
Thanks!

PR #187, opened 2016-03-22
WIP: Add prolog implementation

ljos:
This is a work-in-progress prolog implementation.

Prolog, because of the shift in paradigm, needs to be implemented a little different than the rest of the languages. This is an attempt to do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,8 +85,8 @@ IMPLS = ada awk bash basic c chuck clojure coffee common-lisp cpp crystal cs d d
 	elisp elixir elm erlang es6 factor fantom forth fsharp go groovy gnu-smalltalk \
 	guile haskell haxe hy io java js julia kotlin livescript logo lua make mal \
 	matlab miniMAL nasm nim objc objpascal ocaml perl perl6 php picolisp plpgsql \
-	plsql powershell ps python r racket rexx rpython ruby rust scala scheme skew \
-	swift swift3 swift4 tcl ts vb vhdl vimscript wasm yorick
+	plsql powershell ps prolog python r racket rexx rpython ruby rust scala scheme \
+	skew swift swift3 swift4 tcl ts vb vhdl vimscript wasm yorick
 
 EXTENSION = .mal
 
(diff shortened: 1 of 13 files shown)


kanaka:
@ljos 

I've been following your prolog implementation. The main thing that I like to see (related to merging into the main tree) is that every implementation follows the overall structure and incremental process. Your implementation doesn't seem too far off (similar in many ways to the Haskell implementation actually). I think the main thing I notice is that the core functions should be in a separate module if possible rather than defined inline in the step (step4).

You would probably find the mal FAQ informative if you haven't already seen it: https://github.com/kanaka/mal/blob/master/docs/FAQ.md#why-is-some-code-split-into-steps-and-some-code-not

Also, if you use IRC, I'm often available at #mal on freenode during the day (US central time).

ljos:
Hi,

> the main thing I notice is that the core functions should be in a separate module if possible rather than defined inline in the step (step4).

I've been meaning to do that. I haven't done it because I am not sure if it will stay the way it is. There is something that feels off to me. I will probably rebase a couple more times before it will be ready. You can see in =eval= vs. =env= that in =eval= I used a hash-map instead of the built-in database. I think the database is better, but I feel like there should be a way of doing it.

The ugliest thing right now is how I treat =print_readable=. That is decidedly un-prologesque. I need to find a better way of doing that. I think it should be possible to make the DCGs carry a value with them when unparsing, but I am not sure how to do that. I will have to do some research...

Thanks for the comments. It will probably take some time for me to finish this as prolog requires me to think differently when I solve problems then when I use a other programming languages, so it goes a bit slow. I just thought others might add some ideas or find it fun to look at, even though it is not finished.

ljos:
I'm sorry this is taking some time. Still a while until I can work on this again.

kanaka:
@ljos Are you still interested in pushing this through to completion? If you aren't, or it may be a while please let me know and I'll close the issue until you are able to make progress. Thanks!

ljos:
I’m sorry that I have been slow. I was actually thinking about this the
last couple of days. If I haven’t made any progress over the weekend I
think you should close it. I will try to work on it Saturday and Sunday.

ljos:
It will take a bit longer than anticipated. I will close it and reopen when I have finished.

kanaka:
@ljos Okay, no problem. I'm happy to take a look again in the future. Also no time pressure, apart from the policy that the first submission for each implementation is the one that is included in the tree (unless the subsequent implementations are notably different in interesting ways).

PR #189, opened 2016-03-24
Tweaking text in guide.

Wilfred:
Referring to the pseudocode in the process/ directory is important, but
it's easy to miss that detail when reading the guide for the first time.

Similarly, it's important to add a new Makefile in many cases, so
reformat to draw attention to that.

### DIFF
--- a/process/guide.md
+++ b/process/guide.md
@@ -116,7 +116,9 @@ make "test^quux^stepX"
 
 TODO: If your implementation language is a compiled language, then you
 should also add a Makefile at the top level of your implementation
-directory that will define how to build the files pointed to by the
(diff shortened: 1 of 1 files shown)


kanaka:
Good points. Thanks.

process/guide.md
```diff
@@ -162,6 +164,8 @@ a bunch of tests associated with it and there is an easy script to run
 all the tests for a specific step in the process. Pick a failing test,
 fix it, repeat until all the tests for that step pass.
 
+## Reference Code
```
dubek: Please add a link to the TOC at the top.

PR #191, opened 2016-03-25
An Ada implementation of mal

zmower:
Native and self-hosting tests pass.
Not too happy with the perf test.  Seems do a lot of smart_pointer finalization according to gprof.
But at least things work correctly.


kanaka:
@zmower, I've checked out your latest changes and everything looks good. Just two more minor things:

- In stepA, right before entering the REPL loop (but only in REPL mode, not in command/script mode), print a startup header that contains the host language name before the first prompt (it's described in the guide but it's not tested so it's easy to miss). 

- Add your implementation to the README.md and bump the count to 51!

Once you've done that I'll merge it in.

zmower:
Done.  I've also posted on comp.lang.ada to spread the word a bit.

kanaka:
@zmower congrats! Good thought on posting it to comp.lang.ada.

Now that it is merged, the only remaining thing is for you to tweet about it and I'll retweet (unless you prefer not to in which case I'll just tweet it).

So what is your next implementation language going to be? ;-)

PR #192, opened 2016-03-25
Several fixes to the guide

joostkremers:
Hi,

I worked through the guide and created an implementation of Mal in Python 3. I mainly wrote it to learn Python, so I guess it's of little interest to you, but along the way I found some typos and such in the guide, which you may be interested in, which is why I'm creating this pull request.

I really enjoyed working through the guide. I learnt a lot, not just about Python, but also about implementing a Lisp interpreter (which turned out to be much less complicated as I had though, BTW.) Thank you very much for writing Mal!


kanaka:
@joostkremers thanks!

PR #198, opened 2016-04-08
Port swift3 to OSX.

keith-rollin:
Here are some changes to get swift3 compiling and running under OSX. There were only a few tweaks, so it should be pretty easy to vet.

I'm not sure why I had problems with the implementations of "pr-str" and "str". Perhaps it's because the compiler I'm using (DEVELOPMENT-SNAPSHOT-2016-03-24-a) is more recent than the one used for original development.

### DIFF
--- a/swift3/Makefile
+++ b/swift3/Makefile
@@ -1,3 +1,9 @@
+ifneq ($(shell which xcrun),)
+	SWIFT = xcrun -sdk macosx swiftc
+else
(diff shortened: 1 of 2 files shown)


kanaka:
Thanks!

kanaka:
I updated travis to build Swift3 for both OS X and Linux and it is passing for both: https://travis-ci.org/kanaka/mal/builds/121703156

PR #199, opened 2016-04-08
c: Add Boehm garbage collector

dubek:
Add the [Boehm garbage collector](http://hboehm.info/gc/) to the C implementation. The README and Dockerfile changes describe the new dependency for building.

## Build

By default, garbage collection is enabled.  You can disable the garbage collection by building with an empty `USE_GC` setting:

    make USE_GC=

## Testing

Performance testing used to cause OOM killer on the `stepA_mal` process in my VM:

    Performance test for c:
    Running: ../c/stepA_mal ../tests/perf1.mal
    "Elapsed time: 2 msecs"
    Running: ../c/stepA_mal ../tests/perf2.mal
    "Elapsed time: 2 msecs"
    Running: ../c/stepA_mal ../tests/perf3.mal
    Killed

Here are the relevant syslog lines:

    [938589.062424] Out of memory: Kill process 3769 (stepA_mal) score 877 or sacrifice child
    [938589.063582] Killed process 3769 (stepA_mal) total-vm:957152kB, anon-rss:891592kB, file-rss:64kB

and now it completes OK:

    Performance test for c:
    Running: ../c/stepA_mal ../tests/perf1.mal
    "Elapsed time: 1 msecs"
    Running: ../c/stepA_mal ../tests/perf2.mal
    "Elapsed time: 2 msecs"
    Running: ../c/stepA_mal ../tests/perf3.mal
    iters/s: 5248

I also tried this endless loop:

    (def! silent-loop (fn* [n] (if (> n 0) (silent-loop (- n 1)))))
    (def! loop (fn* [n] (do (prn n) (silent-loop 100000) (loop (+ 1 n)))))
    (loop 1)

Before this addition, this used to hog the VM's memory:

     $ ./stepA_mal endless.mal
     1
     2
     3
     4
     5
     6
     7

and with garbage collection it keeps running as far as I was willing to wait.

Note that it's still possible that some Mal functions/flows are still leaking
memory (especially with the interaction with glib); maybe we need to run all
the regression functional tests in a loop (inside one Mal process) and monitor
memory growth to find such a situation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,8 @@ bash stepX_YYY.sh
 ### C
 
 The C implementation of mal requires the following libraries (lib and
-header packages): glib, libffi6 and either the libedit or GNU readline library.
+header packages): glib, libffi6, libgc, and either the libedit or GNU readline
+library.
 
 ```
 cd c
--- a/c/Dockerfile
+++ b/c/Dockerfile
@@ -25,4 +25,4 @@ WORKDIR /mal
 RUN apt-get -y install g++
 
 # Libraries needed for the C impl
-RUN apt-get -y install libglib2.0 libglib2.0-dev libffi-dev
+RUN apt-get -y install libglib2.0 libglib2.0-dev libffi-dev libgc-dev
--- a/c/Makefile
+++ b/c/Makefile
@@ -1,4 +1,5 @@
 USE_READLINE ?=
+USE_GC ?= 1
 CFLAGS += -g -O2
 LDFLAGS += -g
 
@@ -36,6 +37,12 @@ RL_LIBRARY ?= edit
 else
 RL_LIBRARY ?= readline
 CFLAGS += -DUSE_READLINE=1
+endif
+
+ifeq (,$(USE_GC))
+else
+CFLAGS += -DUSE_GC=1
+LDFLAGS += -lgc
 endif
 
 CFLAGS += $(GLIB_CFLAGS)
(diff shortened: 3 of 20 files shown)


kanaka:
Yay! Merged.

I also pushed an updated docker image prior to merging so the travis build for the merge should pass.

dubek:
Does it improve the `make perf^c` results for you?

kanaka:
Yes, it makes a big improvement. I have a 12 GB Core i7 system. Here are six runs without the fix:

```
iters/s: 2077
iters/s: 2974
iters/s: 3516
iters/s: 4367
iters/s: 4791
ters/s: 4838
```

And during the execution it gets up to about 5GB of memory. I believe the increase in speed is because it causes other running programs to be swapped out with each run.

Here is after the fix:

```
iters/s: 5287
iters/s: 5435
iters/s: 5244
```

And the max memory usage (which was steady and not increasing during the run) was about 24MB.

I would say that's a slam dunk. Boehm is awesome. I'm not sure how much C application development I'll do in the future, but if I do, Boehm is certainly going to be part of my toolset.

dubek:
Nice.

Boehm was very easy to integrate, *except* where it came to 3rd party code (in our case - glib). Luckily glib allows you to set your own malloc/free, so I could make it use Boehm's GC_malloc etc. But still it caused some issues (for example `g_print` kept causing crashes, so I replaced it with puts).

If you're going use Boehm, make sure your dependencies play nicely; or that they handle their own memory objects which never interact with Boehm-allocated objects.

kanaka:
Good advice, thanks.

PR #203, opened 2016-04-27
Fix evaluation of hash-map keys

dubek:
Add instructions to guide (in step 2) regarding evaluation of vector and hash-map elements.  These were already tested in step 2 tests.

The new test for hash-map keys evaluation in step 4 uses `str` which is introduced in step 4.

Travis will fail this PR on impls that don't evaluate the hash-map keys correctly. Failed impls:

* [ ] ada
* [ ] awk
* [ ] bash
* [ ] c
* [ ] cpp
* [ ] coffee
* [ ] cs
* [ ] crystal
* [ ] d
* [ ] elisp
* [ ] erlang
* [ ] es6
* [ ] fsharp
* [ ] go
* [ ] guile
* [ ] haskell
* [ ] haxe
* [ ] io
* [ ] java
* [ ] js
* [ ] kotlin
* [ ] make
* [ ] mal
* [ ] matlab
* [ ] miniMAL
* [ ] nim
* [ ] objpascal
* [ ] objc
* [ ] perl
* [ ] php
* [ ] ps
* [ ] python
* [ ] r
* [ ] racket
* [ ] rpython
* [ ] rust
* [ ] scala
* [ ] swift3
* [ ] tcl
* [ ] vb
* [ ] vimscript

### DIFF
--- a/process/guide.md
+++ b/process/guide.md
@@ -543,6 +543,15 @@ make "test^quux^step2"
 
 You now have a simple prefix notation calculator!
 
+#### Deferrable:
+
+* `eval_ast` should evaluate elements of vectors and hash-maps.  Add the
+  following cases in `eval_ast`:
+  * If `ast` is a vector: return a new vector that is the result of calling
+    `EVAL` on each of the members of the vector.
+  * If `ast` is a hash-map: return a new hash-map that is the result of calling
+    `EVAL` on each of the keys and the values of the hash-map.
+
 
 <a name="step3"></a>
 
--- a/tests/step4_if_fn_do.mal
+++ b/tests/step4_if_fn_do.mal
@@ -453,3 +453,7 @@ nil
 ;=>true
 (= [1 2 (list 3 4 [5 6])] (list 1 2 [3 4 (list 5 6)]))
 ;=>true
+
+;; Test evaluation of hashmap keys and values
+{ (str "a" "b" 34) (+ 2 3) }
+;=>{"ab34" 5}


dubek:
aaaa, Python dict can't have list as a key because list is unhashable. So for python this fails when the reader tries to call `_hash_map` with a key that is a List... Maybe replace the underlying List's parent class `list` to a `tuple`? And that's just the first impl I'm looking into...

dubek:
Re: Python: I can simply implement `__hash__` on List (the mal type); should allow it to be a dict key.

kanaka:
For now we should just fix the key eval issue and just error when a key evals to something that is not a string. That's a fairly straightforward problem to solve and shouldn't require any heroics for any of the implementation languages. Supporting hash-map keys that are arbitrary objects could require some major refactoring across a number of implementations (storing the key objects separately and having some sort of string based object reference for the keys). If implementations happen to support non-string keys in hash-maps, that's fine, but I don't think it should be required.

A user of mal could always build their own new map type directly in mal that supports object keys if they really wanted (just without an object literal syntax, but whatever).

dubek:
But python fails during *reader* for the input `{ (str "a") 123 }` . You don't want the reader to evaluate that `(str "a")` expression...

kanaka:
@dubek oh shoot, you're right. Unfortunately, I don't think there is a way to solve the problem without some nasty complexity across a bunch of implementations. Given that this can be worked around using hash-map or assoc (e.g. (hash-map (str "a" "b") 123) ) I think I would just like to punt on (skip) this one. Sound reasonable?

dubek:
@kanaka Yes, I agree. This can be hairy. You can close this PR as wontfix, I'll open another one with changes to the guide that are needed regardless.

kanaka:
Closed.

PR #204, opened 2016-04-27
process/guide: add vector and hash-map evaluation in step 2

dubek:
I bet the phrasing can be improved. Feel free to modify.


kanaka:
Merged.

PR #205, opened 2016-04-30
Fix broken link

Wilfred:
Previously, this would link to https://github.com/kanaka/mal/blob/master/factorcode.org


kanaka:
Thanks!

PR #208, opened 2016-05-09
vimscript:  implement time-ms in C

dubek:
1. Rename libvimreadline to libvimextras
2. Implement `time-ms` in a C function instead of shelling out to `date`. Vimscript numbers are limited to 32-bit so we return the number of milliseconds since 2000-01-01 instead of 1970.
3. Fix `docs/Hints.ms` - now only bash & make shell out to `date` to implement `time-ms`.


dubek:
Travis failed on docker pull timeout on ocaml.

PR #209, opened 2016-05-09
python: Fix interop (py*, py!*)

dubek:
`py!*`: Fix to work on both Python 2 and 3.

`py*`: Python types are converted to Mal types where applicable.

Added tests for Python interop.


kanaka:
Cool, thanks.

PR #211, opened 2016-05-11
racket: Set *ARGV* correctly

dubek:
I started working on automatic test for `*ARGV*` (expect a PR soon) and while testing I found that Racket wasn't setting *ARGV* at all. Now fixed.


dubek:
I cancelled all travis jobs and restarted racket - it's green: https://travis-ci.org/kanaka/mal/jobs/129420278

PR #212, opened 2016-05-11
tests: Test *ARGV* is set correctly in step 6

dubek:
Add new mini test harness run_argv_test.sh to run the Mal interpreter
with different command-line arguments and test the stdout of that
process.

The main Makefile will automatically run the new harness whenever step 6
is tested (either directly or during REGRESS=1 of a more advanced step).

----

I wrote this as a simple bash script because we only need these 3 test cases. If we happen to add more complex testing here, I think it's worth considering a proper test runner like `runtest.py`.

Also note that I test the stdout of the Mal process byte-by-byte. Any Mal implementation that output some garbage bytes at beginning or end will fail these tests. I'm not sure if it's going to be an issue.

----

Broken:

* [x] cpp
* [x] erlang
* [x] factor
* [x] fsharp
* [x] io
* [x] java
* [x] js
* [x] make
* [x] matlab
* [x] ocaml
* [x] r
* [x] scala


dubek:
Eventually I'll probably ease up the tests to look only at the actual argv output line, to circumvent this failure:

```
Testing ARGV of test^make^step6; step file: make/step6_file.mk
Running: ../run_argv_test.sh make -f ../make/step6_file.mk 
FAIL: Expected '("aaa" "bbb" "ccc")' but got 'make[1]: Entering directory '/mal/make'
("aaa" "bbb" "ccc")
make[1]: Leaving directory '/mal/make''
```

but first I want to clear out all the others (for example, debug print leftover in ocaml and so on).

dubek:
Remaining problems:

* **io**: The io prints "Registering Regex: Regex" to stdout whenever it starts (actually it's the first time the Regex module is loaded, which is the beginning for all relevant steps). We can re-compile the module with this printf commented out, but that requires cmake. Maybe write a wrapper script that removes that "prefix", but `sed` isn't good here because it works line-by-line so it hides the `user> ` prompt when working interactively.
* **java**: Arguments with spaces are broken. Possible to fix by adding a `run_java.sh` wrapper that will put arguments inside single quotes and then pass them to `-Dexec.args=`.
* make: Need to add `--silent` to the command-line, and arguments with spaces are broken. Don't know if possible to easily fix.
* **matlab**: For Octave, I can fix with a wrapper script which will single-quote args and join them with comma and pass that to the `--eval` command line; also adding `--no-history` supresses some error messages. I assume matlab is similar but can't test.
* **scala**: sbt prints some info log lines to stdout and messes up the script's output. Also the call to `System.exit(0)` causes a `Exception: sbt.TrapExitSecurityException thrown from the UncaughtExceptionHandler in thread "run-main-0"`. Maybe we can change the main function to return at the end instead of calling `exit`.

So, I think that:

1. Arguments with spaces are overrated. Let's remove mandatory support for that.
2. Some impls are noisy - print stuff before/after the actual script's output. Make the argv test harness ignore that noise.

This leads me to a future (longer) issue - I think that we should replace the xxx_RUNSTEP system with a `run.sh` that will be in *every* impl directory. It'll accept a clean step name as first argument and optional command-line args: `./run.sh step6_file ../tests/print_argv.mal aaa bbb` . So for Python it'll be:

```
#!/bin/bash
exec python $1.py "${@:2}"
```

(probably can be a one-liner somehow)

And for octave it'll be:

```
#!/bin/bash
step=$1
shift
args=$(for a in "$@" ; do echo -n "'$a'," ; done)
args=${args%,} # Remove comma suffix
exec octave --no-gui --no-history -q --traditional --eval "$step($args);quit;"
```

(and we already have other wrapping scripts for plpgsql, vimscript and so on.)

Or, put more generally - I'd prefer all the implementation-specific code (how-to-build, how-to-run, how-to-pack/dist) will be sitting inside the implementation dir, and the Makefile (ideally) will only hold a list of implementation names.

kanaka:
I agree with spaces in arguments being optional (although I think make is probably the only case that can't be fixed). And I think run.sh for every implementation is a good idea. For the simple cases of run.sh (i.e. one liners) I think we should still list the raw way to run the implementation (without run.sh) in the README (along with an overall note saying run.sh is universal optional for invoking the implementation).

Also, I think that run.sh should accept the full step file name including the extension. Then run.sh can trim it off the path prefix and the extension if it needs to for the target implementation. That makes it so that most implementations have almost the same run.sh script code and it also makes tab completion more usable in that context.

dubek:
@kanaka ack. I'll fix the remaining problems according to these guidelines, we can slowly introduce `run.sh` files as needed. 

Anyone with experience with sbt wants to help on scala? Otherwise I'll try whatever in my powers...

dubek:
@kanaka Re run.sh first argument - what about compiled languages? Should it receive `stepA_mal.c` or `src/stepA_mal/stepA_mal.go` even though the `run.sh` would not compile it but just run `stepA_mal` (the compiled executable) ?

kanaka:
@dubek for compiled programs it should be the compiled program. Basically it should be the result of foo_STEP_TO_PROG that is passed to run.sh as the first argument.

dubek:
@kanaka OK. What do you think about calling it `run` (or maybe `runstep`) instead of `run.sh`? I assume it'll be mostly implemented in `/bin/sh` but this way we can implement in other languages if makes sense. 

My initial idea of passing just the step name (`step6_file`) was to eventually eliminate both the list of `xxx_RUNSTEP` and list of `xxx_STEP_TO_PROG` from the main Makefile. But it doesn't really matter.

Now I also see that `mal` might need a special treatment (I need to know how to execute the `MAL_IMPL`). Hmm.

kanaka:
Oh, related to scala, maybe now is the time to refactor the Makefile to only use sbt to build the jar for the requested step. Then the jar file can just be invoked directly (and more efficiently):

    java -jar path/to/stepX_Y.jar ../tests/arg.mal 7 8 9

This does mean figuring out how to get sbt to build a specific step which much be a hassle. But what it's doing right now is a hack that I'm not happy with.

kanaka:
I think `run` is good (I had the same thought actually).

So I think you are right that we should pass the step name (so that it's the same regardless of run file). However, rather than passing in on the command line, let's make it an environment setting and default to stepA_mal.

In other words, the js/run script would look like this:

    #!/bin/bash
    exec node $(dirname $0)/${STEP:-stepA_mal}.js "${@:1}"

And the mal/run script would look like this:

    #!/bin/bash
    cd $(dirname $0)
    MAL_FILE=./${STEP:-stepA_mal}.mal
    STEP=stepA_mal # force MAL_IMPL to use stepA
    exec ../${MAL_IMPL:-js}/run ${MAL_FILE} "${@:1}"

The top level Makefile can then just set STEP (and MAL_IMPL) in the environment and call the run script for the current implementation. Bonus: we can get rid of a lot of the self-hosted vs regular shenanigans that it is doing now.

An alternative that I don't like as much, would be for the Makefile to have a single utility routine get_RUNSTEP (rather than the big foo_RUNSTEP switch). Might look something like this:

    get_RUNSTEP = $(strip $(if $(filter mal,$(2)),\
                    ../$(MAL_IMPL)/run ../$(call $(MAL_IMPL)_STEP_TO_PROG,stepA) ../$(call mal_STEP_TO_PROG,$(1)) $(3),\
                    ../$(2)/run ../$(call $(2)_STEP_TO_PROG,$(1)) $(3)))

    # Examples:
    $(warning (get_RUNSTEP,stepA,ruby,7 8 9): $(call get_RUNSTEP,stepA,ruby,7 8 9))
    $(warning (get_RUNSTEP,step6,guile,7 8 9): $(call get_RUNSTEP,step6,guile,7 8 9))
    $(warning (get_RUNSTEP,step4,mal,7 8 9): $(call get_RUNSTEP,step4,mal,7 8 9))

However, that's basically means not having a run script in mal/.

dubek:
@kanaka Good, I like this solution with `$STEP` and that the Mal implementation can run any other impl (`$MAL_IMPL`) just by calling its `run`. 

In this PR I'll only work on those broken above.

Later (in separate PRs) we can slowly replace more and more `xxx_RUNSTEP` with a simple calls to `run`, with Mal being the last to change. Once all are replaced we can eliminate all `xxx_RUNSTEP` from Makefile (and probably simplify the `test^xxx`, `repl^xxx` and `pref^xxx` targets).

kanaka:
Plan sounds good.

dubek:
Status: all solved except scala.

I might try to convert the scala source tree to separate "projects" for each step (and one "lib" project for the common files like core/reader/printer/...), similar to Go. I have zero experience with sbt, so I'm mainly guessing.

dubek:
Update: I sorted out scala, now all impls are green. @kanaka , I understand if you want to wait with this merge until after the conference.

Scala now runs `sbt assembly` once; execution of specific tests are done via  a `run` wrapper script that simply runs java (doesn't need scala and sbt at this point). Also, sbt can prepend the shell script header to the jar file, so this task is now done in sbt instead of in the Makefile (for the `dist` target).

kanaka:
Looks good. The way scala is built/run still feels weird to me. I would like to figure out how to build separate jar files for each step at some point, but it's pretty low on my priority list. Maybe somebody will step up at some point to do a more idiomatic scala implementation and clean up the build process too.

dubek:
BTW - Can Clojure also be compiled to jar (per step) and then run it with `java -jar stepX_YYY.jar` ?

kanaka:
Yeah, that can be done. I'll take a look.

PR #213, opened 2016-05-12
cpp, erlang: time-ms fixes

dubek:
* C++: Use int64_t for MalInteger type so it can properly hold `(time-ms)` values
* Erlang: Fix (time-ms) to return milliseconds instead of seconds


dubek:
cpp & erlang travis jobs are green.

PR #215, opened 2016-06-07
Add Perl 6 implementation

hinrik:
All tests pass, but readline support (via Linenoise module) is commented
out in step0_repl.pl as it is not a core module. Should maybe change it
when docker support is added.


dubek:
Very nice @hinrik - thank you for the contribution!

Have you considered adding interop function, say something like `(perl6-eval "2 ** 5")` ? (assuming `**` still works in perl6, I have no idea...). It can be just another core function called `perl6-eval` which accepts a string. If it makes sense, you can also convert the perl6 return types back to Mal types (so if the perl6 code returned an array, it will be converted to Mal list, etc.).

hinrik:
Sure, I'll add that.

hinrik:
```
$ perl6 perl6/stepA_mal.pl 
Mal [perl6]
user> (perl6-eval "2 ** 5")
32
```

:)

dubek:
`MONKEY-SEE-NO-EVAL` :-) I learned something new today... Looks great. 

Few more comments:

1. Github reports a merge conflict. Can you try to solve it? I assume it's on Makefile and/or README.md.
2. README.md - update the number of total impls, and add your name and github profile to the description of the perl6 implementation (see example in the elisp description).

I'll leave the docker/linenoise decisions to @kanaka  (he has to push a new docker container anyway for the tests to pass).

hinrik:
There you go, I rebased it on top of master again.

Makefile
```diff
@@ -200,6 +201,129 @@ noop =
 SPACE = $(noop) $(noop)
 export FACTOR_ROOTS := .
 
+<<<<<<< HEAD
```
dubek: That didn't merge nicely... We now have `run` scripts in each impl dir instead of these `xxx_RUNSTEP` vars. Sorry for the mess...

hinrik: Yeah, sorry, just fixed the merge.

dubek: fixed, nevermind.

kanaka:
@hinrik can you suggest the best method for installing Rakudo Perl 2016.04 under Ubuntu Linux so that I can build it into a dockerfile for travis testing?

hinrik:
I'm not too familiar with Docker, but you can make it do this, right?

```
sudo apt-get install build-essential git
wget http://rakudo.org/downloads/star/rakudo-star-2016.04.tar.gz
tar zxvf rakudo-star-2016.04.tar.gz
cd rakudo-star-2016.04
perl Configure.pl --gen-moar --gen-nqp --backends=moar
make
make install
export PATH="$PATH:$PWD/install/bin"
```

kanaka:
@hinrik Yep, that's what I needed.

kanaka:
@hinrik okay, I pushed a docker image for rakudo 2016.04. Modify .travis.yml to add perl6 and push it out. that will trigger a travis build that includes the perl6 in the tests.

sdt:
If it helps, there's an official rakudo-star image: https://hub.docker.com/_/rakudo-star/

kanaka:
@hinrik I track down the reasons why travis tests aren't passing yet:

- the reason why the build isn't starting at all is because there is no implementation level Makefile. The travis build assumes there is one. Try adding this Makefile to your implementation directory:

```
all:
	@true
```

- once you have that, then the tests will fail because they assume there is a executable "run" script in the implementation directory. The following should suffice:

```
#!/bin/bash
exec perl6 $(dirname $0)/${STEP:-stepA_mal}.pl "${@}"
```

- Once those are there, step0 tests pass, but step1 tests fail. It appears that the reader is not correctly detecting closing delimiters for lists, vectors, hash-maps and strings.

- The step2 and step9 tests fail because errors aren't being caught. (i.e. evaluating an undefined symbol causes the REPL to stop)

- The implementation seems surprisingly slow. Slow enough that step5 fails the TCO test. This can be worked around by adjusting test settings in the top-level Makefile to wait for longer but this is usually only done for implementations that we expect to be especially slow such as self-hosted mode,  miniMAL, the SQL variants and vimscript. Perhaps the build I made of rakudo doesn't have the JIT enabled or maybe the implementation is doing something that is preventing the JIT for doing optimizations. @sdt you have any thoughts on that?

sdt:
I don't think the speed issues are related to hinrik's implementation or this particular rakudo build, rather rakudo itself.

Perl 6 is a huge language and my understanding is that rakudo is currently concerned with completeness and correctness rather than speed.

Right now it's just a bit slow.

hinrik:
Alright, I added the missing `Makefile`/`run` files and fixed the test failures. Also rewrote a little part of the printer that was causing the huge slowdown (apparently the implementation of currying is quite slow in Rakudo at the moment).

kanaka:
@hinrik looks like it's broken doing make perf^perl6, specifically the 3rd microbenchmark: ```env STEP=stepA_mal ../perl6/run ../tests/perf3.mal```

hinrik:
Yeah, just fixed that. `swap!` was setting the atom's value (e.g. to `(1 2 3 4 5 6 7 8 9 0)`) and then returning that value, which is what I interpreted from the process guide ("The new atom's value is returned." But as the perf test revealed, it's actually supposed to return the atom as a whole, containing the new value (`(atom (1 2 3 4 5 6 7 8 9 0))`).  :)

dubek:
That's very strange. I remember tests in step 6 actually expect the value
to be returned. I can look into it later.

hinrik:
Yeah, you're right, since step6 fails now. Must be something else broken in my implementation here...

hinrik:
The issue was a bug with `do`.

kanaka:
Looks good! Merged.

PR #217, opened 2016-06-12
Add Lua interop and describe interop in the guide

dubek:
First commit adds `lua-eval` core function and tests. Note that in Lua both "lists" and "hashmaps" are tables, so if the evaluated function returned an empty table we always return an empty Mal list. If it's not empty, we decide whether the table is a "list" or "hashmap" according to the first key - if it's numeric it'll be a Mal list, otherwise a Mal hashmap.

Second commit adds description of the optional `quux-eval` core function to the guide.

### DIFF
--- a/lua/core.lua
+++ b/lua/core.lua
@@ -5,7 +5,7 @@ local printer = require('printer')
 local readline = require('readline')
 local socket = require('socket')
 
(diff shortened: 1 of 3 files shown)


dubek:
The travis error is in the Io implementation - it hit the timeout in step5 tests.

kanaka:
Looks good. Merged.

PR #218, opened 2016-06-13
io: Add interop function io-eval

dubek:
Implements `io-eval` and provides test cases.


dubek:
Cancelled travis job and restarted only the IMPL=io build: passed OK: https://travis-ci.org/kanaka/mal/jobs/137145777

PR #220, opened 2016-06-13
tests: Verify case sensitivity of names and strings

dubek:
Encountered problems during self-hosting while implementing in Logo which is by default case insensitive. These tests should allow early discovery of such problems.


kanaka:
Good catch.

PR #222, opened 2016-06-15
Replace elisp* with elisp-eval

wasamasa:
While I was at it, I made elisp-eval a core function (as opposed to a special form) and made that one interop test more readable.

### DIFF
--- a/elisp/core.el
+++ b/elisp/core.el
@@ -217,4 +217,6 @@
                             mal-nil))
                          (t
                           mal-nil))))))
+
+    (elisp-eval . ,(mal-fn (lambda (string) (mal-string (format "%S" (eval (read (mal-value string))))))))
     ))
--- a/elisp/stepA_mal.el
+++ b/elisp/stepA_mal.el
@@ -113,8 +113,6 @@
             (throw 'return (mal-env-set env identifier value))))
          ((eq a0* 'macroexpand)
           (throw 'return (MACROEXPAND a1 env)))
-         ((eq a0* 'elisp*)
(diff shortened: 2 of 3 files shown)


wasamasa:
The failing build is from perl6.

elisp/core.el
```diff
@@ -217,4 +217,6 @@
                             mal-nil))
                          (t
                           mal-nil))))))
+
+    (elisp-eval . ,(mal-fn (lambda (string) (mal-string (format "%S" (eval (read (mal-value string))))))))
     ))
```
kanaka: What should really be happening here is that the resulting elisp data types are converted to equivalent mal data types instead of just returning a string representation. Maybe just removing the (mal-string (format)) would do the trick?

wasamasa: I'd like to have this as well, but it's problematic:

- `nil` and `()` are equivalent, therefore it's impossible to know which is meant on the MAL side
- loads of not representable types, such as buffers, bit arrays, markers, etc.

I recall complaining about it on `#mal` and getting the suggestion of returning a string with the printed representation instead, so that's why I'm doing this here.

kanaka: I went ahead and merged.

I would probably suggest returning nil for nil/(). For the other data types, you could convert as best you can, but fallback to the string representation for things that don't have any usable conversion.

dubek: I suggested returning a string as a starting point. Once that works, I suggest trying to convert the relevant types, as possible.

Lua suffers from a similar problem - an empty Lua table can be empty list or empty hashmap. In lua-eval I decided it's going to be an empty list.

PR #223, opened 2016-06-16
Improve elisp-eval to use MAL types

wasamasa:
Done as sugested in #222.


wasamasa:
As usual, perl6 spoils the party, err, CI.

PR #224, opened 2016-06-17
Logo implementation

dubek:
Tested on UCBLogo 6.0 with some minor tweaks (for performance and adding a native `timems` function).  The tweaks are performed during Docker image creation (see Dockerfile).

Tests of step 5 are skipped because UCBLogo is too slow.

Interop is available via `(logo-eval "logo code to run")`.

The `logo/examples` directory contains a Mal example of drawing a tree using turtle graphics.

So far I haven't had the patience of waiting for the entire self-hosted Mal tests. My guess is that tests like `(or false nil false nil false nil true)` will time out...

I managed to perform some other tweeks to UCBLogo which improve it's performance by 10-20% but we need something drastic to make this reasonable. I looked into NetLogo but: (a) haven't found a simple way for interactive input/output ; (b) it doesn't support all the keywords supported by UCBLogo, so it'll mean defining more macros (or modifying the code to use only the NetLogo keywords).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ matrix:
     - {env: IMPL=js,        services: [docker]}
     - {env: IMPL=julia,     services: [docker]}
     - {env: IMPL=kotlin,    services: [docker]}
+    - {env: IMPL=logo,      services: [docker]}
     - {env: IMPL=lua,       services: [docker]}
     - {env: IMPL=make,      services: [docker]}
     - {env: IMPL=mal BUILD_IMPL=js NO_PERF=1, services: [docker]}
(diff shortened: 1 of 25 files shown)


dubek:
Logo is now green: https://travis-ci.org/kanaka/mal/jobs/138470066

Performance, as mentioned, is really bad; this is from travis:

```
Performance test for logo:
Running: env STEP=stepA_mal ../logo/run ../tests/perf1.mal
"Elapsed time: 21145 msecs"
Running: env STEP=stepA_mal ../logo/run ../tests/perf2.mal
"Elapsed time: 59115 msecs"
Running: env STEP=stepA_mal ../logo/run ../tests/perf3.mal
iters/s: 0
```

PR #225, opened 2016-06-24
Crystal 0.18.4

rhysd:
I fixed Crystal implementation for Crystal 0.18.4.

I found that a test case for `time_ms` failed because it requires 64bit integer.  So I changed underlying type of `int` to `Int64`.  Now it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ coffee ./stepX_YYY
 
 *The Crystal implementation of mal was created by [Linda_pp](https://github.com/rhysd)*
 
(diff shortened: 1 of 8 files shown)


rhysd:
Crystal used in CI looks too old (version 0.10.0).  Could you tell me how to fix this?  Is Version of crystal fixed?

kanaka:
For now the docker images are under my docker hub account so I have to push them. I've rebuilt the image to pull in 0.18.4 and I've pushed it out. I restarted the test and now your changes pass.

rhysd:
@kanaka Thank you 😄

PR #227, opened 2016-07-26
swift3

fredim:
messing with the swift3 beta, and updated the code to work


kanaka:
@fredim thanks! Can you squash this into a single commit with a bit more specific commit message?

I'm not familiar with the recent changes in Swift 3, what's the purpose does the underscore in front of parameter needed for?

fredim:
addressed comments, closed for #228

PR #228, opened 2016-07-29
Squashed commits, updated to latest swift3 (swift-3.0-PREVIEW-2)

fredim:
Main changes:
* consistency of func arguments, while every argument has both an outer and an inner name,
  but the first argument's outer was "unnamed" by default in swift<2. now all arguments are consistent
  and requires the initial "_" to declare the outer "unnamed" for the first argument
* indexes are now simpler types, the Array.index function computes successor/predecessor
* many, many API changes, that result in shorter "verb" names of functions with named arguments
  ex: Array.joinWithSeparator(String) -> Array.joined(separator: String)

### DIFF
--- a/swift3/Sources/core.swift
+++ b/swift3/Sources/core.swift
@@ -6,7 +6,7 @@ import Glibc
 import Darwin
 #endif
 
-func IntOp(op: (Int, Int) -> Int, _ a: MalVal, _ b: MalVal) throws -> MalVal {
+func IntOp(_ op: (Int, Int) -> Int, _ a: MalVal, _ b: MalVal) throws -> MalVal {
     switch (a, b) {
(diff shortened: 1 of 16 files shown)


fredim:
another note, i'm not sure how the build environment works, but swift3 has to be installed...

kanaka:
Thanks. I updated the docker image to Swift 3 Preview 3 and merged your changes. I updated your changes slightly to remove the compile warnings due to unused return values.

kanaka:
Looks like both the swiftc (on Linux) and the Xcode 8 (on OSX) travis tests have passed, so we're good to go: https://travis-ci.org/kanaka/mal/builds/148775046

Thanks again for the update.

PR #229, opened 2016-08-06
ChucK implementation

wasamasa:
Manually run tests work, can't get it to function with the `run` script yet.


wasamasa:
I've added a hack to make it work.  All I need now is some sweet Docker integration...

FWIW, I've tested this on chuck 1.3.5.2.  Older versions of it may not have all required features, so it would be better to compile a newer release than provided on Ubuntu.

kanaka:
Merged!

PR #230, opened 2016-08-27
Add MAL implementation in clisp

iqbalansari:
Hi,

This implements MAL in [CLISP](https://www.gnu.org/software/clisp/) (an implementation of common lisp).

The performance is not upto the mark yet (something I intend to work on), however the implementation is functionally complete capable of self hosting

**EDIT** - The performance should be better now, though there is still room for improvement

Thanks


wasamasa:
I've tried it out locally and was surprised that performance is worse than with every other implementation using a Lisp dialect.  Any idea what the culprit might be?  Benchmarks suggest that clisp is slower than SBCL and CCL by a factor of about 5x, so I suspect it's rather the code's fault.

As an aside, could you eliminate the warnings displayed at startup?  While they go to stderr and are therefore not impacting tests, I'd rather not have them if it's avoidable :>

iqbalansari:
Hi,

Thanks for the feedback. Actually it was through you blog that I came across MAL so thanks for that too :smile: 

> I've tried it out locally and was surprised that performance is worse than with every other implementation using a Lisp dialect.

Exactly my reaction when I compared it to the rest of the implementations.

> Any idea what the culprit might be?

I did some profiling myself, I found the following
1) The tokenizer itself is slow, according to slime the function `tokenize` takes most time while reading even though the functions from the regexp package that the implementation uses are fast (it takes around 0.005 seconds to complete while corresponding regexp functions complete in 0.00005 seconds), so my only guess here is that my usage `loop` macro is somehow problematic.
2) I too found the `is-macro-call` call to be slow, I am not really sure about this though, is handling errors that expensive? Let me try that anyway
3) I also have the feeling using CLOS for MAL data types was not a good idea, `defstruct` would have been faster and (in this case) expressible enough

> The special forms comparison in mal-eval can be made significantly faster by comparing the symbols instead of creating objects over and over again.

That is what I started with and then realized that since common lisp has namespaces the symbols are interned to the `reader` package so a simple `equal` would not work.

Right now the `mal-value=` actually converts symbols to strings for comparison, which is ugly and something I want to fix, it might be better to keep symbols as string instead of `read`ing them

wasamasa:
Interesting, didn't realize the tokenizer could be problematic.  Yes, exceptions are a relatively expensive construct, even more so in CL as it has conditions.  As for qualified symbols, I've run into that with reader macro constructed objects.  The easy way out here is using `eq` on both qualified and unqualified symbols.

wasamasa:
Nice, now it's within the times of the other Lisp implementations.

kanaka:
Merged! I think basically just two things remaining. Update the top-level README implementation count and brief description of the implementation. And then tweet the official announcment (which I can do if you aren't on twitter or otherwise just prefer me to do it).

kanaka:
Also, if you use IRC at all and are interested in general mal discussion / announcements, then please join #mal on freenode.

iqbalansari:
> Update the top-level README implementation count and brief description of the implementation.

I have opened a pull request with last bit of changes.

> And then tweet the official announcment (which I can do if you aren't on twitter or otherwise just prefer me to do it). 
> ...
> Also, if you use IRC at all and are interested in general mal discussion / announcements, then please join #mal on freenode.

While I have a twitter account I am unfortunately not very active on it so probably you tweeting would reach a lot more people :smile:. Joined the IRC channel.

Thanks for MAL it has been a great learning experience.

@wasamasa Thanks for your help in debugging the performance issues

PR #232, opened 2016-09-10
Ada patches sep16

zmower:
Hi,

I recently received four PATCHes from Nicholas Boulenguez :
0001-Replace-all-GNAT-projects-with-D-gnatmake-option.patch - Removes project files and changes the Makefile to compensate.  Applied with some changes for stepA_mal.
0002-Delegate-obj-directory-creatinon-to-gnatmake-p-optio.patch - Doesn't work because of patch 0001!
0003-Delegate-Ada-dependency-handling-to-gnatmake-instead.patch - gnatmake works out its own dependencies so there's no need to repeat them in the Makefile.  Applied in spirit.
0004-Use-Get_Line-fonction-instead-of-insisting-on-a-maxi.patch - Removes explicit limit on string length.  Applied with some minor changes.

Both sets of tests pass.  I compared the stats before and after 0003.  They were the same.
Regards,
Chris

### DIFF
--- a/ada/Makefile
+++ b/ada/Makefile
@@ -1,36 +1,22 @@
 DIRS=obj
 PROGS=step0_repl step1_read_print step2_eval step3_env step4_if_fn_do \
-      step5_tco step6_file step7_quote step8_macros step9_try stepA_mal
+      step5_tco step6_file step7_quote step8_macros step9_try
 
-STEP0_DEPS=${DIRS}
-STEP1_DEPS=${STEP0_DEPS} types.ad[bs] types-vector.ad[bs] types-hash_map.ad[bs] \
-	   reader.ad[bs] printer.ad[bs] smart_pointers.ad[bs]
-STEP2_DEPS=${STEP1_DEPS}
-STEP3_DEPS=${STEP2_DEPS} envs.ad[bs] eval_callback.ads
-STEP4_DEPS=${STEP3_DEPS} core.ad[bs]
+SOURCES_LISP=envs.ad[bs] eval_callback.ads core.ad[bs] stepa_mal.adb
+SOURCES=$(SOURCES_LISP) \
+        types.ad[bs] types-vector.ad[bs] types-hash_map.ad[bs] \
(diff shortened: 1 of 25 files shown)


kanaka:
It's still important to specify the dependencies in the Makefile (even if the compiler can resolve dependencies) so that make itself knows when and which files need to be rebuilt. So patch 0003 is not correct. For example, if a developer modifies core.adb and then types `make all`, steps4-A should get rebuilt but steps0-3 should not because they do not depend on core.adb. Also, the use of the phony `force` dependency completely breaks dependencies such that every make call will rebuild the target(s) regardless of whether it is needed or not.

zmower:
Are we to disagree about computers expending unnecessary CPU cycles? Surely the Makefile is fit if it builds the software.  And since the dependencies are already specified in the source code, this makes it more maintainable.  I also think its idiomatic for Ada e.g. http://www.dwheeler.com/adacgi/adacgi-1.5/makefile

kanaka:
@zmower it's not an issue of CPU cycles, it's an issue of developer time (especially my time). Don't forget I have 59 implementations that I maintain and test. I try and be diligent to make sure that the Makefile for every implementation is not spending any extra time doing unnecessary build/dep steps. If every implementation spent a couple seconds doing a little extra (unnecessary) work to check dependencies, that can add many minutes to "simple" maintenance and testing tasks for me.

I'll take your word for this being a more idiomatic Ada build file. It's certainly not an idiomatic makefile style. Since gnat has a dep tool, the best way would actually be to generate dependency files (ala C .d files). That way you generate the deps with language specific tools that know how to figure out deps, but you are still letting make do the right thing (quickly) with regards to making the decision of what and when to build/rebuild.

Anyways, I've gone ahead and merged the change given that the 20X increase in build time is still only 0.2 seconds (compared to 0.01 seconds for make to figure it out). Note that if every implementation did this, it would take 12 seconds (instead of half a second) to check that nothing needs to be rebuilt.

PR #233, opened 2016-09-15
Update CLISP implementation

iqbalansari:
Hi,

I have pushed some updates to the CLISP implementation

1) Fixed a typo in core `/` function which caused some tests to fail
2) Change `env` implementation to use structs instead of classes

Thanks


kanaka:
Looks good. Merged!

PR #237, opened 2016-09-20
Compile guile files before running them

iqbalansari:
Hi,

I have enabled auto compilation of guile files, the files are store in the `guile/.cache` directory.

Thanks


kanaka:
Thanks! Merged.

Issue #226, opened 2016-07-15
haskell Type.hs ERROR


ruoshuiwyl:
Env :  opensuse 42.1 ghc 7.10.3
runhaskell step2,...,stepA 
Types.hs:64:1:
    Non type-variable argument
      in the constraint: Control.Monad.Error.Class.MonadError MalError m
    (Use FlexibleContexts to permit this)
    When checking that ‘throwStr’ has the inferred type
      throwStr :: forall (m :: * -> *) a.
                  Control.Monad.Error.Class.MonadError MalError m =>
                  String -> m a

Types.hs:65:1:
    Non type-variable argument
      in the constraint: Control.Monad.Error.Class.MonadError MalError m
    (Use FlexibleContexts to permit this)
    When checking that ‘throwMalVal’ has the inferred type
      throwMalVal :: forall (m :: * -> *) a.
                     Control.Monad.Error.Class.MonadError MalError m =>
                     MalVal -> m a

kanaka:
This is due to changes in ghc-7.10.1: http://stackoverflow.com/questions/28287329/ghc-7-10-non-type-variable-argument-use-flexiblecontexts-to-permit-this

You can get things to compile by adding `-XFlexibleContexts` to the ghc call in the makefile, however, there were also some refactorings to Error/Exception handling in 7.10.1 so you get lots of build warnings.

I think the correct fix for this issue is to refactor and clean up the error/except handling code in the mal implementation so that things compile cleanly without the compiler options. Ideally, 7.6 should also be supported since that's still seems pretty typical version in the wild.

ruoshuiwyl:
Thank you. 
I try to compile by addding  -XFlexibleContexts. 
It works.

kanaka:
@ruoshuiwyl I would be interesting in hearing if you are able to use the latest changes from @prt2121 without requiring the -XFlexibleContexts hack

ruoshuiwyl:
thank you. I try to it.

PR #238, opened 2016-10-02
support GHC 7.10.1+ : update error handling

prt2121:
i'm relatively new to Haskell but this should fix https://github.com/kanaka/mal/issues/226.


kanaka:
@prt2121 sorry for the slow response, I was on vacation until today. The changes look good. I pushed a new docker image with ghc 7.10.3 which gets the tests to pass.

kanaka:
I also just pushed an update the to Haskell Dockerfile and the top level README to indicate that ghc 7.10.1+ is now required.

prt2121:
thx 👍

PR #239, opened 2016-10-03
Picolisp

wasamasa:


wasamasa:
@kanaka I made all tests work in self-hosted mode.  Could you please add a Dockerfile and ping me if anything else needs work?

kanaka:
@wasamasa I created Dockerfile with picolisp 3.1.9.7 (I pushed it to docker hub as kanaka/mal-test-pil). When I test your implementation, I get some string errors though. Seems like regular "n" within strings is being converted to a carriage return. Maybe 3.1.9.7 has broken regex and you're using a newer version?

Example step1 test failure:
```
TEST: "abc (with parens)" -> ['',"abc (with parens)"] -> FAIL (line 67):
    Expected : '"abc (with parens)"\r\n"abc (with parens)"'
    Got      : '"abc (with parens)"\r\n"abc (with pare\\ns)"'
```

wasamasa:
The CHANGES file lists "Accept \n, \r, \t and \ddd\ in strings" as a 3.1.11 feature.  This doesn't sound terribly fun, but possible to work around.  I'd settle for building Picolisp from source instead, it's as simple as running `make` in the `src`, then the `src64` directory.  Globally installing it is a bit trickier (see [picolisp.install](https://aur.archlinux.org/cgit/aur.git/tree/picolisp.install?h=picolisp)), but not strictly needed as one could just extend PATH to point at the locally built thing.

kanaka:
Okay, I update the dockerfile to be based on Wily. That pulled in picolisp-3.1.11.1. All tests including self-host seem to pass now and I've pushed an update to the kanaka/mal-test-pil image so you should be able to add this to the Travis yaml file now for testing. Probably want to mention in the README that picolisp 3.1.11 or later is required.

Here is the Dockerfile (if you want to add it yourself):

```
FROM ubuntu:wily
MAINTAINER Joel Martin <github@martintribe.org>

# General requirements for testing or common across many
# implementations

RUN apt-get -y update

# Required for running tests
RUN apt-get -y install make python

# Some typical implementation and test requirements
RUN apt-get -y install curl libreadline-dev libedit-dev

RUN mkdir -p /mal
WORKDIR /mal

# picolisp
RUN apt-get -y install picolisp
```

kanaka:
And that makes 60!

dubek:
:fireworks:

PR #242, opened 2016-10-24
Emacs 25.1 fixes

wasamasa:


wasamasa:
See also https://debbugs.gnu.org/cgi/bugreport.cgi?bug=24786

PR #243, opened 2016-10-24
Ensure assoc updates maps properly

wasamasa:
I wrote this after getting bitten by this when running step 3 in self-hosted mode.

### DIFF
--- a/tests/step9_try.mal
+++ b/tests/step9_try.mal
@@ -260,6 +260,15 @@
 (keyword? (nth (vals {"a" :abc "b" :def}) 0))
 ;=>true
 
+;; Testing whether assoc updates properly
+(def! hm4 (assoc {:a 1 :b 2} :a 3 :c 1))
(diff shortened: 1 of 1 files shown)


wasamasa:
Hm, perhaps the tests are not ideal.  The first test is misleading as the resulting map does not need to be deduplicated, as long as the lookup results are correct (see the Forth implementation for an example).  The second test is not needed and more of a sanity check that most, but not all implementations pass (see Forth and PL/pgSQL).

dubek:
I thought the problematic scenario was:

```
(def! hm4 (assoc {:a 1 :b 2} :a 3 :a 7))
(get hm4 :a)
;=>7
```
?

wasamasa:
@dubek The MAL implementation does never use duplicate keys in the rest argument of `assoc`, so that's not a problematic scenario for the self-hosting case.

PR #244, opened 2016-10-25
Allow web exec of a .mal file via symlinked php script.

chr15m:
To test this:

    cd php
    make mal.php
    echo '(prn "Hello world!")' > myscript.mal
    ln -s  mal.php myscript.php
    php -S 0.0.0.0:8000

Then browse to http://localhost:8000/myscript.php and you should see "Hello world!" in your browser as the .mal script is run.

### DIFF
--- a/php/README.md
+++ b/php/README.md
@@ -1,0 +1,15 @@
+### Running .mal scripts on PHP hosting ###
+
+Create a symlink to `mal.php` with the same name as your `.mal` script and your script will be executed as if it was PHP.
+
+Here's an example using local dev:
+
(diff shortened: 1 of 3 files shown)


kanaka:
@chr15m before I merge, can you add some documentation for this. A README.md in the php directory with the example from you comment would be sufficient.

chr15m:
@kanaka thanks for reviewing! I've pushed a README.md to my branch - let me know if it is ok or needs editing.

kanaka:
Thanks! Merged.

PR #245, opened 2016-10-26
Vimscript: Use Vim 8.0

dubek:
Changes:

* Dockerfile now installs Vim 8.0 (from source)
* Use lambdas for most functions in `core.vim` (this removed a lot of boilerplate code!)
* Use map+lambdas instead of `for` loops in several places
* Get rid of useless functions like ObjValue and ObjType to somewhat improve performance
* Fix bug in `time-ms` returning negative numbers

Not sure if there's any other improvement (Vim 8.0 has jobs, channels and JSON support - but not sure how these can help mal).

### DIFF
--- a/README.md
+++ b/README.md
@@ -814,8 +814,7 @@ make
 
 *The Vimscript implementation was created by [Dov Murik](https://github.com/dubek)*
 
-The Vimscript implementation of mal requires Vim to run.  It has been tested
-with Vim 7.4.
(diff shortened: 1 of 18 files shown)


dubek:
I cancelled the travis build - first a new docker image should be pushed.

kanaka:
I pushed out an updated Dockerfile and restarted the test and it passed. Merged.

PR #247, opened 2016-10-29
Php interop extended

chr15m:
Tighter PHP interop using `$` and `!` fns.
`($ _SERVER)` returns a mal hash-map holding the contents of the `$_SERVER` superglobal from PHP.
`(! file_get_contents "https://github.com")` returns a string containing the github homepage.
`file_get_contents` is a native PHP function.


chr15m:
The failing plpgsql test here seems unrelated to this patch - let me know if you think otherwise!

PR #250, opened 2016-11-20
Add Skew implementation

dubek:
See http://skew-lang.org/ for details on the Skew language. Currently Mal only compiles to Javascript, as there are some issues with the C# backend for Skew (https://github.com/evanw/skew/issues/19).

Tested with Skew 0.7.42.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,7 @@ r/lib
 scala/mal.jar
 scala/target
 scala/project
+skew/*.js
 tcl/mal.tcl
 vb/*.exe
 vb/*.dll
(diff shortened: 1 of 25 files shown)


dubek:
hmm, I just learned that Skew has an (undocumented?) `is` operator for `a is MalString` etc. Which means I can get rid of the `type` field and enum. I'll try to improve this.

dubek:
OK I re-pushed 034e82a which discards my homemade `type` field and relies on Skew's `x is Y` construct.

After Docker image is built and pushed, we can retry travis on `IMPL=skew` .

kanaka:
Built and pushed docker image. Tests now pass. Merged. I'll watch for your tweet ;-)

dubek:
@kanaka: https://twitter.com/dubek/status/800783541034422272

PR #254, opened 2017-02-06
Go: Use type assertions where possible instead of reflection

cosmos72:
It seems a bit more idiomatic to me - even I am just a beginner in Go.
Also, make "test^go" seems marginally faster (0.56s instead of 0.57s on my system)


kanaka:
Yes, that's a good improvement. Thanks.

PR #255, opened 2017-02-11
Fixed erroneous fn* -> native PHP function conversion.

chr15m:
Added test for callbacks during interop which is where this failure showed up.


kanaka:
Thanks.

PR #257, opened 2017-04-12
Fix guide's explanation of Regex

boynedmaster:
it didn't incude the parantheses


kanaka:
Thanks for the catch! Merged.

boynedmaster:
yea, it messed me up bc the language i'm using doesn't support regex so i had to remake it myself based on your explanations

PR #267, opened 2017-07-20
Add Rexx implementation

dubek:
* Tested with Regina Rexx 3.6 (`apt-get install cpp regina-rexx`).
* Uses cpp to preprocess files so we'll have `#include` capabilities.
* No garbage-collection.
* step5 tests without TCO crash with segmentation fault (so I left them out).
* `rexx-eval` evaluates several statements and returns the value of the last one (which must be an expression, not a statement). See `rexx/tests/stepA_mal.mal` for examples.

Hopefully someday I'll be able to test it on a real mainframe... :-)


dubek:
@kanaka I cancelled the travis build. After you push the docker container, you can start the specific `rexx` job.

kanaka:
Done. And it passes.

dubek:
https://twitter.com/dubek/status/888044584261038081

PR #270, opened 2017-07-22
Elm implementation

c0deaddict:
It took some time, but here it is!

Elm's immutable data structures were a pain in the ass for implementing the environment. I ended up using integers as "pointers" to frames in the environment. Had to write a small garbage collector to clean up unused frames. Performance of recursion and macros is not great, but it works :-)

Another thing that took some figuring out was how to achieve side effects (like print). Elm is side effect free except for the update loop which can tell the Elm runtime to execute a  side effect (Cmd in Elm terms). When executing a side effect from Mal code, the evaluator uses continuations to restore the caller execution context after the side effect has been run.


dubek:
Impressive!

kanaka:
@c0deaddict I'll try and take look sometime later today. In the meantime, can you resolve the Makefile conflict?

kanaka:
@c0deaddict I've built and pushed a docker image so once the conflict is resolved, Travis should automatically build/test the elm implementation.

In terms of the implementation itself ... wow. That is much larger than I was expecting. Sometimes the more verbose implementations (especially for modern languages) indicate that a language is "non-Lispy". For example, the ruby implementation is very small. Even though ruby's syntax does not resemble Lisp, the language itself feels very much like a Lisp (available constructs, core library, etc). Do you have a general sense as to why the Elm implementation is so large?

The implementation is not as incremental as I generally prefer. It's nothing that would keep from merging it, but I wonder how difficult it would be to make it a bit more incremental? Specifically I'm think of the early use of the Env (in step1 and step2). If it would require major refactoring, don't worry about it, but if you could think of some minor tweaks to avoid that, it would be a better fit to the incremental (learn as you go) nature of most implementations.

c0deaddict:
@kanaka Conflict is resolved, Travis is running.

Elm definitely is non-Lispy. At first I thought implementing Mal in a strong statically typed language would be the hardest part, but that wasn't the case. What made the implementation hard to do is the purity of Elm. No function can have side effects and all values are immutable. I couldn't find a way to get something like IORef in Haskell (in older Elm versions there was [this](http://package.elm-lang.org/packages/karldray/elm-ref/latest), but that isn't supported anymore).

Implementing the Env without mutable values/references was hard. I'm not proud of the solution. It feels backwards using integers as "soft pointers" to (parent) frames in a Dict. Also a garbage collector was needed because Elm's (actually Javascript's) GC can't determine which frames are used/references and which are not.

Side effects we're easier to implement but also introduced some accidental complexity. Being a pure language, Elm doesn't have an IO monad like Haskell. The only way of producing a side effect is passing some data out of the loop function (it's called "update").

Bottom line: mutable values and side effects are really useful when writing an interpreter.

I'll have a look if i can remove the use of Env from step1 and step2.

kanaka:
@c0deaddict The build was running into permissions issues when running under Travis (permissions a re a bit different in that context). I was able to get past the permissions issues by adding the following to the end of the Dockerfile:

```
ENV HOME /mal
ENV NPM_CONFIG_CACHE /mal/.npm
```

However, now the following problem occurs:

```
node_modules/.bin/elm-package install -y
Starting downloads...

  ✗ elm-lang/core 5.1.1
  ✗ Bogdanp/elm-combine 3.1.1
  ✗ elm-lang/lazy 2.0.0
[91mError: The following HTTP request failed.
<https://github.com/Bogdanp/elm-combine/zipball/3.1.1/>

FailedConnectionException2 "github.com" 443 True getProtocolByName: does not exist (no such protocol name: tcp)

Makefile:18: recipe for target 'elm_packages' failed
make: *** [elm_packages] Error 1
Makefile:305: recipe for target 'elm/step0_repl.js' failed
make: *** [elm/step0_repl.js] Error 2
```

This happens locally as well. You should be able to replicate the problem like this (assuming you have docker and have pulled down the kanaka/mal-test-elm image:

```
make DOCKERIZE=1 "test^elm^step0"
```

c0deaddict:
@kanaka I don't know why, but apparently the netbase package is needed. I tried it locally and the build works. Can you build and push a new docker image?

kanaka:
Merged! Congrats on your second implementation. Keep them coming :-)

I made a minor modification to your Dockerfile to move netbase lower down (more specific deps lower makes rebuilds faster).

Please do try and make the incremental changes if you have a chance.

Also, would you like to tweet to announce this? Or perhaps blog about the experience and then tweet it :-) If you do write it up, just be warned that you might get critique (hopefully constructive) from the Elm community.

kanaka:
Also, please feel welcome to join #mal on freenode. It's fairly quiet with occasional bursts of mal and Lisp related conversation.

c0deaddict:
Thanks! It was a pleasure making this one :-)

I've made the incremental changes, i'll open a up a new pull request for it.

This time i'll try to write a blog post. Haven't done that before, but i'll give it a shot.

c0deaddict:
Blog post is finished: https://medium.com/@c0deaddict/make-a-lisp-in-elm-7945c017524e

PR #271, opened 2017-07-25
Elm: Removed usage of Env from step1 and step2

c0deaddict:
- Removed the (direct) import of Env in step1 and step2.
- Simplified use of Eval in the first 3 steps.
- Cleaned up some debugging code in Printer.

### DIFF
--- a/elm/Eval.elm
+++ b/elm/Eval.elm
@@ -223,3 +223,16 @@ inGlobal body =
                 else
                     body
             )
+
+
+runSimple : Eval a -> Result MalExpr a
+runSimple e =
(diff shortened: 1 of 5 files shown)


kanaka:
Thanks!

kanaka:
BTW, any chance you could use direct JS interop for environment mutation?

c0deaddict:
Good point! I considered using Elm ports for the environment but decided that was too much hassle (ports are treated the same in Elm as side effects). But I haven't looked at Elm Native modules. Maybe that can work. I'll have a look that.

PR #273, opened 2017-09-10
R7RS implementation

wasamasa:
This was pretty fun, I've handed in a dozen bug reports over the course of this, nearly all of which have been resolved in a timely manner.  Now comes the hard part for me, dockerization.  I've ensured that all tests (save one optional one with metadata in native functions) pass, including self-hosted mode.  How exactly shall I proceed regarding test runs, do I need to write a docker file that installs all Scheme implementations and teach Travis to do a run for each?


kanaka:
In terms of Dockerization and testing, for now I think everything in one docker image is probably the least intrusive to the overall build/test process. You can then specify the mode environment variable for the different modes in separate travis phases. High-level support for sub-implementations can be tackled separately.

Would "scheme" be a better directory name for this? SCM seems to be the name of a specific implementation of R4RS/R5RS scheme (https://en.wikipedia.org/wiki/SCM_(Scheme_implementation)) so it seems a bit misleading (and doesn't communicate the scope of what you've accomplished either). I've personally been a bit inconsistent about using file extension versus full name for the directory name, but I would lean towards full-name (and most do follow that) unless its universally recognized by the extension (like vb, tcl, js).

wasamasa:
Alright, I'll try to get everything built in one image.  About the name, I've considered both "scheme" (somewhat generic, one might assume R6RS) and "r7rs" (doesn't convey the language to lesser experienced people), but didn't consider "scm" being problematic.

kanaka:
Okay, no problem. Maybe adding an "(scm/R7RS)" parenthetical to README?

kanaka:
@wasamasa BTW, I just pushed a change to make MODE setting more generic. It's now <impl>_MODE and when you add <impl>_MODE variable to the top-level Makefile it should be conveyed automatically to the run script and implementation specific Makefile without having to modify anything else in the top-level Makefile.

wasamasa:
Great, I've changed it so that running `make` on its own will set up the necessary dependencies depending on what `scheme_MODE` is set to.   That leaves the Dockerfile, I'm afraid none of the Scheme implementations except for CHICKEN is in any Ubuntu repository and I'm not sure whether their version will even work.

kanaka:
@wasamasa Pulling from distro repository is certainly best, but it's acceptable to pull binaries directly from the project site, or if there are no builds, then build them from sources in the Dockerfile. Basically, try and replicate how you got them installed in the first place but in the Dockerfile. Also, it's worth searching for existing docker images for the implementations in question so you can reference the steps from their Dockerfile.

wasamasa:
The PR should be good to go now, all that's missing is pushing the Docker image and retrying the relevant builds.

kanaka:
Okay, I pushed the docker image and restarted the scheme implementations. I'll merge once/if they pass. 

BTW, can we set the default mode to one from the list so that `make test^scheme` works without requiring the mode to be set?

wasamasa:
That should already be the case with Chibi, I've set it in several places as the default.

kanaka:
Looks like foment times out during the perf test: https://travis-ci.org/kanaka/mal/jobs/275101258

wasamasa:
This is weird because I cannot reproduce this with `scheme_MODE=foment DOCKERIZE=1 make "perf^scheme"`.  Foment is be roughly at the same speed as Chibi so the timeout isn't inherent to the implementation.

kanaka:
Huh, yes, it appears to be something specific to the Travis environment. Running the same command locally works fine for me too:

```bash
docker run -it -u 2000 -v `pwd`:/mal kanaka/mal-test-scheme make 'TEST_OPTS=--debug-file ../perf.err' MAL_IMPL=js scheme_MODE=foment 'perf^scheme'
```

That's pretty weird since the dockerization should remove almost all software differences. That leaves: hardware, kernel, docker engine, environment/permissions. Anything about foment that might be sensitive to hardware features or the kernel version?

kanaka:
BTW, if you want to disable just that mode in .travis.yml, I'm fine with merging it. Or it can wait until you debug foment. Either way is fine with me.

wasamasa:
The only thing I can think of is its setup for stdin/stdout where it tries detecting a tty.  Perhaps that check fails and leaves it in a state where the rest of the rig doesn't work either.  I'm fine with deactivating it for the time being and reactivating in CI when the source of the bug has been discovered and fixed.

kanaka:
Okay, the 6 remaining have passed. Merged.

wasamasa:
Blog post: http://emacsninja.com/posts/x-rrs.html

PR #274, opened 2017-09-12
Fix for PHP 7 compatibility 

Poeticode:
This PR renames the Error class to _Error, so that it doesn't interfere with PHP 7's own internal Error class. Mal is otherwise running just fine with PHP 7.0.

### DIFF
--- a/php/core.php
+++ b/php/core.php
@@ -6,7 +6,7 @@ require_once 'reader.php';
 require_once 'printer.php';
 
 // Error/Exception functions
-function mal_throw($obj) { throw new Error($obj); }
+function mal_throw($obj) { throw new _Error($obj); }
 
 
 // String functions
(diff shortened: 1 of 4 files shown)


dubek:
Thanks for that fix @Poeticode . The travis failure is in another implementation (haxe-cpp), not sure exactly why. The PHP impl passes all tests.

Poeticode:
Huh, that is really odd. I only renamed all occurrences of Error in php/*.php to _Error. No clue how that could affect the test in `/haxe/tests/step5_tco.mal`.

dubek:
No worries, it's not your fault; as long as the PHP implementation passes the tests, you're clear. Some of the travis builds are unstable (we used to have a problem with rpython builds).

kanaka:
@Poeticode the Haxe breakage was me. I've been activating some sub-implementations in Travis tests and Haxe's cpp target doesn't pass step5. It should be fixed now. And I've merged your changes. Thanks!

PR #276, opened 2017-09-12
Update core.py

ideasman42:
get() returns None when not found


kanaka:
This breaks (get nil "a"): https://travis-ci.org/kanaka/mal/jobs/274743121#L1452

Also hash-map comparison: https://travis-ci.org/kanaka/mal/jobs/274743121#L1515

ideasman42:
Corrected

kanaka:
Looks good. Merged.

PR #285, opened 2017-09-17
Add struct

zaoqi:


wasamasa:
Why are you using `define` (this is Mal, not Scheme)? Also, what exactly is the benefit of adding this to `core.mal`?

@kanaka What exactly is the job of `core.mal`, other than providing arrow macros for macro tests?

kanaka:
@zaoqi I would be willing to accept a pull request with the struct functionality if you fix it to be mal syntax and move it to examples/struct.mal. Also, please document how to use it (probably just in comments in the same file).

@wasamasa the original purpose was that `core.mal` would be the mal implemented core library. Whereas `IMPL/core.EXT` is the part of the core library implemented in the target language. However, my intention at some point is to add a `lib/` directory and move some of these functions and macros into logically grouped files under `lib/` (along with a some of the files from `examples/`).

PR #292, opened 2017-10-01
Another Java impl, just require jdk. 

guxingke:


kanaka:
@guxingke can you explain what is unique about this compared to the original Java implementation that would make it interesting of merging into the main tree (apart from using javac directly rather than maven)? If it's a re-implementation that is fairly similar to the original, I am happy to link to your repo (on the README) as an alternate implementation. Alternately, if this implementation is fairly similar, but superior in some way (e.g. more idiomatic code, more concise, faster runtime performance, better language interop, etc) I would be happy to replace the existing implementation I did in Java with this one. But you will need to demonstrate/explain how that is the case.

Some other general comments:
- You have a binary file in the pull request (jline jar). Binary files should not be stored in git. The preferred way to resolve dependencies is with the typical dependency resolution tools for the language in question. I.e. for java this is often done via maven. However, since not using maven seems to be part of the point of your implementation, having the Makefile download the package as part of a build rule is also acceptable.
- The commit history is pretty messy with a lot of unnecessary merges. Before I would merge this upstream the history would need to be cleaned up.

guxingke:
@kanaka thanks for your comments, I am in holiday now, the detail message will be offer after some days, thanks.

java2/Makefile
```diff
@@ -1,5 +1,12 @@
 %.class: src/mal/
 	@rm -rf $*
+	@test -d bin || mkdir -p bin 
+	@if [ ! -f bin/jline-2.13.jar ]; \
```
wasamasa: This could be a target/prerequisite to another one.

java2/Makefile
```diff
@@ -1,5 +1,12 @@
...
+		then \
+		echo "downloading jline-2.13.jar form maven repo";\
+	  wget -O bin/jline-2.13.jar "http://central.maven.org/maven2/jline/jline/2.13/jline-2.13.jar" ;\
```
wasamasa: Why use `wget` if you've installed `curl` in the Dockerfile?

guxingke: thanks,  it fixed.

java2/Makefile
guxingke: thanks,  it fixed.

guxingke:
@kanaka 
(Sorry, my English is not very good.)

executable, easy to running,  just require jdk8, jline2 is optional for console inline edit . (Another one depends on maven2, I am try run it but failed, it unfriendly for learning) 
simplly, no magic, no complex abstract, the code is simple and easy to understand .
 readable,  more idioms, exact naming. 

in short words, it friendly for trial, in-depth understanding. 
if merge this, I am glad to operate and upgrade this impl.

guxingke:
@kanaka 
Excuse me, Can you give me some message about this PR?

PR #294, opened 2017-10-05
Update guide.md

zaoqi:


kanaka:
The basic IMPL-eval form is specifically for evaluating a string that contains target language source code.

PR #297, opened 2017-10-07
Create numberorprocedure.mal

zaoqi:


wasamasa:
This doesn't make much sense.  It would be better to change step A to provide all relevant predicates and patch all implementations to do the same.

PR #298, opened 2017-10-07
Add number?, fn? and macro? predicates

zaoqi:
Here is a list of all the implementations that are tested by Travis:

- [x] ada
- [x] awk
- [x] bash
- [x] basic
- [x] c
- [x] cpp
- [x] coffee
- [x] cs
- [x] chuck
- [x] clojure
- [x] common-lisp
- [x] crystal
- [x] d
- [x] dart
- [x] elisp
- [x] elixir
- [x] elm
- [x] erlang
- [x] es6
- [x] factor
- [x] forth
- [x] fsharp
- [x] go
- [x] groovy
- [x] gst
- [x] guile
- [x] haskell
- [x] haxe
- [x] hy
- [x] io
- [x] java
- [x] js
- [x] julia
- [x] kotlin
- [x] livescript
- [x] logo
- [x] lua
- [x] make
- [x] mal
- [x] matlab
- [x] miniMAL
- [x] nim
- [x] objpascal
- [x] objc
- [x] ocaml
- [x] perl
- [x] perl6
- [x] php
- [x] pil
- [x] plpgsql
- [x] ps
- [x] powershell
- [x] python
- [x] r
- [x] racket
- [x] rexx
- [x] rpython
- [x] ruby
- [x] rust
- [x] scala
- [x] scheme
- [x] skew
- [x] swift
- [x] swift3
- [x] tcl
- [x] ts
- [x] vb
- [x] vhdl
- [x] vimscript
- [x] yorick

### DIFF
--- a/process/guide.md
+++ b/process/guide.md
@@ -1635,7 +1635,7 @@ For extra information read [Peter Seibel's thorough discussion about
     new vector is returned with the elements added to the end of the
     given vector.
   * `string?`: returns true if the parameter is a string.
-  * `procedure?`: returns true if the parameter is a function.
+  * `fn?`: returns true if the parameter is a function.
   * `number?`: returns true if the parameter is a number.
   * `seq`: takes a list, vector, string, or nil. If an empty list,
     empty vector, or empty string ("") is passed in then nil is
--- a/process/guide.md
+++ b/process/guide.md
@@ -1635,6 +1635,8 @@ For extra information read [Peter Seibel's thorough discussion about
     new vector is returned with the elements added to the end of the
     given vector.
   * `string?`: returns true if the parameter is a string.
+  * `procedure?`: returns true if the parameter is a function.
+  * `number?`: returns true if the parameter is a number.
   * `seq`: takes a list, vector, string, or nil. If an empty list,
     empty vector, or empty string ("") is passed in then nil is
     returned. Otherwise, a list is returned unchanged, a vector is


dubek:
Things to consider in case we want to add these to the language:

1. Should be added for all 70+ implementations.
2. Clojure has `fn?` and `ifn?` -- I'd choose `fn?` instead of `procedure?`
3. Do we want to distinguish internal functions from user-defined ones?
4. What about macros?

wasamasa:
1. Yes, obviously, this could be done the same way as with the previous `(seq "\\n")` one.
2. `fn?` sounds good to me.
3. I'd treat them the same, there isn't really a benefit or is there?
4. Macros being functions with a macro flag is an implementation detail, so a `macro?` predicate might be warranted.  It looks like with this addition all observable types have a predicate.

kanaka:
I like `fn?`. Since macros are first class in mal we should also have a `macro?` predicate for completeness. I agree that we can do this like we did the `(seq "\\n")` addition. I think these should be optional tests, however, for visibility in Travis, lets add the tests in the required section of the stepA tests and then once we have all implementations passing, we can move the test definition into the optional section (the optional tests don't cause a top-level visible failure in Travis).

I have less free time than normal this week. @dubek or @wasamasa do you want to create a branch with the initial test addition (you both should have ability to create it in the main repo but let me know if you have trouble)? We can then commit against that. And @zaoqi, if you want to participate in fixing this in implementations, you can send pull requests against that branch once it is created.

dubek:
I added the `add-predicates` branch with those new tests and guide changes. Then I implemented them for Ruby for now. You're welcome to push changes there or build PRs on top of that branch.

I'll also be away starting Wednesday, so feel free to take over.

kanaka:
Thanks. Here is the current Travis status: https://travis-ci.org/kanaka/mal/builds/285770799 Once it completes, I'll add the list of TODO implementations here.

dubek:
I just pushed fix for go, rexx, vimscript, yorick.

wasamasa:
I'm claiming my own implementations: elisp, chuck, picolisp, gst, scheme.

kanaka:
@dubek @wasamasa I'd probably be more inclined to have `fn?` return true even for macro functions. There isn't really a direct parallel here in Clojure since Clojure doesn't have first class macros. But I think of  macros in mal as functions that get unevaluated arguments. I think this is going to be easier to implement across most implementations (although that's not my main reason) since most implement macros as an attribute on functions. Note that `(and (fn? x) (not (macro? x)))` will still answer the question "is this a non-macro function". I'm willing to be convinced otherwise especially if there some precedent in other Lisps (i.e. predicates on f-exprs or something).

wasamasa:
@kanaka The only Lisp with fexprs I've dealt with is Picolisp which encodes this fact in the argument list.  You only have a `fun?` predicate there and if you want to you can do pattern matching on its result to check whether you're dealing with a fexpr.

dubek:
I think it's clearer if `fn?` returns false for macros. I also agree it's a bit messier to implement it this way (for most implementations), but there's no rush. I don't know what would be the "right" decision in terms of compatibility with other lisps/languages.

kanaka:
Okay, we'll stick with `fn?` returning false for macros.

I'll take bash and basic.

zaoqi-unsafe:
Is macros value?

wasamasa:
Claiming js, lua, perl, python and racket.

@zaoqi-unsafe Yes.

kanaka:
I fixed CoffeeScript, Clojure and C#

dubek:
I fixed forth, skew, tcl.

I changed the user-defined function checks to check my own `add1` function instead of `not`, because some implementations (Forth) define `not` as a native function.

dubek:
Fixed Awk and OCaml.

wasamasa:
I tried adjusting the MAL implementation in MAL itself and noticed that you can't just add `fn?` and `macro?` to the array of existing functions as the implementation-provided one cannot possibly detect how the self-hosting code set the macro flag.  Is it worth it to work around that problem or should I just omit them?

dubek:
@wasamasa I think Mal sets the "ismacro" key in the function's meta (see `defmacro!` and `is-macro-call` in `stepA_mal.mal`). Shouldn't `fn?` and `macro?` perform a similar check?

dubek:
Fixed Crystal and Logo.

dubek:
Fixed Postscript and Rust.

wasamasa:
Claiming common-lisp, es6, guile, java, mal.

dubek:
Fixed D and Nim.

dubek:
Fixed Io, Swift3, VB.

dubek:
Fixed F#, LiveScript, TypeScript.

wasamasa:
Claiming hy, php, powershell, scala.

dubek:
Claiming make, objc, perl6, r .

kanaka:
There is a issue in how make does macros (it stores the macro property in the environment instead of associated with the function value itself). I've already started looking into fixing that. So unless you've already made significantly progress with make, I can take that.

dubek:
Yes, I just noticed that. I finished the other three (not yet pushed). @kanaka - please take Make.

dubek:
Fixed Objective C, Perl6, R.

kanaka:
I'll take make and miniMAL.

kanaka:
Fix make and miniMAL.

dubek:
Fixed Factor.

kanaka:
Fixed Dart, Groovy, Haskell, Haxe, Matlab

dubek:
Claiming Elixir and Erlang.

dubek:
Fixed Elixir, Erlang, Julia, Kotlin.

wasamasa:
Fixed elm, objpascal, plpgsql.  That leaves ada (completely incomprehensible and verbose code) and swift (impossible to test for me).

dubek:
Fixed Ada.

kanaka:
Fixed Swift. Most of the methods were already there and just needed to be exposed. Edit locally, build/test on Travis ;-)

kanaka:
Resolved and merged here: https://github.com/kanaka/mal/pull/306

Thank you @zaoqi, this was a good suggestion. Thanks @dubek and @wasamasa for the fixes.

PR #300, opened 2017-10-09
Elisp fixes

wasamasa:
I've remembered there was still one TODO left.  Speaking of which, is it OK to do a shallow copy for `with-meta` or could that be problematic for the atom type?  If I understand correctly this is about ensuring that the meta flag is independent for objects, even if their other properties are structurally shared.


dubek:
I checked Clojure and it doesn't allow adding meta to atoms. So my intuition would be that for Mal it's unspecified and therefore you can do shallow copy if you want.

kanaka:
I can't think of any instances where a shallow copy would be a problem.

PR #301, opened 2017-10-09
C++ implementations of fn?/macro?/number?

sdt:
Re: #298 

Also comes with a bonus Makefile fix to make macOS builds not depend on the version of readline installed by homebrew


wasamasa:
I wouldn't merge yet until there has been a decision whether `macro?` and `fn?` should be disjoint types or not.  See https://github.com/kanaka/mal/pull/298#issuecomment-335298522.

sdt:
Sorry, I didn't realize it was still undecided.

@wasamasa I can close this PR for now if you want.

wasamasa:
Apparently we're lucky as the stricter interpretation has been chosen, so no need to close the PR.

PR #307, opened 2017-11-02
vhdl: Fix time-ms: Return milliseconds since last midnight UTC to fit in 32-bit integer

dubek:
We used to return milliseconds since 2000 but that number is too large to fit in a signed 32-bit integer.  Instead we now return the number of milliseconds since last midnight UTC, which is between 0 and 86399999 (still fits in a 32-bit signed integer).

Also removed unused variables from `define_core_function`.

### DIFF
--- a/vhdl/core.vhdl
+++ b/vhdl/core.vhdl
@@ -226,17 +226,16 @@ package body core is
     assert false severity failure;
   end function gettimeofday;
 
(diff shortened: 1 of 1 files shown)


dubek:
The VHDL travis job is green: https://travis-ci.org/kanaka/mal/jobs/296182530

PR #310, opened 2017-12-28
NASM x86-64 assembly implementation

bendudson:
Thank you for putting together this great guide. I had a lot of fun. Perhaps I wouldn't have started if I'd known how much was involved, but glad I did.  

I'm not sure how idiomatic this code is, as it was a learning experience, but I hope it is not to odd. The code is written for Linux, though the only Linux-specific parts are in `system.asm` (233 lines, 8 syscalls). The rest is quite self-contained, with no external dependencies. Includes memory management, string manipulation etc. in `types.asm`. The other non-standard file is `exceptions.asm` which has a simple exception handling implementation.

I added three tests, to catch issues which appeared in self-hosting. These were the ones which seemed like they might be more general; most of the bugs I found were specific to this implementation, particularly the reference counting.

The Docker image builds and tests pass, but I get permission denied when running `docker push kanaka/mal-test-nasm`. I've never used Docker before though, so may be doing something wrong. The Travis script therefore fails.

### DIFF
--- a/nasm/Dockerfile
+++ b/nasm/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu
+FROM ubuntu:xenial
 MAINTAINER Joel Martin <github@martintribe.org>
 
 ##########################################################
@@ -21,6 +21,6 @@ WORKDIR /mal
 # Specific implementation requirements
 ##########################################################
 
-# Install nasm
-RUN apt-get -y install nasm
+# Install nasm and ld
+RUN apt-get -y install nasm binutils
 
--- a/nasm/Dockerfile
+++ b/nasm/Dockerfile
@@ -1,0 +1,26 @@
+FROM ubuntu
+MAINTAINER Joel Martin <github@martintribe.org>
+
+##########################################################
+# General requirements for testing or common across many
+# implementations
+##########################################################
+
+RUN apt-get -y update
+
+# Required for running tests
+RUN apt-get -y install make python
+
+# Some typical implementation and test requirements
+RUN apt-get -y install curl libreadline-dev libedit-dev
+
+RUN mkdir -p /mal
+WORKDIR /mal
+
+##########################################################
+# Specific implementation requirements
+##########################################################
+
+# Install nasm
+RUN apt-get -y install nasm
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,7 @@ matrix:
     - {env: IMPL=mal BUILD_IMPL=js NO_PERF=1, services: [docker]}
     - {env: IMPL=matlab, services: [docker]} # Uses Octave
     - {env: IMPL=miniMAL,   services: [docker]}
+    - {env: IMPL=nasm,      services: [docker]}
     - {env: IMPL=nim,       services: [docker]}
     - {env: IMPL=objpascal, services: [docker]}
     - {env: IMPL=objc NO_DOCKER=1, os: osx, osx_image: xcode7}
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ Mal is implemented in 71 languages:
 * mal itself
 * Matlab (GNU Octave and MATLAB)
 * [miniMAL](https://github.com/kanaka/miniMAL)
+* NASM
 * Nim
 * Object Pascal
 * Objective C
@@ -615,6 +616,19 @@ IMPL_STEPA_CMD ../mal/stepX_YYY.mal
 ```
 cd make
 make -f stepX_YYY.mk
+```
+
+### NASM
+
+*The NASM implementation was created by [Ben Dudson](https://github.com/bendudson)*
+
+The NASM implementation of mal is written for x86-64 Linux, and has been tested
+with Linux 3.16.0-4-amd64 and NASM version 2.11.05.
+
+```
+cd nasm
+make
+./stepX_YYY
 ```
 
 ### Nim 0.17.0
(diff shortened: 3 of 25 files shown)


dubek:
Wow! Thank you for this very hard implementation!

nasm/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM ubuntu
```
dubek: Not sure if we should set the exact ubuntu version as we do elsewhere.

kanaka:
Cool! I'm on vacation at the moment so I probably won't get a chance to
review until early next week.

kanaka:
@bendudson I built and pushed a kanaka/mal-test-nasm container (with an updated FROM line to ubuntu:xenial). Also, I fixed the Swift build error in travis (due to Travis upgrading XCode). If you update your nasm/Dockerfile to have ubuntu:xenial, activate nasm in the travis file and push, I suspect everything should pass now including nasm.

A couple of miscellaneous things:
- Bump the language count in the README.
- Can you add stats and stats-lisp targets to your Makefile? This allows a LOC comparison between different implementations. You should be able to look at Makefiles for other implementations for reference. The only differences between implementations are the specific list of low-level files (such as exceptions.asm) and how to match comments.

I still probably won't have a chance to review and merge until next week, but the brief glance looks good. If we need to add new features or fix bugs across all implementations, will you be available to do those changes in the nasm implementation? @wasamasa, @dubek and I can handle most of the others but this one is a bit more complex than average and I personally only have surface level assembly experience.

bendudson:
Happy new year! Yes I should be around, and am happy to keep tinkering and occasionally fixing this implementation when needed. 

Sorry, it looks like the ubuntu:xenial docker image doesn't have `ld` installed (a bit odd). I've added "binutils" to the Dockerfile install and it seems to work now. I haven't added nasm to the Travis build yet though.

kanaka:
@bendudson I pushed an updated version of the docker image "kanaka/mal-test-nasm" that has binutils (still xenial).

Also, in addition to the stats rules I notice there is no "clean" rule. Something like this should work:

```patch
--- a/nasm/Makefile
+++ b/nasm/Makefile
@@ -5,6 +5,9 @@ COMPONENTS=core.asm reader.asm printer.asm types.asm system.asm
 
 all: $(STEPS)
 
+clean:
+       rm -f $(STEPS)
+
 %.o: %.asm $(COMPONENTS)
        nasm -felf64 $<

```

kanaka:
It passes in Travis :-)
I'm doing a bit more in depth review now and should have some feedback shortly.

kanaka:
@bendudson In general, it looks good. My assembly is rusty and never that great to begin with so the feedback that follows is mostly structural/code organization in nature:

- A primary goal of mal is as a learning tool both for those doing an implementation and for those who want to learn from existing implementations. So as a general rule, doing a `diff -urp stepX*.asm stepY*.asm` between two adjacent steps should be useful for somebody wanting to see only what changes were necessary to implement just that step. More detail in the FAQ: https://github.com/kanaka/mal/blob/master/docs/FAQ.md#why-are-the-malmake-a-lisp-steps-structured-the-way-they-are Here are a few things I found in this category:
  - There are some general indent/spacing between some of the steps that could be normalized. This is common while developing later steps to change spacing/comments/structure of course. But going back and making things consistent in earlier steps is desirable.
  - step0 and step1 have a bunch of stuff that moves into types.asm in later steps. I would suggest just including types.asm if you're able. Another general principle is that the step files should mostly just contain the Lisp-ish stuff (admitedly it's vague and not crisply defined) and data-type/memory layout issues should mostly be in types or other shared files.
  - found some occasional string differences that are probably unintentional. For example not_found_string has period in step8 and not in step9
  - read_eval should probably be defined starting in step6 and used for load-file for step6-A
  - It looks like exceptions.asm is included in every step except for step7 where the exception code is defined in step7. I suspect this was code that was in the step files and moved to exceptions.asm but step7 was never updated.
- I notice that step2 includes env.asm. The typical pattern is that env.asm is not included until step3_env. Is there some common code in env.asm that could move to types.asm (or another common file) so that step2 doesn't need to include env.asm? Like-wise for core.asm which is included in step2 but is generally not included until step4.

Other than the cleanup above, I think it's almost ready to merge.

bendudson:
@kanaka Thanks for the review, it's a lot of code to get through! I think I have made all the changes, and have tried to minimise the changes between steps. 

* `env.asm` has been removed from step2. I replaced the Env object with a Map, which is defined in `types.asm`

* `core.asm` has been removed from steps 2 and 3. This means I had to copy the arithmetic functions into the step 2 and 3 files, but they go into `core.asm` in step 4.

kanaka:
@bendudson That's basically the same solution most other implementations use (separation of Map and Env and implementing arith in step2/3). Merged.

I think the only remaining task is a tweet announcing the implementation :-). If you don't have a twitter account (or otherwise don't want to) I can do so.

Also, if you use IRC (or want to start) there is a #mal channel on freenode where some of the core implementors and other people interested in mal hang out. It's very low traffic on average.

As a note of trivia, even though this is a single implementation, it's approximate 10x the median implementation size. Also, it's 7th on the highly unscientific micro-benchmark (after rpython, java, kotlin, scala, nim and factor).

bendudson:
Thanks! I don't have a Twitter account, sorry. 

The JVM is impressively fast, but I'm glad I could beat some compilers. I definitely appreciate more having a compiler now. 

Good luck with the next implementation!

PR #240, opened 2016-10-10
Ada2

asarhaddon:
I hope that review will be easyer with a pull request.
Please warn me of any change by email, as I do not intend to connect to github on a regular basis.
Thanks.

PR #311, opened 2017-12-29
New tests for steps 7 and 9

bendudson:
Split from PR #310, to avoid accidentally breaking other implementations.
(after reading comment in PR #240) 

These catch three errors in the NASM implementation which only appeared in self-hosting.


kanaka:
Looks good. Only failure in existing implementations was in Swift but for a unrelated reason. Merged.

PR #312, opened 2017-12-30
Update guide with note on reader macro with-meta

jonaslu:
Guide was missing the deferrable step of having
a reader macro for ^ (which is included in the
tests for step A) which caused some confusion
as to what it was.

It's included in the tests/step1_read_print.mal
but not mentioned in the docs.


kanaka:
Thanks for catching that!

PR #317, opened 2018-06-22
Rename short abbreviations

wasamasa:
As discussed previously, `gst` and `pil` aren't exactly clear and shouldn't have been used in the first place.


kanaka:
I tagged the new docker images and pushed and restarted the test. gnu-smalltalk and picolisp passed. Merged. Apparently external deps for java and ts have changed though so they are newly broken. Sigh.

PR #319, opened 2018-06-24
PHP: refactor web running mechanism.

chr15m:
Hi @kanaka,

This change is to make it easier to re-use `php/mal` in other projects. It pulls the web runner code out into `webrunner.php` and includes it at build time instead of including that code directly in `stepA_mal.php`.

I made this change to help with [frock](https://github.com/chr15m/frock), which is an experimental tool to build single-artifact PHP scripts from `mal` code. It relies heavily on and is itself written in `mal`. I also cleaned up the documentation a little.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,6 +90,7 @@ objpascal/regexpr/Source/RegExpr.ppu
 perl/mal.pl
 perl6/.precomp/
 php/mal.php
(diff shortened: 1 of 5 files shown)


kanaka:
Looks good and php test passes. Merged.

PR #320, opened 2018-06-24
Updated PHP native interop interface.

chr15m:
Hi @kanaka,

This is a change to the way the PHP interop presents to the end user and to clean up the interface. Interop is now exposed via the `php/` prefix for reaching into the PHP "namespace" and is the same for referencing functions and variables.

Previously to make native function call you'd do: `(! str_replace "l" "x" "hello")` whereas this patch changes it to work like this: `(php/str_replace "l" "x" "hello")`.

Likewise for referencing PHP global vars you previously used `($ "_SERVER")` whereas with this patch you instead use `php/_SERVER`.

These changes make the interop more Clojure-like which has e.g. the `js/...` namespace for referencing JavaScript objects. An added benefit is that editors now know you mean `php/str_replace` as a function call and will highlight it appropriately.

I made this change to help with [frock](https://github.com/chr15m/frock), which is an experimental tool to build single-artifact PHP scripts from `mal` code. I'm interested in your feedback on that tool.

### DIFF
--- a/php/README.md
+++ b/php/README.md
@@ -32,11 +32,11 @@ Eval PHP code:
 
 Native function call:
 
-	(! date "Y-m-d" 0)
-	1970-01-01
+	(php/date "Y-m-d" 0)
+	"1970-01-01"
 
 Accessing PHP "superglobal" variables:
 
-	(get ($ "_SERVER") "PHP_SELF")
-	./mal
+	(get php/_SERVER "PHP_SELF")
+	"./mal"
 
--- a/php/interop.php
+++ b/php/interop.php
@@ -45,4 +45,20 @@ function _to_mal($obj) {
     }
 }
 
+function _to_native($name, $env) {
+  if (is_callable($name)) {
+    return _function(function() use ($name) {
+      $args = array_map("_to_php", func_get_args());
+      $res = call_user_func_array($name, $args);
+      return _to_mal($res);
+    });
+  } else if (in_array($name, ["_SERVER", "_GET", "_POST", "_FILES", "_REQUEST", "_SESSION", "_ENV", "_COOKIE"])) {
+      $val = $GLOBALS[$name];
+  } else if (defined($name)) {
+      $val = constant($name);
+  } else {
+      $val = ${$name};
+  }
+  return _to_mal($val);
+}
 ?>
--- a/php/reader.php
+++ b/php/reader.php
@@ -27,7 +27,7 @@ function _real_token($s) {
 }
 
 function tokenize($str) {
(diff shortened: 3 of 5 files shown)


kanaka:
Now that I've merge #319 this no longer merges cleanly. The changes look fine though and I'll be happy to merge it once the conflict is resolved.

In terms of frock: I approve :-) I would also support adding a section to the main mal README with mal projects. In particular, another project that should go in that section is https://github.com/dubek/malc Speaking of which, a while back @dubek and I discussed making something like a lib/ directory to contain some standard library type things that a person might want to use with mal. Some of the files that are currently in examples/ (like protocols.mal, memoize.mal and pprint.mal should go in lib/). Since it seems like you are motivated to create some of the Clojure-like macros/functions for frock, I would be happy to see them in the main mal tree under lib/*.mal assuming they are generic and not PHP specific (things like destructuring, let, defn, etc). Even the core.mal file at the top-level should probably be moved into lib/ (although there are some things that assume it's at the top-level that would need to be updated such as some of the test files).

chr15m:
@kanaka thanks! I've clobbered the old PR with a new commit which fixes the merge issues and a couple of minor nits.

On the projects README section, where would you prefer this goes? Somewhere down near [Dockerized testing](https://github.com/kanaka/mal/#dockerized-testing)?

The `lib/` folder re-organisation sounds like an excellent idea. I hadn't seen the examples folder :man_facepalming: but I will certainly check that out and see what I can use in frock.

My strategy for frock was going to be to build minimally functional macros for common Clojure things like `let` and then if I ever get them to the point where they are not just a hack and they support e.g. destructuring etc. then I'd make PRs to bring them into `mal/lib`. You can see [here](https://github.com/chr15m/frock/blob/master/src/alias-hacks.clj#L6-L11) that what I have working so far is basically just wrappers around the mal builtins and I don't think those are useful enough to bring in until they have more of the functionality of the Clojure equivalents.

I'd be happy to do that macro work inside `mal/lib` instead though but I can't promise I'll have the competence or time to implement e.g. destructuring any time soon. I took a look at the problem of destructuring and some other implementations (like hy) and realised I am out of my depth. :) What is required is some LISP macro ninja which I am not.

kanaka:
A new section near Dockerized testing would be fine.

It's probably not a bad idea to test thing in frock and then when you are a bit more confident move them to mal/lib/. However, don't worry about completeness before putting them there. You can build them incrementally. For example, destructuring doesn't need support all forms of destructuring. I.e. I suspect list/sequence destructuring is a good place to start and it's not necessarily that complicated.

The clojurescript, destructuring is done via a function. Here is the destructure function: https://github.com/clojure/clojurescript/blob/d91207cb7386365a07b563b09b6444846657a364/src/main/clojure/cljs/core.cljc#L626 The vector destructure code (pvec) covers about 3 dozen lines, but it's spread out due to long lines. The function basically returns a rewritten let block that is then unquoted into the let* form in macros that use it like let, defn, etc.

PR #323, opened 2018-07-06
Go: Fix panic on call to function with wrong number of arguments

jig:
Core functions do not check argument count when called. If there are less than needed the program panics. If there are more they are ignored.

Now return an error in case the argument count is not expected instead of panicking.

Fixed on all steps from step2 to stepA.


kanaka:
@jig sorry for the slow followup. Your updates look good and I have merged them.

PR #324, opened 2018-07-09
Test that (not nil) returns true

seven1m:
In step 4, by mistake I implemented the `not` function in my host language rather than in mal itself as the guide instructs. And I did it poorly. `(not nil)` was returning `false`, and I managed to make it all the way past stepA, not finding the error until the self-hosting tests.

This little test in step 4 should help others I hope!

### DIFF
--- a/tests/step4_if_fn_do.mal
+++ b/tests/step4_if_fn_do.mal
@@ -243,6 +243,8 @@ a
 ;; Testing language defined not function
 (not false)
 ;=>true
+(not nil)
+;=>true
 (not true)
 ;=>false
 (not "a")


dubek:
Thanks for the addition. Most implementations implement this correctly because they use the mal-fn implementation according to the guide:

```
# core.mal: defined using the language itself
REP("(def! not (fn* (a) (if a false true)))")
```

This surfaced other build errors, though:

@wasamasa can you look at the scheme-cyclone failures in stepA? https://travis-ci.org/kanaka/mal/jobs/401600084

@kanaka The java implementation fails on something related to maven, here's the relevant log part:

```
Downloading: https://repo1.maven.org/maven2/org/apache/maven/plugins/maven-resources-plugin/2.3/maven-resources-plugin-2.3.pom
[WARNING] Unable to get resource 'org.apache.maven.plugins:maven-resources-plugin:pom:2.3' from repository central (https://repo1.maven.org/maven2): Error transferring file: Received fatal alert: protocol_version
Downloading: https://repo1.maven.org/maven2/org/apache/maven/plugins/maven-resources-plugin/2.3/maven-resources-plugin-2.3.pom
[WARNING] Unable to get resource 'org.apache.maven.plugins:maven-resources-plugin:pom:2.3' from repository central (https://repo1.maven.org/maven2): Error transferring file: Received fatal alert: protocol_version
```

Scala has similar (?) issues:

```
:::: ERRORS
	Server access Error: Connection reset url=https://jcenter.bintray.com/org/scala-sbt/sbt/0.13.9/sbt-0.13.9.pom
	Server access Error: Connection reset url=https://jcenter.bintray.com/org/scala-sbt/sbt/0.13.9/sbt-0.13.9.jar
	Server access Error: Connection reset url=https://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt/0.13.9/ivys/ivy.xml
	Server access Error: Received fatal alert: protocol_version url=https://repo1.maven.org/maven2/org/scala-sbt/sbt/0.13.9/sbt-0.13.9.pom
	Server access Error: Received fatal alert: protocol_version url=https://repo1.maven.org/maven2/org/scala-sbt/sbt/0.13.9/sbt-0.13.9.jar
```

any idea? (full log - https://travis-ci.org/kanaka/mal/jobs/401600046 )

@vvakame Can you please take a look why Typescript fails entirely? https://travis-ci.org/kanaka/mal/jobs/401600090

wasamasa:
@dubek I rebuilt the Scheme parts with the latest Cyclone Scheme (0.8.1) and the test works fine, so I suspect rebuilding the Docker image should fix it. That's not something I can do though, @kanaka would need to push a new one.

vvakame:
I wonder if I do not do anything?

kanaka:
I think I may have fixed java, scala and TypeScript. I think it's either the deps couldn't be retrieved anymore (Java, Scala) or something got updated and trigger a new compilation error (TypeScript).

I also change the time-ms function check that the value is non-zero because Java also had a non-fatal test failure in time-ms because it was returning a negative value. But signed values are fine for the time-ms return value as long as the start non-zero and they increase over time.

We'll see if all three pass now: https://travis-ci.org/kanaka/mal/builds/403273094

kanaka:
I also rebuilt the scheme container and restarted all the scheme tasks and everything is now passing. I've merged this pull.

@wasamasa BTW, can you remind me why foment is not activated in Travis? Was it that it is temperamental or maybe something about running in Travis? I tried running a few times on my local system and it passed each time.

wasamasa:
@kanaka Foment did hang in Travis, IIRC it was at the performance measuring step. In case that was it, I've got an elegant workaround: Why not measure performance with the testing rig? This would alleviate the need for implementing `time-ms` (which sometimes requires workarounds, such as shelling out to `date`...) and it would ensure time is measured correctly (some languages give you the CPU time instead of wall time...).

wasamasa:
Hm, I thought you pretty much have to print banners for that approach to work. You could even use that approach for the third test by modifying it a bit, print after 10, 100, 1000, 10000, 100000 iter/s and kill off the process if it exceeds a given time limit. I'm thinking of a Python library here which does automatically figure out the ideal amount of iterations for benchmarking.

As for Foment, feel free to reenable. It would be nice if the bug's gone.

kanaka:
@wasamasa looks like foment still hangs during the performance testing. I've deactivated it again.

PR #330, opened 2018-10-27
Update Crystal implementation

omarroth:
Ran `crystal tool format` with latest version of Crystal (0.26.1), as well as a couple changes to fix type inference. Mostly cosmetic changes. I removed the custom `readline` binding since it's identical to the version in the standard library.


omarroth:
Ah, looks like Travis is using [this](https://hub.docker.com/r/kanaka/mal-test-crystal/tags/) container, which hasn't been updated for about 2 years.

crystal/core.cr
```diff
@@ -1,456 +1,454 @@
...
 require "./types"
 require "./error"
 require "./printer"
 require "./reader"
-require "./readline"
 
 module Mal
-
-macro calc_op(op)
+  macro calc_op(op)
```
rhysd: Is this formatting correct? Indentation looks weird. I think other `macro ... end` blocks are also the case.

kanaka:
I've updated the version of the crystal docker container to 0.26.1. Once you've resolved the issue suggested by @rhysd then I'll merge in these changes.

kanaka:
I restarted the crystal Travis task and it passed with the update image.

crystal/core.cr
omarroth: Looks like the formatter doesn't touch macros, fixed!

PR #333, opened 2018-12-18
ojbc: fix conj, meta and with-meta

LispLY:
There are 15 failing tests with the objc implementation in stepA. ( Tested in Mac OS. )
I've tried to modify it to pass the tests.


kanaka:
I was on vacation so I'm just getting to this now. Merged. Thanks!

PR #335, opened 2019-01-21
Fixing function name in guide

Windfarer:
According to [Cheatsheet](http://kanaka.github.io/mal/process/cheatsheet.html) step 1 and some of implementations like [Python](https://github.com/kanaka/mal/blob/master/python/reader.py#L21) or [Javascript](https://github.com/kanaka/mal/blob/master/js/reader.js#L17), this function name should be `tokenize` in guide.


kanaka:
Yes, you are correct. Thanks for the correction.

PR #337, opened 2019-01-27
upgrade clojure and tools.reader

AndreaCrotti:


kanaka:
Thanks. I also just realized that the tools.reader dep is actually no longer required but I'll remove that in a subsequent commit.

PR #338, opened 2019-01-30
In step1, test non-numbers starting with a dash.

asarhaddon:


kanaka:
Looks like the only two failures are F# and Swift 3.

PR #339, opened 2019-02-02
Check that the eval built-in uses Repl instead of a local environment.

asarhaddon:
The description in process/guide.md states:
> The closure calls the your `EVAL` function using the `ast` as the first argument and the REPL environment (closed over from outside) as the second argument.
Reading this, I would expect
```
   (def! a 1)
   let* (a 1) (eval 'a))

```
to answer 1.
The only failing implementation is nim, and it seems to use a local environment instead of repl_env when evaluating the (locally evaluated) first argument.
The suggested test is after
`;>>> optional=True
and should be visible as an optional test.
I have no opinion about this, but process/guide.md should be updated if repl_env is not required.

### DIFF
--- a/nim/step5_tco.nim
+++ b/nim/step5_tco.nim
@@ -2,7 +2,7 @@ import rdstdin, tables, sequtils, types, reader, printer, env, core
 
 proc read(str: string): MalType = str.read_str
 
-proc eval(ast: MalType, env: var Env): MalType
+proc eval(ast: MalType, env: Env): MalType
 
 proc eval_ast(ast: MalType, env: var Env): MalType =
   case ast.kind
@@ -19,8 +19,9 @@ proc eval_ast(ast: MalType, env: var Env): MalType =
   else:
     result = ast
 
-proc eval(ast: MalType, env: var Env): MalType =
+proc eval(ast: MalType, env: Env): MalType =
   var ast = ast
+  var env = env
 
   template defaultApply =
     let el = ast.eval_ast(env)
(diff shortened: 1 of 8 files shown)


asarhaddon:
Of course, the code should be
 ```
  (def! a 2)
   let* (a 1) (eval 'a))
```

**Comment by asarhaddon** [comment]
Well, sorry. The code should be as in the pull request:
```
(def! a 1)
  let* (a 2) (eval 'a))
```
and the output should be 1.

**Comment by asarhaddon** [comment]
The new test shows two problems. Both seem actual bugs.

There factor implementation:
```
    Expected : "\\(let\\*\\ \\(a\\ 2\\)\\ \\(eval\\ \\'a\\)\\)\r\n1"
    Got      : "(let* (a 2) (eval 'a))\r\nError: No suitable method"

```
The nim implementation does attempt to use repl_env, but breaks let* between steps 3 and 7. Here is a test showing the regression.
```
TEST: '(def! a 1)' -> ['',1] -> SUCCESS
TEST: "(let* (a 2) (eval 'a))" -> ['',1] -> FAIL (line 189):
    Expected : "\\(let\\*\\ \\(a\\ 2\\)\\ \\(eval\\ \\'a\\)\\)\r\n1"
    Got      : "(let* (a 2) (eval 'a))\r\n2"
TEST: 'a' -> ['',1] -> FAIL (line 191):
    Expected : 'a\r\n1'
    Got      : 'a\r\n2'

```

asarhaddon:
Last commit fixes the nim implementation. The problem was a regression in step5, not detected because REGRESS=1 is not the default.
I suggest to make REGRESS=1 the default, or at least recommend it in guide.md.
Just in case: I had to modify nim/reader.nim to compile locally. If you encounter similar problems, I may submit the changes.

PR #345, opened 2019-03-17
Ada.2 distinct Ada implementation

asarhaddon:
Hello.
As discussed at #240, here is a pull request with a cleaner history, as well as a new large commit.
Most recent changes should be straightforward, but you may want to check the two first items in the header. Both would be easy to revert.


asarhaddon:
The .travis.yml script fails because of the bug described at the end of the README file.
I have added a commit removing the -gnatp option.

kanaka:
@asarhaddon looks like there is a different failure now with perf3 (range check error in envs.adb).

kanaka:
@asarhaddon nice work. I've merged it and update the README to reflect that this is a second implementation in Ada. I noticed that in addition to being a fair bit more concise, this implementation does a fair bit better on the microbenchmarks too.

PR #348, opened 2019-03-31
ada.2: fix memory leaks with garbage collection. Various simplifications.

asarhaddon:
Cyclic references were never deallocated by reference conuting.
Symbols cannot create cyclic structures and are less frequent (one
allocation per symbol), keep reference counting for them.

This slightly improves performances even though many previous
optimizations are removed (environment stack, reuse of memory).

Step caching hash of symbols. This does not seem to improve
performances. Hashing them instead of ordering them does.

Define Repl in the step file instead of globally. Move the eval
built-in function from core into the step file.

When possible, pass Ada records instead of explicit pointers.

In the reader, construct more objects directly as described in the MAL
process, reserve the buffer for sequences and maps

In eval, iterate on vectors without delegation. The increased
complexity was not improving performances.  Keep demonstrating Ada
type-safe genericity for maps, where iterating outside Types.Maps
would be less easy and/or efficient.

In quasiquote_list, concatenate in one buffer instead of allocating a
list for each element. The buffer may be reallocated behind the
curtain, but not once per element anymore.

In environments, illustrate tail call optimization when recursion is
more readable than a loop.

### DIFF
--- a/ada.2/Makefile
+++ b/ada.2/Makefile
@@ -37,6 +37,7 @@ TYPES := \
   envs.ads                 envs.adb                 \
   err.ads                  err.adb                  \
   eval_cb.ads                                       \
+  garbage_collected.ads    garbage_collected.adb    \
   printer.ads              printer.adb              \
   reader.ads               reader.adb               \
   readline.ads                                      \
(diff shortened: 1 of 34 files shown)


kanaka:
Merged.

PR #351, opened 2019-04-08
guide.md: Don't use vectors in step 4 when they're still deferrable.

bjh21:
Thank you for mal.  I've been enjoying working my way through the process. I've reached step 4, but not yet implemented vectors. The manual tests of `fn*` in step 4, though, use vectors to specify the function parameters. This may be idiomatic mal, but it doesn't work if you've deferred implementing vectors. Thus, I think it would be better to use lists here, like the automated tests do.

### DIFF
--- a/process/guide.md
+++ b/process/guide.md
@@ -765,10 +765,10 @@ the apply section of `EVAL`.
 
 Try out the basic functionality you have implemented:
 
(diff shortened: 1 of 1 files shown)


kanaka:
@bjh21 yes, I agree. That's a good improvement. Thank you.

PR #352, opened 2019-04-09
Mark step 4 tests that use strings as deferrable.

bjh21:
Hello again. This is another discovery stemming from my extreme laziness. I haven't implemented strings yet, but a bunch of non-deferrable tests at step 4 use strings. I propose fixing this by two means: where the test is of string functionality, I've marked it as deferrable. Where the strings are just constants to be returned or printed by the code, I've replaced them with integers.

### DIFF
--- a/tests/step4_if_fn_do.mal
+++ b/tests/step4_if_fn_do.mal
@@ -18,10 +18,10 @@
 ;=>0
 (count nil)
 ;=>0
-(if (> (count (list 1 2 3)) 3) "yes" "no")
-;=>"no"
-(if (>= (count (list 1 2 3)) 3) "yes" "no")
-;=>"yes"
+(if (> (count (list 1 2 3)) 3) 89 78)
+;=>78
+(if (>= (count (list 1 2 3)) 3) 89 78)
+;=>89
 
 
 ;; Testing if form
@@ -38,8 +38,6 @@
 (if nil 7 8)
 ;=>8
 (if 0 7 8)
-;=>7
-(if "" 7 8)
 ;=>7
 (if (list) 7 8)
 ;=>7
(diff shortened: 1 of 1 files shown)


kanaka:
Yes, that's a good improvement. Keep them coming!

Out of curiosity what language are you doing you implementation in right now?

bjh21:
My implementation is in BBC BASIC V, the BASIC shipped with the first ARM-based computers in the mid-1980s.  It's the language I did most of my serious teenage programming in.  Unlike the dialect used by the existing BASIC implementation, it's got recursion, named procedures, local variables, sensible control structures, and even tolerable exception handling.

kanaka:
@bjh21 huh. I wasn't familiar with BBC Basic. That's really interesting and ahead of its time. If you are able to complete your implementation we'll have to decide on a naming scheme to disambiguate it from the "basic" dir that holds the c64/qbasic implementation. Have you made any progress in the past 6 days?

bjh21:
I'm currently using a directory called "bbcbasic", but perhaps with the precedent of "gnu-smalltalk" it should be "bbc-basic".  I think keeping the existing BASIC port called "basic" would be sensible: it's probably portable to almost any BASIC dialect, including the BBC ones.

I've not made much progress recently because I found that step 5 needs a garbage collector, and that needed some thinking. I have a plan now, though, so I'll see if it works...

kanaka:
@bjh21 "bbc-basic" sounds fine to me. And yes, if your target language doesn't have automatic GC, then GC does tend to be the most complicated aspect of a mal implementation. I have a compact memory layout and reference counting pattern (with some simple GC needed during macro expansion) that I used for the basic impl that I adopted for the wasm implementation too. I also have a C implementation that uses the same model that I haven't pushed yet but will at some point. And I should probably do a better job of documenting how it works so that other low-level languages could use it if they want. Out of curiosity, how much memory does BBC Basic leave available to user programs? Hopefully more than the 38911 bytes available to C64 basic. :-)

bjh21:
I was thinking that Hints.md could do with an explanation of how to write a minimal garbage-collector. Unfortunately, I may not be the person to write it, since I found reference-counting too confusing and have implemented a naïve mark-and-sweep collector instead.

Because BBC BASIC V was written for a 32-bit CPU (unlike BASIC I through IV, which were for the 6502), it can happily use several megabytes of memory. My test environment (using Brandy, an impementation of BBC BASIC V for Linux) happens to give me half a megabyte of user memory by default, which is conveniently not quite enough that I can get away without garbage collection.

PR #354, opened 2019-05-02
Trivial simplifications in core.mal.

asarhaddon:
Replace many (= (count xs) 0) with (empty? x).

Stop definining 'not', 'or' and 'cond' in core.mal, they are defined
in the process steps.

In 'some', avoid computing the same result twice.

Remove the final nil evaluation.

### DIFF
--- a/core.mal
+++ b/core.mal
@@ -6,30 +6,28 @@
 
 (def! reduce
   (fn* (f init xs)
-    (if (> (count xs) 0)
-      (reduce f (f init (first xs)) (rest xs))
-      init)))
+    (if (empty? xs)
+      init
+      (reduce f (f init (first xs)) (rest xs)))))
 
 (def! identity (fn* (x) x))
 
 (def! every?
   (fn* (pred xs)
-    (if (> (count xs) 0)
+    (if (empty? xs)
+      true
       (if (pred (first xs))
         (every? pred (rest xs))
-        false)
-      true)))
-
-(def! not (fn* (x) (if x false true)))
+        false))))
 
 (def! some
   (fn* (pred xs)
-    (if (> (count xs) 0)
+    (if (empty? xs)
+      nil
       (let* (res (pred (first xs)))
-        (if (pred (first xs))
+        (if res
           res
-          (some pred (rest xs))))
-      nil)))
+          (some pred (rest xs)))))))
 
 (defmacro! and
   (fn* (& xs)
@@ -40,25 +38,6 @@
         (let* (condvar (gensym))
           `(let* (~condvar ~(first xs))
             (if ~condvar (and ~@(rest xs)) ~condvar)))))))
-
-(defmacro! or
-  (fn* (& xs)
-    (if (empty? xs)
-      nil
-      (if (= 1 (count xs))
-        (first xs)
-        (let* (condvar (gensym))
-          `(let* (~condvar ~(first xs))
-             (if ~condvar ~condvar (or ~@(rest xs)))))))))
-
-(defmacro! cond
-  (fn* (& clauses)
-    (if (> (count clauses) 0)
-      (list 'if (first clauses)
-            (if (> (count clauses) 1)
-                (nth clauses 1)
-                (throw "cond requires an even number of forms"))
-            (cons 'cond (rest (rest clauses)))))))
 
 (defmacro! ->
   (fn* (x & xs)
@@ -84,4 +63,6 @@
             (list form x))
           `(->> (->> ~x ~form) ~@more))))))
 
+; This `nil` is intentional so that the result of doing `load-file` is
+; `nil` instead of whatever happens to be at the end of `core.mal`.
 nil


core.mal
```diff
@@ -83,5 +62,3 @@
             `(~(first form) ~@(rest form) ~x)
             (list form x))
           `(->> (->> ~x ~form) ~@more))))))
-
-nil
```
kanaka: This `nil` is intentional so that the result of doing `load-file` is `nil` instead of whatever happens to be at the end of `core.mal` (in this case `->>`).

asarhaddon:
Hello.
It may make more sense to move the native definitions the other way around: or, not, cond, gensym fit in ../core.mal better than in the step files.
It make sense that ARGV, *host-language*, and load-file are part of the process, as well as 'not' as an early example, but other native definitions increase the redundancy between steps and implementations for little benefit. For example, the gensym explanation would be more interesting as comments in core.mal than in the process guide (at least as long as it is implemented in MAL).
If you approve the idea, I am interested in replacing this pull request and applying the change to each implementation.

asarhaddon:
While at it, I see that the C++ implementation contains `
(def! list (fn* (& items) items))
(def! >= (fn* (a b) (<= b a)))
(def! < (fn* (a b) (not (<= b a))))
(def! > (fn* (a b) (not (<= a b))))
`
Would you agree to replace all possible builtins with native definitions?
- Performances would be slightly affected, but this does not matter much.
- New implementors would spend less time in core.qnx, which is not the most interesting part.
- Maintenance and comparison between languages would be easier.
- core.mal should be split in several coreSTEP.mal distinct sources adding the definitions (instead of the process asking to implement them in core.qnx). This is a bit of work, but trivial and shared by all implementations.
- Much less tests would be necessary, the shared code can be trusted.

In case you are not convinced yet, please read the C++ definition of `<` above again :-)

asarhaddon:
Loading `../core.mal` instead of defining the `cond`, `gensym` and `or` macros seems straightforward in most implementations. Only forth, guile and livescript seem to require non-trivial changes. It is probable that testing will reveal problems with the hardcoded path though.
I may update this pull request and trigger a travis run if you are interested by the idea.
By the way, I may have found a way to make the counter private in the `gensym` implementation.
`(def! gensym (let* [counter (atom 0)] (fn* [] (symbol (str "G__" (swap! counter inc)))))`.

kanaka:
The modifications to core.mal look good and I've gone ahead and merged them.

I fully agree with your gensym proposed and would be happy to take a pull request with that change to the implementations.

I don't think builtins like `list`, `>=`, etc should be implemented in mal. These are very much part of the pedagogic aspect of the make-a-lisp process. In fact, I would be happy to merge a pull request that made C++ consistent with the other implementations.

As far as the other change you proposed regarding `core.mal`, `or` and `cond`, the reasons you suggest for the change make sense, but there are some reasons why the core/builtin functionality is split the way it is. The current split represents a balance between minimal code for the implementor, maximal functionality of the implementation, and reasonable performance for really slow target languages. As an example, a more minimal process would get rid of vectors, and implement quoting, macros in the mal language itself (instead of having steps for that). An even more minimal process would implement `let*` and `do` using `fn*` and implement maps using lists (as some older Lisps do). Those changes would certainly mean the mal implementor would be able to write less code. But there are several target languages that are already quite slow that would basically be unusably slow. But more importantly, I think it would make the learning curve less gentle for implementors who are interested in learning Lisp by creating a Lisp implementation (which is one of the primary goals of mal).

The basic rule of thumb for the current split is that once you've completed stepA and all deferred tasks, you can self-host a reasonably nice implementation of mal itself without needing to load `core.mal` first. I think you rightly identified `or` and `cond` as odd things to have in the main step files. But they are there because the mal implementation (i.e. the implementation in `mal/`) would be MUCH less friendly without those two constructs. In particular, the functionality of `or` and `cond` can be replicated with a series of nested `if` statements, but the code becomes deeply nested and the understandability of the code drops significantly. core.mal's purpose is to serve as common code for testing and also as a library of useful functions and macros that an implementator can use once their implementation is complete. I don't consider the code in core.mal as an essential part of the foundational make-a-lisp process.

All that being said, I think a page that describes how many of the mal features could be implementing using mal itself would be a really useful resource. One interesting way to present it would be as a set of problems to solve. "Now that you've implemented mal, here are some followup challenges." Examples, "Implement >, <=, and >= using = and <", "Implement list using fn*", "Implement let* as a macro that uses fn*", etc, etc. I think a wiki page or pages that captured this sort of information would be really valuable and I would happily consider adding that to the project if you're interested in creating it.

dubek:
I think that `def!` inside `do` will "leak" to the environment surrounding the `do` (unlike `let*`):

```
$ ./run
Mal [ruby]
user> (do (def! key 123) key)
123
user> key
123
```

as opposed to:

```
user> (let* (key2 456) key2)
456
user> key2
Error: 'key2' not found
        /mnt/c/personal/mal/ruby/env.rb:34:in `get'
        ./stepA_mal.rb:49:in `eval_ast'
        ./stepA_mal.rb:69:in `EVAL'
        ./stepA_mal.rb:160:in `block in <main>'
        ./stepA_mal.rb:185:in `<main>'
```

asarhaddon:
That is why I only suggest such changes inside a `fn*` (frequent) or another `let*` (rare). Both commands already create a subenvironment, and most of the time it can be reused.

kanaka:
@asarhaddon I'm not sure what happened to Option B, but I'm fine with any of option A, C or D. Perhaps a slight preference for D but that is also the one that would need the most work across implementations and need additional tests so I'll leave that to you. I do want to avoid using `def!` within lexical environments (e.g. `fn*` and `let*`). In other Lisps, `def*` forms are really for namespace mutation. Mal doesn't have named namespaces, but it's fairly simple in most implementations to add support for them by extending symbols and environments. And in that case `def!` should modify the namespace and not the current lexical environment. So I don't want to add implicit code or tests that implicitly assume that `def!` modifies the current lexical environment.

PR #355, opened 2019-05-02
Clarify and optimize ada.2.

asarhaddon:
Makefile:
Drop OPT and -gnat2020, not used anymore.
Simplify file list now that each unit has a body.

README:
Remove obsolete items.

Global:
Restrict most pointers to a provable non-null value.

Types: merge intermediate Types.Mal into the Types package.  (the
intermediate package was created in order to prevent a circular
dependency, but is not needed anymore).
Most of the noise in the diff is caused by this change.
This allows to remove most Elaboration pragmas.
Declare most types abstract in the visible part,
enforcing the use of the constructor outside the declaring package.

Envs:
Replace the Get recursion with a more efficient loop.
Use MAL objects as key, string pointers do not change speed.
This delegates some checks from the step files.
Split the constructor and Set_Binds, so that an existing environment
can be reused during TCO.

Err:
Attempt to group the calls.
Avoid computing the message when the assertion holds.

Fns:
Declare and use the eval callback only here.
Separate function and macro interfaces.
Keep a reference to the provided parameter list instead of copying them.

Garbage_Collected:
Make explicit that Keep is not inherited.

Printer:
Remove obsolete inline indications and redundant Print_Function helper.

Maps:
Provide a cleaner interface copied from standard library.

Sequences: stop encapsulating the implementation because of the
performance hit.

Steps:
Move map and vector evaluations into separate functions for readability.
Replace return blocks with normal blocks (MAL values are not finalized
anymore).
Rename standard arrays instead of sequence_ptr when possible.
Remove some duplication and indentation from the apply phase.
Move the frequent special forms in front of the test cascade.
When an environment has been created in the same Eval, reuse it.

Strings:
Use the same garbage-collected storage model for all strings.
This seems faster than the default (mutable) string types.
Hide most of the implementation to avoid leaks.

Symbols: stop ensuring unique allocation of symbols. The reduced
garbage collection and comparison time was compensed by the
maintainance of a global hash.


kanaka:
@asarhaddon Merged. Thanks for the detailed commit message. This is a pretty big commit. I generally prefer commits to be broken up into more logical chunks. It both helps me to review changes and serves a pedagogical purpose in that people looking at mal implementations can look at the history for an implementation to see how it was developed. But the quality and of the implementation is primary to the commit history and if you go hunting you'll find some omnibus commits that I made to the project as well. :-) So it's not a strict requirement, but something to consider in the future.

PR #356, opened 2019-05-05
swift4: remove duplicate definition of load-file

asarhaddon:


kanaka:
Thanks for the catch.

PR #357, opened 2019-05-06
Another round of minor doc/test fixes

bjh21:
Hello.  Here are two more fixes for tests that used deferrable features, a fix for an erroneous heading in a test, and a slight improvement to the ordering of the process guide.

PS: My BBC BASIC port has passed step 9, so now I'm busy implementing all the things I deferred.

PR #358, opened 2019-05-06
A couple of extra tests

bjh21:
You might like these tests that I've written to expose bugs in my own implementations. I'm submitting these as a separate PR because there's a danger the bugs will turn out to exist in other implementations too.

### DIFF
--- a/tests/step7_quote.mal
+++ b/tests/step7_quote.mal
@@ -164,6 +164,8 @@ b
 ;=>(1 2 3)
 (concat [1 2] (list 3 4) [5 6])
 ;=>(1 2 3 4 5 6)
+(concat [1 2])
+;=>(1 2)
 
 ;; Testing unquote with vectors
 (def! a 8)
--- a/tests/step8_macros.mal
+++ b/tests/step8_macros.mal
@@ -141,6 +141,8 @@ x
 ;=>()
 (rest [10 11 12])
 ;=>(11 12)
+(rest (cons 10 [11 12]))
+;=>(11 12)
 
 ;; Testing EVAL in vector let*
 


bjh21:
I thought both of those tests would be uncontroversial, but it looks like a most implementations treat `"\"` at the end of a line as equivalent to `"\\"`. I should probably split that test off this PR.

kanaka:
@bjh21 I'm not too surprised that a lot of implementations fail although the fact that only 18 pass is a bit surprising. I agree that the string test should be split off. The way we usually do something like this is to add a comment to the ticket with a check-list of all the implementations that fail and then multiple people can work the problem and check off fixed implementations as they have time. This type of problem is usually pretty quick to fix in each implementation, but it may take a while to fix 56 implementation (76-18).

bjh21:
Right. I've removed the string test and added another vector test that I thought the BBC BASIC port would fail. It didn't, but that's a different story.

kanaka:
Looks like this new test found issues in the following languages (clojure and java failures are for a different reason):

- [ ] nasm
- [ ] rpython
- [ ] rust
- [ ] swift4
- [ ] ts

kanaka:
I added ocaml and ps (postscript) to the list since they fail the `(concat [1 2])` case (they fail to change the type to a list).

bjh21:
It looks to me like most of those failures are actually caused by my previous PR, #357, which replaced `(nth [] 1)` with `(nth () 1)`:

- [nasm](https://travis-ci.org/kanaka/mal/jobs/528862557#L1296): returns `nil`
- [rpython](https://travis-ci.org/kanaka/mal/jobs/528862576#L5040): exits
- [rust](https://travis-ci.org/kanaka/mal/jobs/528862579#L1329): exits (with a panic)
- [swift4](https://travis-ci.org/kanaka/mal/jobs/528862591#L1254): exits (with a fatal error)
- [ts](https://travis-ci.org/kanaka/mal/jobs/528862594#L1366): throws `"Cannot read property 'type' of undefined"`

64c11d8  might want to be reverted, but in the meantime these tests shouldn't hold up merging _this_ PR.

kanaka:
@bjh21 fixed ps here: https://github.com/kanaka/mal/pull/362

kanaka:
I considered reverting 64c11d8 but I think at this point there is enough other churn that I'll leave it there and we'll just fix them. I fixed several last night and pushed them. I looks like crystal has an issue. This run should give a better idea of which implementations have failures with the `(nth () 1)` test. I've filed https://github.com/kanaka/mal/issues/364 to track the empty list bug fixes.

kanaka:
The ocaml fix is pending verification on some other travis jobs, but once it runs and I've verified that ocaml is fixed then I'll have you rebase this pull request so we can test the current state against these changes and make sure that everything passes.

kanaka:
@bjh21 I merged your ocaml change. Please rebase this pull request so we can make sure there are no other issues that might have been hidden by the empty list handling failures.

kanaka:
Thanks. Here is the test run: https://travis-ci.org/kanaka/mal/builds/529530168

dubek:
nasm fails on the test `()` (step 8):

```
TEST: '()' -> ['',()] -> FAIL (line 32):
    Expected : '\\(\\)\r\n\\(\\)'
    Got      : '()\r\nnil'
```

kanaka:
@bjh21 everything passed except nasm (due to the empty list bug) so I've merged.

@dubek yep, that's the unrelated failure caused by me not waiting for checks when I merged https://github.com/kanaka/mal/commit/64c11d81a0735a2d7603770a8fd407c80fae6491. The tracker is https://github.com/kanaka/mal/issues/364 and nasm is the only failure left. @bendudson said he would address that in the next few days (unless somebody else wants to come up to speed on nasm and fix it).

PR #359, opened 2019-05-06
New test for reading an unterminated string that nonetheless ends with a quotation mark

bjh21:
These tests (pulled out of PR #358) are to ensure that implementations correctly see these as unterminated strings:
```
"
"\"
```
Implementations that just check for a string token that ends with `"` will be fooled by these. The following implementations seem to be broken:

- [x] ada
- [x] awk
- [x] bash
- [x] basic
- [x] c
- [x] clojure
- [x] coffee
- [x] cs
- [x] crystal
- [x] d
- [x] dart
- [x] elixir
- [x] elm
- [x] es6
- [x] factor
- [x] fantom
- [x] go
- [x] groovy
- [x] gnu-smalltalk
- [x] guile
- [x] haxe
- [x] hy
- [x] io
- [x] java
- [x] js
- [x] julia
- [x] kotlin
- [x] livescript
- [x] lua
- [x] matlab
- [x] miniMAL
- [x] nim
- [x] objpascal
- [x] objc
- [x] ocaml
- [x] perl
- [x] perl6
- [x] php
- [x] picolisp
- [x] plpgsql
- [x] powershell
- [x] ps
- [x] python
- [x] r
- [x] racket
- [x] rpython
- [x] ruby
- [x] rust
- [x] scala
- [x] scheme
- [x] skew
- [x] swift
- [x] swift3
- [x] tcl
- [x] ts
- [x] vala
- [x] vb
- [x] vhdl
- [x] vimscript
- [x] wasm
- [x] yorick


kanaka:
This is going to be a bit more challenging than I thought due to escaping. For example, the awk implementation does this to make sure the string token ends in a double-quote:

```
if (token ~ /"$/) {
```

I tried a simple change to make sure there is no backslash before the quote:

```
if (token ~ /[^\]"$/) {
```

Which correctly catches the `"\"` case but now errors out on a legitimate case `"\\"`.

And I think a lot of the implementations are going to have this challenge. The current regex suggested by the guide for lexing has the double-quote at the end of the string as optional. This is because it's often more convenient to do the error checking during the parsing (read_atom) rather than during lexing/tokenizer (because many implementations use higher order logic for tokenizer). But that design means the we basically have to re-lex the string again in read_atom to get things right. So the approach that I'm currently experimenting with is to make the ending quote required and detect a failure to tokenize/lex in tokenizer. I'm happy to hear other creative thoughts. But at the least, we should probably share here things that we try that work or don't work so that we can avoid re-inventing the wheel.

bjh21:
A language that has regular expressions can obviously validate the string in `read_atom`, just by re-using the relevant part of the `tokenize` regexp but requiring the closing quotation mark, or by using a simpler one that just counts backslashes (e.g. `[^\\](\\\\)*"$`). See #366 for examples.

Alternatively, if you're parsing out backslash escapes by hand it's easy to check that you haven't run off the end of the string. At least that's what I did in BBC BASIC.

bjh21:
Apologies for the spurious Travis builds. I should have polished the branch a bit more _before_ pushing it.

kanaka:
@bjh21 fixed awk, bash, perl, python, ruby, tcl with https://github.com/kanaka/mal/pull/366

kanaka:
@bjh21 can you update this pull request to include "\\" as a test? We want to make sure that we don't break that case with our fixes.

bjh21:
Done.  I've also added a couple of tests to check the implementations aren't just matching a static string at the end of the input, and rebased to master to pick up the fixes so far.

bjh21:
So as expected, the nineteen-backslash test fails precisely where the single-backslash test failed. The two-backslash test reveals another broken implementation:

- [ ] ada

and the eighteen-backslash test curiously breaks matlab (even though the two-backslash test passes).

kanaka:
Just [pushed fixes](https://github.com/kanaka/mal/commits?author=kanaka&since=2019-07-30&until=2019-07-31) for coffee, dart, elixir, elm, guile, matlab, nim, ps, rpython, swift3, and wasm.

kanaka:
The previous fix failed one of the modes of wasm. The fix for that is: https://github.com/kanaka/mal/commit/ddd00f9d327de92c2553f772df4ca1642e83b8d8

kanaka:
Fix for swift: https://github.com/kanaka/mal/commit/9525bdd1ef2239da1f6c430bb7577b693859abd7

kanaka:
Fixed perl6, r, and skew: https://github.com/kanaka/mal/commit/c780f815bae1d3cc97f25eef78ab5b76a9bd8576

kanaka:
Fixed factor: https://github.com/kanaka/mal/commit/66ab0383061c32ac13817f3214db7cc7ea4918c1

The only one left to fix VHDL. @dubek, you think you might be able to take at VHDL?

dubek:
@kanaka VHDL fixed in #436 .

kanaka:
@dubek thanks!

All the implementations are checked off so I've merged. Here is the travis build: https://travis-ci.org/kanaka/mal/builds/566003824 I wouldn't be surprised if there are new failures, but we'll fix those after the fact.

kanaka:
Actually, all pass. Excellent. Thanks for everyone's work on this!

PR #360, opened 2019-05-07
Convert divergent native functions to built-in functions.

asarhaddon:
See last comment of pull request #354.


kanaka:
The modified implementations pass (there are some unrelated failures) so I've merged it. Thanks for the cleanup.

PR #361, opened 2019-05-07
Draft exercise, native implementations of some built-in functions.

asarhaddon:


kanaka:
This is looking good so far.

docs/exercise.md
```diff
@@ -0,0 +1,49 @@
...
+  form will hide your implementation, so in order to test it, you will
+  need to give it another name and adapt the test accordingly.
+
+- Implement `let*` as a macro that uses `fn*` and recursionn. The same
+  remark applies.
+
+- Implement `apply` as a macro.
+
+- Implement maps using lists.
+  FIXME: Is dissoc use anywhere? It makes this implememtation and the
```
kanaka: dissoc is an optional feature. The test is in step9 in the optional section. So you could make it an optional addition here too.

kanaka: BTW, one way you can implement dissoc is by assoc'ing a replacement value that is a magic delete keyword (e.g. :__..DELETED..__) which allows you to shadow values in the lower levels of the structure. The hash map functions have to detect that and do the right thing. e.g. `(keys ...)` might have to keep track of deleted values as it is scanning the tree and not add those keys when it finds them further down the tree.

kanaka:
You feel like this is ready to merge? Doesn't preclude you from continuing to make more changes but it seems like it's at a useful enough point to merge now.

asarhaddon:
Please merge and I will close the PR. It has been a nice opportunity to learn about recursivity, but I do not intend to dig further for now.

kanaka:
Done. I think this has the potential to be a good resource for other people to learn about the same concepts and the power of Lisp in general. Thanks for working on this.

PR #362, opened 2019-05-07
ps: Fix result of (concat [1 2]), so it returns a list not a vector.

bjh21:
A fix for a bug exposed by #358.


kanaka:
Merge. You may noticed that I cancelled all the tests apart and then restarted ps to check just that implementation.

PR #365, opened 2019-05-08
Cpp ada.2 conformance

asarhaddon:
This PR follows the discussion at #360.

In Ada.2, convert `do` from built-in function to special form.
In C++, convert `map` from native definition to built-in function.
Also in C++, merge native functions and macros, the distinction was artificial.


kanaka:
Looks good and those implementations checked out. Merged.

PR #366, opened 2019-05-08
Some fixes for unterminated strings

bjh21:
These are some initial fixes for the unterminated-string bugs tested for by #359.  The first three are simple regexp matchers using the string part of the `tokenize` regexp.

### DIFF
--- a/awk/reader.awk
+++ b/awk/reader.awk
@@ -18,7 +18,7 @@ function reader_read_atom(token)
 	case /^:/:
 		return ":" token
 	case /^"/:
-		if (token ~ /"$/) {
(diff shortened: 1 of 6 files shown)


bjh21:
Apologies for the spurious Travis builds. I should have finished polishing the branch _before_ pushing it.

kanaka:
@bjh21 okay, all the relevant implementations have passed so I merged this. BTW, for your own pull requests I believe you can cancel travis builds that were triggered by it. What I often do is if I'm wanting to see just results for a few implementations, I cancel a build and then restart just those tasks I'm interested in. I think you can do that with your own pull request triggered builds.

PR #367, opened 2019-05-09
D: Get the code to compile with all three current d compilers

kuettler:
- Work around a D bug to get exception handling going
- Fix imports
- Modernize pointer downcasts

### DIFF
--- a/d/step2_eval.d
+++ b/d/step2_eval.d
@@ -16,28 +16,24 @@ MalType READ(string str)
 
 MalType eval_ast(MalType ast, Env env)
 {
-    if (typeid(ast) == typeid(MalSymbol))
+    if (MalSymbol sym = cast(MalSymbol)ast)
     {
-        MalSymbol sym = verify_cast!MalSymbol(ast);
         auto v = (sym.name in env);
         if (v is null) throw new Exception("'" ~ sym.name ~ "' not found");
         return *v;
     }
-    else if (typeid(ast) == typeid(MalList))
+    else if (auto lst = cast(MalList)ast)
     {
-        auto lst = verify_cast!MalList(ast);
         auto el = array(lst.elements.map!(e => EVAL(e, env)));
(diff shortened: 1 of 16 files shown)


d/step2_eval.d
```diff
@@ -16,28 +16,24 @@ MalType READ(string str)
 
 MalType eval_ast(MalType ast, Env env)
 {
-    if (typeid(ast) == typeid(MalSymbol))
+    if (MalSymbol sym = cast(MalSymbol)ast)
```
dubek: should you use `auto sym = ...` here like you did in the other cases?

kuettler: Yes, absolutely. I tried to stick to the original code here. Better to use `auto`.

kuettler:
Thanks a lot for mal. I enjoy it immensely.

I updated the `Dockerfile` to install all three D compilers (gdc and ldc as provided by Ubuntu 18.04 and the latest dmd.) In addition, the `Makefile` has been enhanced to support either of those compilers.

kanaka:
I pushed a docker image to docker hub with the additions and restarted the travis test and D passed. The `apt-get updates` in the Dockerfile aren't necessary (you can do a full rebuild like this `docker build --no-cache -t kanaka/mal-test-d .`). But I'll clean that up myself because I'm also going to add the other two compilers to the travis build. Thanks for the additions!

kanaka:
@kuettler I made the Dockerfile changes I mentioned (plus making dmd available on the path), and I added the linkage to get all three modes available from the top-level: https://github.com/kanaka/mal/commit/c424b18de1a42081417b375daaab12efd8a22e61 All three are now passing in travis: https://travis-ci.org/kanaka/mal/builds/530892861

kuettler:
This is awesome! Thanks for providing mal. Thanks for the D implementation. This is a fantastic way to learn.

PR #368, opened 2019-05-10
Vala

sgtatham:
Hello. I've written an implementation of mal in Vala, sufficient to pass all the tests and self-host successfully. Would you like it?

Also in this PR, a small typo-fixing patch in `guide.md`.


kanaka:
Awesome! Looks good from an initial brief review. I do notice when running `make perf^vala` that the memory usage continues to grow while perf3 is running so I suspect you may have a memory leak somewhere. This isn't a blocker to merging your implementation but you might want to valgrind your implementation and track it down for a later improvement.

I have pushed out a docker image for vala to docker hub so the next steps to getting this merged are:

- Update the counts at the top of README.md
- Add to .travis.yml to trigger travis build of vala

Once the travis build checks-out with vala as part of the test set, then I'll merge it in. After it's merged if you have a twitter account you can announce it and I'll retweet it. If you don't (or otherwise don't want to) then I'll tweet something to announced the new implementation and count.

sgtatham:
OK, I've made those two changes.

I hadn't even tried running the performance tests, so no wonder I missed that memory issue. I'll look into it, although off the top of my head I'm not sure _how_ you go about debugging memory leaks in a ref-counted language – finding out what leaked will surely be the easy part compared to figuring out what code where should have decremented its ref-count and didn't...

I have no Twitter account, I'm afraid.

kanaka:
@sgtatham I stopped the travis build and restarted just the vala impl and it passed so I've merged it. Congratulations! And yes, it's definitely a challenge finding memory leaks. My general approach is use valgrind on each successive implementation step and see where it starts to leak (that narrows down the functionality greatly that might trigger it). From personal experience the usual suspects for memory leaks are basic types (step1), TCO (step5), swap! (step6), macroexpansion (step8), map/apply (step9),  and metadata (stepA).

kanaka:
Here is the tweet: https://twitter.com/bus_kanaka/status/1126931763345883137

sgtatham:
Thanks for merging it!

Unfortunately, I've looked into the memory leaks now, and ... yuck. It's rather late to spot this, but there's a _big_ leakage problem with this implementation: its entire data representation is completely full of cycles, which plays very badly with the language using reference counts for garbage collection.

Specifically, the top-level environment object contains references to function objects defined in it by the prologue `def!` statements. Each of those function objects contains a reference to the function's own environment, and – of course – each of _those_ contains a reference back to the top-level environment. Bingo, a reference cycle. And of course the same thing will be true _whenever_ you define a function using `def!` so as to store the function value in the same environment that its own env is descended from – i.e. more or less whenever you define one at all.

Well, my original aim was to find out what were the easy mistakes to make when programming in Vala, and I guess in those terms I've been wildly successful!

The very _biggest_ leak has an easy fix, though: that's to do with the default `nil` in the metadata system. So I could at least fix that, and add a note in `vala/README.md` about the more fundamental issue?

sgtatham:
... or I could fix it properly by writing a GC, which I decided might be fun. :-)

kanaka:
Well you achieved that in short order. Well done.

PR #370, opened 2019-05-11
gensym: hide the counter in an environment, define inc in stepA.

asarhaddon:
tests: check that `inc` is present in stepA.
It could be advanced to step4, but this is a lot of work with no benefit.

nasm: split lines in mal_startup_string for readability.
objpascal: remove obsolete .orig file
swift: remove an unneeded line in template
swift4: remove duplicate definition of `or` macro


asarhaddon:
This PR follows the discussion at  #354.
I have cheated by adding `inc` only at stepA, it does not seem to be used in previous steps.
Sorry if I am missing a point in your explanation but:
why not duplicate `not`, `inc``cond`, `gensym` and `or`  in `core.mal` (for tests and examples) and `mal/core.mal` (for self-hosting)? If I understand well your concerns, this would bring the same benefits than a duplication in each implementation.

kanaka:
Thanks for the improvement to gensym! Merged.

Regarding your questions. Part of this stems from the fact that I misnamed the top-level `core.mal`. There is actually a TODO to move the core.mal functionality into a `lib` directory (and as part of that probably split it into multiple files). If I had done that change a long time ago, it would be clearer that `core.mal` is not indended to be core functionality but rather a library of useful macros and functions that can be used with any mal implementation to add some higher level useful functionality that is often found in full Lisp languages.

Moving some of those definitions out of the step files would certainly make it a bit easier to update and maintain multiple mal implementations (although those definitions are just dupes for the most part so not particularly onerous to fix). But I do want to keep the educational/learning aspects of mal as the primary goals even if that means a bit more work maintaining multiple implementations. I want the implementors to finish their implementations with a sense that they implemented everything necessary to run the mal implementation (self-host) without relying on any external Lisp magic. In fact, I've seriously considered describing how to implement `cond` and `or` in the guide (step 8) rather than providing the raw strings. Although, in the end, I decided that might be a bit much to ask of most implementors for whom this might be their first contact with Lisp. However, that part of the guide could certainly use a brief explanation of how those two macros work.

The mal implementations also serve a Rosetta stone of sorts (another manifestation of the educational goals for mal). The `not`, `cond`, `gensym`, `or`, `cond` definitions are currently in the same location of the implementations regardless of whether it is the self-host mal implementation or one of the regular implementations. In other words, the mal implementation of mal does not have any special structure.

Note that my above analysis is not based on strong empirical evidence but my personal intuition after having done a bunch of implementations myself and having watched others do the implementations. I am willing to change my mind if you have reasonable argument that supports the pedagogic goals of mal. I would also want @dubek and @wasamasa to weigh in on a structural change like that.

asarhaddon:
Most of your answer would be welcome as comments on top of `core.mal`. The name would not matter with an explicit description.
As someone who has discovered LISP here, I fully agree that I would never have been able to write `cond` and `or` in the MAL language. An obvious solution is to ask the implementor to define `cond` and `or` as built-in specials in stepA. This is very similar to step4, and would improve the performance for limited implementations. But I guess that you have already considered this option.

PR #371, opened 2019-05-11
Extend core.mal

asarhaddon:
The right fold would be useful for the answers to exercises.


kanaka:
Looks good. I wouldn't be surprised if a few implementations fail due to statically sized slurp/read buffers or due to the extra memory needed to hold the core.mal file in memory while parsing. This might push us to split core.mal into multiple files under `lib/` sooner rather than later. 

A couple of changes:
- typos: Conjonction/Disjonction -> conjunction/disjunction
- Either re-implement foldr to be tail call recursive or just note in the comment for `foldr` that it is not tail call recursive (unlike `reduce`) so it may exhaust stack space when run on large collections.

asarhaddon:
Could you explain how to make `foldr` tail-recursive, or give a reference?

kanaka:
Nice. I was about to answer your question about tail-recursive foldr but looks like you have figured it out.

kanaka:
So basic/qbasic and nasm are failing due to running out of memory during `tests/perf3.mal`. ps is probably failing in step 8 because it truncated core.mal and didn't actually evaluate all the contained forms.

We could tackle each implementation individually, but more problems may crop up as we grow `core.mal`. I think the solution that solves the three immediate problems, relieves pressure on the memory constrained implementations, and gives us the freedom to expand the core library without impacting testing would be to split `core.mal` into multiple files under `lib/`. The documentation you have added is great, but it has significantly increased the size of `core.mal`. Note that there are also some code that lives `examples/` that might also make sense under `lib/` such as `pprint.mal`, `protocols.mal`, `memoize.mal` (although those also contain usage/tests). I think when/if we split core.mal we also want to move out some of the tests that are more about validating the core library routines than testing mal implementations. Those tests could be split out into dedicated files under tests such as `tests/lib_reducers.mal` or maybe just `tests/reducers.mal`. We might want some scaffolding in the top-level Makefile to invoke that can come later. Manual invocation using `runtests.py` would suffice until that lands).

I am motivated to address the core.mal split myself but it will probably be a couple of weeks until I can get around to it. If you want to tackle that rather than (or in addition to fixing the 3 individual implementations) I would support that.

@dubek and @wasamasa: I think we may have discussed this briefly in the past, but does the above proposal to split `core.mal` (the top-level one, not the one in the mal implementation) into separate modules under `lib/` sound reasonable? I think by doing that we can free people up to extend the "core library" of functions available to mal without those additions affecting the normal mal implementation process.

dubek:
Sounds good to me.

I see that as part of the official tests, only step8 and perf1,2,3 load `core.mal`. So they could each load just the part of library that is needed for the tests.

kanaka:
I glanced at a few of the failures and it looks like most are related to path resolution in the perf tests.

kanaka:
Heh, x-post. I cancelled the current build so that the new one will start (it would be nice if Travis automatically stopped when it detects force push to a pull request, but oh well).

kanaka:
Interesting. It looks like postscript/ps has an issue with three top-level comments in a row.

kanaka:
I tracked it down and pushed a fix. Please rebase onto master for a final check.

kanaka:
Also, a couple other minor changes I would like to see before merging:
- rename: `fold.mal` -> `reducers.mal`, `composition.mal` -> `threading.mal`
- Add example of running the tests to the `lib/README.md`
- heavy_computations doesn't feel very lib like and is only current used by tests. I think my preference would be for it to live under `tests/` as a support file for testing.

wasamasa:
@kanaka Sounds good to me.

kanaka:
@asarhaddon Thanks for doing this refactor work! Changes look good and all implementations are passing so I've merged it.

PR #372, opened 2019-05-12
step1: Test that an empty hashmap can pass through read and print.

bjh21:
My BBC BASIC implementation currently fails this, and it turns out that I'm not the only one:
- [x] haskell
- [x] make
- [x] swift3


bjh21:
This may not be testing what I thought it was testing. The make implementation, at least, fails on `{ }` but passes `{}`. It similarly rejects `( )` and `[ ]`.

kanaka:
@bjh21 Interesting. That is not the list of implementations that I expected to fail this way. And I wouldn't have expected a space there to make a difference for any implementations. But this is why we test ... good catch. And yes, apparently testing both `{}` and `{ }` is worthwhile. :-)

kanaka:
Ah, one commonality between those 3 is they don't use the regex tokenizer and have their own bespoke parser. Also I wrote them which is probably more relevant ;-). I just pushed a fix for make and swift3: https://github.com/kanaka/mal/commit/b137ff4f44794d6ea9b6e4df893c9a10e503d64d I'll also fix Haskell but the fix isn't immediately apparent so it will be a bit before I can address it. Feel free to take a stab at it if you want.

kanaka:
@bjh21 okay, Haskell should now be fixed. Can you rebase and add the second test (e.g. `{}`). I'll merge this once Travis checks out.

bjh21:
I've added that one and also ones for `( )` and `[ ]` so that people who implement hash-maps late (like me) can still catch this bug early. Travis is now testing merging that into master.

kanaka:
All greens. Merged.

@bjh21 so how is BBC Basic coming?

bjh21:
I'm currently working on the self-hosting parts of step A. I'm now confident that it's going to work and not be an embarrassment, so I've pushed the current state of it to https://github.com/bjh21/mal/tree/bjh21-bbc-basic. Still a fair amount of work (and polishing) to do, though.

kanaka:
@bjh21 I'm not too concerned about slow for self-hosted mode if it passes everything. Once you're reasonably happy with it, don't hesitate to send a pull request. In the meantime, if you have a Dockerfile that can build and run it, send me a copy so I can build and push out an image. That will allow you to activate and get it testing in Travis too.

kanaka:
@bjh21 I saw that you pushed a Dockerfile. With the following patch I was able to build an image (it gave an error with it) and I pushed it to docker hub:

```patch
--- a/bbc-basic/Dockerfile
+++ b/bbc-basic/Dockerfile
@@ -20,8 +20,6 @@ WORKDIR /mal
 
-RUN sed 's/^deb/deb-src/' \
-    < /etc/apt/sources.list > /etc/apt/sources.list.d/deb-src.list
 RUN apt-get update
 RUN apt-get build-dep -y brandy
 RUN cd /tmp && apt-get source brandy && cd brandy-* && \
```

Assuming that's actually a working image, you should now be able to activate your implementation with Travis and start testing it there too.

bjh21:
That's very odd: if I make that change, I get
```
E: You must put some 'source' URIs in your sources.list
```
Still, your image seems to work, so I'll apply that change to my branch.

kanaka:
Yeah, I got an error with it in there. Probably means ubuntu:18.04 isn't specific enough and we are getting two different base images. Mine is probably older and has the deb-src already. Anyways, shouldn't be a big deal and easy to fix later. Might want to leave that in your file as a commented out item for future reference.

PR #373, opened 2019-05-13
Mark some stepA core functions as not optional.

sgtatham:
I found that all of these were needed for self-hosting: without them,
even trying to run stepA_mal.mal gave me undefined-symbol errors.

This commit edits both guide.md and the stepA tests to move those core
functions out of the optional section.

### DIFF
--- a/process/guide.md
+++ b/process/guide.md
@@ -1547,6 +1547,28 @@ diff -urp ../process/step9_try.txt ../process/stepA_mal.txt
   to print a startup header:
   "(println (str \"Mal [\" \*host-language\* \"]\"))".
 
+* Add the following new core functions:
+  * `time-ms`: takes no arguments and returns the number of
+    milliseconds since epoch (00:00:00 UTC January 1, 1970), or, if
+    not possible, since another point in time (`time-ms` is usually
+    used relatively to measure time durations).  After `time-ms` is
+    implemented, you can run the performance micro-benchmarks by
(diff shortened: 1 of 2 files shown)


bjh21:
I've just run into this, and my approach was the reverse: to make them properly optional by arranging that the self-hosted mal was able to do without them. That could either be by conditionally defining them in `core_ns` (bjh21/mal@1df3e4a) or by wrapping each in a function so that, like `fn?`, their absence would only be detected if the wrapper function was called (bjh21/mal@20cb2b4). The reason I've not submitted a pull request yet is I'm still dithering about which approach is better.  The wrapper function is simpler but adds overhead to every call to these functions.

kanaka:
As @bjh21 identified, these aren't actually used by the mal self-host implementation. I do want to keep these optional so that implementors don't do extra unnecessary work to achieve self-hosting. I'm also a bit leery of adding any more complexity to `mal/core.mal` because the self-hosted mal implementations is already pushing the limits of what some implementations can achieve in terms of memory usage (i.e. C64 BASIC). So I think my preference would be to update the guide to mention that those functions don't need to be implemented, but they do need to have stub functions because the self-hosted mal implementation assumes they exist. The easiest stub is probably just to define those symbols to an existing variable-arity function like `list`.

kanaka:
I consider this addressed by https://github.com/kanaka/mal/pull/375

PR #374, opened 2019-05-13
Vala fixes

sgtatham:
This patch series tidies up various parts of the Vala implementation, of which the most important is that I've added a garbage collector, fixing the memory leak @kanaka pointed out in #368.


kanaka:
Nice work!

PR #375, opened 2019-05-14
Document the requirements for optional step A functions

bjh21:
As discussed in #373, there are several functions that, while marked as optional in step A, do actually need to be defined in order for self-hosting to work, but don't need to do anything useful. This PR documents this situation and adds a test for it.


kanaka:
Yep, that makes it much clearer. Thanks.

PR #376, opened 2019-05-14
step A: Test that metadata doesn't break empty collections.

bjh21:
In the BBC BASIC implementation, the empty list, vector, and hash-map are distinguished values and detected by their address.  Adding metadata to them makes a copy, and then they don't work properly any more.

Let's see if anyone else made that mistake...

### DIFF
--- a/tests/stepA_mal.mal
+++ b/tests/stepA_mal.mal
@@ -222,10 +222,19 @@
 (meta (with-meta [1 2 3] "abc"))
 ;=>"abc"
 
+(with-meta [] "abc")
+;=>[]
+
 (meta (with-meta (list 1 2 3) {"a" 1}))
 ;=>{"a" 1}
 
 (list? (with-meta (list 1 2 3) {"a" 1}))
+;=>true
+
+(with-meta (list) {"a" 1})
+;=>()
+
+(empty? (with-meta (list) {"a" 1}))
 ;=>true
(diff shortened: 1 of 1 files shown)


kanaka:
Haha, I can guess. Wait for it ...

kanaka:
Oh ... I was actually expecting basic to fail. I know there are lurking metadata issues there but I guess this case isn't one of them.

bjh21:
All the new tests are optional, so failures of this test won't cause a Travis failure. The BASIC port already fails several of the metadata tests, and this adds another failure (producing a list that looks like `()` but isn't `empty?`).

kanaka:
@bjh21 oh, of course. I was only looking at the top-level result. Silly me. I've gone ahead and merged it.

BTW, I document here how to download full Travis build logs using the travis ruby gem: https://github.com/kanaka/mal/blob/master/docs/graph/README.md I use that for gathering performance stats but also to look at soft failuring tests which don't trigger an overall failure.

bjh21:
For the record, there are two implementations that passed all the existing metadata-on-collections tests but fail one of the new ones:
* coffee - `(with-meta {} {"a" 1})` returns `{"__meta__" {"a" 1}}`
* wasm - `(empty? (with-meta (list) {"a" 1}))` returns `false`

PR #377, opened 2019-05-14
runtest: fix pre-eval option

asarhaddon:


kanaka:
Looks good. I implemented this a while ago and then ended up going a different direction and so it bit rot. Out of curiosity what are you using it for?

asarhaddon:
To load the redefinitions of built-in in exercises, before running the REGRESS testsuite.

kanaka:
Gotcha. That makes sense.

PR #378, opened 2019-05-14
Test that defining a macro does not mutate an existing function.

asarhaddon:


tests/step8_macros.mal
```diff
@@ -173,3 +173,12 @@ x
 (->> [4] (concat [3]) (concat [2]) rest (concat [1]))
 ;=>(1 3 4)
 
+;; Test that defining a macro does not mutate an existing function.
+(def! f (fn* [x] (number? x)))
+;=>#<function (x) -> (number? x)>
+(defmacro! m f)
+;=>#<macro (x) -> (number? x)>
```
kanaka: The way that macros and functions render is up to the implementation author. I would just remove the two lines that test the rendering of functions and macros (runtest will just consider them passing tests no matter what happens). The following two tests are the interesting part anyways. And that's a nice concise way to distinguish between macros and functions!

kanaka:
I cancelled the Travis run since it's pretty clear that not much is going to pass apart from `ada.2`.

kanaka:
Looking through the failures here jogged my memory. I intentionally left this ambiguous in the process guide and the description can be interpreted as as setting the function to be a macro. My intention is that implementors can defer implementing cloneable functions until they need to support metadata. I would be fine with this being an optional step and test in stepA. I.e. "If you implemented as defmacro! to mutate an existing function without copying it, you can now use the function copying mechanism used for metadata to make functions immutable even in the defmacro! case. ..."

process/guide.md
```diff
@@ -1537,6 +1537,10 @@ diff -urp ../process/step9_try.txt ../process/stepA_mal.txt
     result of reading the next next form (2nd argument) (`read_form`) and the
     next form (1st argument) in that order
     (metadata comes first with the ^ macro and the function second).
+  * If you implemented as `defmacro!` to mutate an existing function
+    without copying it, you can now use the function copying mechanism
+    used for metadata to make functions immutable even in the
+    defmacro! case...
```
kanaka: The "..." was there to just to be a placeholder for more explanatory text. :-) In it's place I would put something like this: "In other words, defmacro! should now make a copy of the function and set the macro property on the copy rather than the original"

kanaka:
For reference, here is the Travis build with the implementations that failed with these tests marked as mandatory: https://travis-ci.org/kanaka/mal/builds/532608339

PR #379, opened 2019-05-15
cs, crystal, es6: Detect more unterminated strings.

bjh21:
Here are three more fixes for #359. I've got Docker working, which makes testing obscure languages easier.

### DIFF
--- a/crystal/reader.cr
+++ b/crystal/reader.cr
@@ -81,11 +81,11 @@ class Reader
     when token == "true"    then true
     when token == "false"   then false
     when token == "nil"     then nil
-    when token[0] == '"'
-      parse_error "expected '\"', got EOF" if token[-1] != '"'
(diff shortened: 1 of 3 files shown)


kanaka:
Yes, it's pretty much a necessity to be able to support this many languages. I assume you've discovered the DOCKERIZE flag to the top-level makefile?

```
make DOCKERIZE=1 test^IMPL^stepX
```

Also, I often launch the repl for a language this way:

```
make DOCKERIZE=1 repl^IMPL^stepX
```

And if I just want a shell within the contain, sometimes this is easier than constructing the raw docker run command line:

```
make docker-shell^IMPL
```

bjh21:
Yes, I found `DOCKERIZE` and `make docker-shell`. The tricky bit was learning enough about Docker to stop it breaking my Wi-Fi.

PR #380, opened 2019-05-16
BBC BASIC V implementation

bjh21:
It's not quite ready for merging yet, but I think this needs its own PR. Previous discussion of this implementation is in #352, #357, and #372.

Things I still want to do:
- [x] Support for long strings on ARM BBC BASIC V
- [x] Sort keys in hashmaps properly
- [x] Disallow metadata on strings
- [x] Encode BASIC/OS errors as hash-maps


kanaka:
Metadata isn't disallowed on strings per the process. It's required on mal functions for self-hosting and suggested for collections. But if it's supported on other types that's fine in my opinion. It's difficult to do that in many languages (compared to compound/collection types) which is why I didn't suggest it.

kanaka:
BTW, I think you easily hold the record for the most number of commits to create an implementation. Which isn't bad at all, just noting it ;-)

bjh21:
The problem with allowing metadata on strings is that it breaks the assumption that each string in `S$()` is referenced by precisely one object in `Z%()`, which in turn mean that entries in `S$()` could get freed while still referenced from `Z%()`.  I could fix that by adding special code to clone string-containing objects, or even support reference-counting strings, but just stopping them having metadata seems the simplest fix for this bug.

Also, I have a kind of ambition to have strings and symbols refer directly to offsets in `S$()` without indirecting through `Z%()`, and that would mean losing the ability to attach metadata to them anyway. Of course, that _would_ require extending the garbage-collector to `S$()`, so maybe going that way wouldn't be much of a problem...

kanaka:
Yes, that makes sense regarding strings. BTW, you're probably aware of this but I don't think Travis will initiate a build while the pull request status is marked "draft".

bjh21:
Right, I think this is now kind of ready.  The changes to support long strings were pretty horrific, but I don't feel strong enough to clean them up right now. All the mandatory tests pass in native and self-hosting modes. The only optional tests that fail are for metadata on single-element hash-maps. Empty hash-maps and larger hash-maps are fine, but if your hash-map has precisely one element it can't have metadata.

Tonight's great triumph, though, is that self-hosting now works under RISC OS:

Screenshot from 2019-05-18 23-12-42

It is _rather_ slow, though.

dubek:
> Tonight's great triumph, though, is that self-hosting now works under RISC OS:

Impressive! Good work.

kanaka:
Very nice! Merged. The final task is to announce it. For a implementation this involved a blog write-up would be awesome if you're up to it. Otherwise just a tweet (which I can do if you don't have an twitter account or otherwise don't want to).

bjh21:
I wasn't quite up to writing lots of words (though Simon pointed out that just joining together all the multi-line comments might work), but I've tweeted a bit of a thread leading up to this: https://twitter.com/bjh21_me_uk/status/1130223250192109568

PR #382, opened 2019-05-17
Unterminated string fixes: clojure, fantom, groovy, haxe, hy, io, java, julia, kotlin, livescript, matlab, miniMAL, objc, plpgsql, scala, vb, vimscript, yorick

bjh21:
Here's another batch of fixes for #359.  They're almost all trivial, with clojure being the only oddity. My laptop is so slow that the Java test times out, but it seems to work in the repl.


kanaka:
Nice work, just 21 left (out of 61).

bjh21:
22: MATLAB still fails on `"\\\\"` (and longer strings of backslash).

kanaka:
Okay, 22 then.

PR #387, opened 2019-05-25
test macroexpand without assuming that quasiquote is a special form

asarhaddon:
This allows to test rewriting quasiquote as a macro, and should
change nothing else in theory.

### DIFF
--- a/tests/step8_macros.mal
+++ b/tests/step8_macros.mal
@@ -12,7 +12,7 @@
 ;=>7
 (unless true 7 8)
 ;=>8
-(defmacro! unless2 (fn* (pred a b) `(if (not ~pred) ~a ~b)))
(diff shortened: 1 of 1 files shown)


kanaka:
Can you expand on this because I don't quite follow. Can you give an example of what this enables? Also, you only changed the definition of unless2 and not unless right above. Is that what you intended?

asarhaddon:
The exercise-native-implementations suggests to implement `quasiquote` as a macro instead of a special form. I am using the existing tests to validate the answers. In step8,`unless2` is used as a test input for `macroexpand`, but if `unless2` itself starts with another macro (`quasiquote`), it is expanded too so the tests fails. `unless` does not matter because no test remarks the change in the output of `(macroexpand (unless ..))`.
The benefit is very small, but also this does not cost much.

kanaka:
Realized I forgot to merge this. I'm fine with this change. Merged.

PR #389, opened 2019-05-27
mal: Use a keyword to link each environment to its outer one.

bjh21:
Heretofore, the mal implementation has linked each environment to its outer environment by defining a special symbol, `--outer--` in that environment.  This causes that symbol to be unusable by programs running under that implementation.  For instance:

```mal
(def! foo 123)
(def! f (fn* (--outer--) foo))
(f)
```
gives an error because overwriting `--outer--` breaks the environment chain.

Happily, mal provides a way around this.  Keywords are guaranteed to work as hashmap keys, so by using a keyword instead of a string as the key for the outer environment, it can be hidden from user code.  This also provides a motivation for keywords, which aren't otherwise used by the mal implementation.

### DIFF
--- a/mal/env.mal
+++ b/mal/env.mal
@@ -12,8 +12,8 @@
 
 (def! new-env (fn* [& args]
   (if (<= (count args) 1)
-    (atom {"--outer--" (first args)})
-    (atom (bind-env {"--outer--" (first args)}
+    (atom {:outer (first args)})
+    (atom (bind-env {:outer (first args)}
                     (nth args 1) (nth args 2))))))
 
 (def! env-find (fn* [env k]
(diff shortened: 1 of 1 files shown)


kanaka:
I'm working on getting self-host testing working in travis. I haven't been running it in a while and I'm trying to get it a bit more automated this time. But before I make changes to the mal implementation itself I want to make sure they don't introduce new issues with self-hosting (I'm sure there are a bunch of issues that have crept in accidentally in the meantime).

bjh21:
Sounds like a good plan: I can imagine that this might break something.

kanaka:
@bjh21 can you rebase onto master so that I can test this with the Travis changes I made to more easily test self-hosted mode?

kanaka:
I cancelled the current test and started this as a self-host test: https://travis-ci.org/kanaka/mal/builds/539399640

kanaka:
@bjh21 there is a single failure. Care to guess which one? ;-) https://travis-ci.org/kanaka/mal/builds/539399640

bjh21:
#393 should fix that.

kanaka:
@bjh21 I merged #393, want to rebase so we can test just bbc-basic one last time?

PR #390, opened 2019-05-27
bbc-basic: Assorted fixes

bjh21:
When I first submitted the BBC BASIC V implementation, it had some fairly grotty code for handling strings over 255 characters long because I'd written that code in a hurry and a grumpy mood. I've now re-written it and also made a few other clean-ups.

I've checked that the self-hosting tests still work.

### DIFF
--- a/bbc-basic/README.md
+++ b/bbc-basic/README.md
@@ -1,6 +1,6 @@
 # Introduction
 
-This is a (partial) implementation of mal in BBC BASIC V.  While there
+This is an implementation of mal in BBC BASIC V.  While there
 is already an implementation of mal in BASIC (in the "basic"
 directory), it's targeted at much more primitive versions of BASIC and
 relies on a pre-processor, both of which make it fairly un-idiomatic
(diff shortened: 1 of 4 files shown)


kanaka:
Looks like the perf tests have an arithmetic overflow now: https://travis-ci.org/kanaka/mal/jobs/537979303#L1619

bjh21:
Curses!  I forgot to try the perf tests (and to be fair I didn't expect them to stress string handling). I'll look into it.

PS: As far as I can tell, “Arithmetic stack overflow” in Brandy is actually memory exhaustion: the BASIC stack and heap have collided.

bjh21:
It turns out that running the perf tests wouldn't have helped me: the problem only manifests with a 64-bit build of Brandy and I usually use a 32-bit build. The underlying problem is that a string array in 64-bit Brandy consumes 16 bytes per entry (as opposed to five bytes in ARM BBC BASIC V).

kanaka:
Looks good. Merged.

PR #391, opened 2019-05-28
Test mutual recursion in let*

asarhaddon:
Detected while testing to rewrite `let*` with `fn*`.

### DIFF
--- a/tests/step4_if_fn_do.mal
+++ b/tests/step4_if_fn_do.mal
@@ -201,9 +201,9 @@ a
 
 
 ;; Testing recursive function in environment.
-(let* [cst (fn* [n] (if (= n 0) nil (cst (- n 1))))] (cst 1))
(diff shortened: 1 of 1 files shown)


tests/step4_if_fn_do.mal
```diff
@@ -203,6 +203,8 @@ a
 ;; Testing recursive function in environment.
 (let* [cst (fn* [n] (if (= n 0) nil (cst (- n 1))))] (cst 1))
 ;=>nil
+(let* [f (fn* [n] (if (= n 0) 0 (g (- n 1)))) g (fn* [n] (f n))] (f 2))
+;=>0
```
bjh21: Vectors are deferrable until step A, so either this test and the previous one should be deferrable, or (better) the vectors should be replaced with lists.

kanaka:
Looks good, thanks.

PR #393, opened 2019-05-30
bbc-basic: Increase Brandy heap size so we run out of memory less

bjh21:
Self-hosting the performance tests turned out to defeat even my careful hand-tweaking of the balance between mal's heap and BASIC's string storage, so I've resorted to the crude approach of simply doubling Brandy's RAM allocation to a whole megabyte. That seems a bit extravagant, but it's Brandy's fault for having such a huge overhead for string arrays.

### DIFF
--- a/bbc-basic/run
+++ b/bbc-basic/run
@@ -1,3 +1,3 @@
 #! /bin/bash
-exec "${BRANDY:-sbrandy}" \
+exec "${BRANDY:-sbrandy}" -size 1024k \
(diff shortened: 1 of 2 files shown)


kanaka:
bbc-basic passed, I've stopped the test for other impls. Merged.

bjh21:
Ahem. This doesn't seem to be merged yet...

kanaka:
oh ... weird. I'm going to blame my mouse. The left button has been acting up lately. ;-) I think it really is now.

Issue #103, opened 2015-11-07
Quasiquotes appear to be broken.


gilch:
Here's a simplified example in Common Lisp of what I'm trying to do.
```Lisp
* (let ((foo 2)) (eval `(let ((foo 1)) (eval `(print ,foo)))))

1
1
* (let ((foo 2)) (eval `(let ((foo 1)) (eval `(print ,,foo)))))

2
2
```
The equivalent in mal:
```
user> (let* [foo 2] (eval `(let* [foo 1] (eval `(prn ~foo)))))
2
nil
user> (let* [foo 2] (eval `(let* [foo 1] (eval `(prn ~~foo)))))
[long traceback omitted]
Exception: 'unquote' not found
```
The first case gets the wrong value at the wrong time. And the second case just fails altogether. Nested quasiquotes are not unusual in macros, but I can't seem to make them work in mal. Is there a different syntax for this I'm not seeing?

I've tested this in the python implementation, but since the code seems to follow the guide, I have to assume it's broken in the other implementations too.

kanaka:
@gilch I'm not too surprised that nested quasiquote is broken. For the most part, mal syntax/features are based on Clojure, so this site is probably a good reference: https://blog.8thlight.com/colin-jones/2012/05/22/quoting-without-confusion.html

From that page, here is a simpler issue with nested quasiquote

```
user> `(list 1 `(2 ~~(- 9 6)) 4)
Error: 'unquote' not found
```

If you want to take a stab at figuring out the issue with the current quasiquote algorithm in one implementation, I could probably port it to most of the other implementations. However, I won't have time to actually figure out the issue any time in the next few months.

dubek:
I modified the Ruby impl to include quasiquoting nesting level. Here's the modified `quasiquote` function code:

```ruby
def quasiquote(ast, nesting = 0)
    if not pair?(ast)
        return List.new [:quote, ast]
    elsif ast[0] == :quasiquote
        return quasiquote(ast[1], nesting + 1)
    elsif ast[0] == :unquote
        if nesting == 0
            return ast[1]
        else
            return quasiquote(ast[1], nesting - 1)
        end
    elsif pair?(ast[0]) && ast[0][0] == :"splice-unquote"
        expanded_list = nesting == 0 ? ast[0][1] : quasiquote(ast[0][1], nesting - 1)
        return List.new [:concat, expanded_list, quasiquote(ast.drop(1), nesting)]
    else
[...]

PR #398, opened 2019-06-04
step 7: Test `([~a]).

bjh21:
I haven't been following #384 very carefully, but I think while @asarhaddon added a workaround in 2dfe777 for guile's problems with expressions like `` `([~a])``, there wasn't any talk of testing for or fixing the bug.  So here's a test for it.

I think the test is correct based on the guide's specification of `quasiquote`, and at least the perl and python implementations agree with me.

Implementations that would need to be fixed to deploy this test:
- [ ] common-lisp
- [ ] elisp
- [ ] guile
- [ ] scheme
(plpgsql failure is unrelated)

### DIFF
--- a/tests/step7_quote.mal
+++ b/tests/step7_quote.mal
@@ -174,6 +174,8 @@ b
 ;=>(1 a 3)
 ;;; TODO: fix this
 ;;;;=>[1 a 3]
+`([~a])
+;=>((8))
 
 ;; Testing splice-unquote with vectors
 (def! c '(1 "b" "d"))


asarhaddon:
I have prepared a bunch of regression tests, but we should agree on the intended behaviour before adding tests. See the discussion at the end of #103.

asarhaddon:
This change is now part of PR #401.

asarhaddon:
Now that PR #401 is merged, I suggest to close this one.

kanaka:
Agreed. Closed.

PR #399, opened 2019-06-05
load-file: Require a newline before adding a paren

Wilfred:
If the last line of a file isn't newline terminated, and ends with a comment, then the closing paren won't close the `(do`.

### DIFF
--- a/process/guide.md
+++ b/process/guide.md
@@ -990,7 +990,7 @@ diff -urp ../process/step5_tco.txt ../process/step6_file.txt
 
 * Define a `load-file` function using mal itself. In your main
   program call the `rep` function with this string:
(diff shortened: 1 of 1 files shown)


asarhaddon:
While at it, why not add a final `nil` at the end of `load-file`? We currently put it manually inside each file.

kanaka:

@Wilfred change seems reasonable. However, none of the implementations currently do this. If you're willing to update the PR to make the change (including the good suggestion from @asarhaddon) to all the implementations, once Travis passes then then I would be happy to merge this.

asarhaddon:
Your idea is implemented as #415.

kanaka:
Resolved by merged of #415

PR #401, opened 2019-06-15
New soft tests, quasiquote algorithm described at  #103 on some impls.

asarhaddon:


scheme/lib/core.sld
```diff
@@ -275,7 +275,7 @@
                      ((func? x)
                       (let ((func (make-func (func-ast x) (func-params x)
                                              (func-env x) (func-fn x))))
-                        (func-macro?-set! func (func-macro? x))
+                        (func-macro?-set! #f)
```
wasamasa: Why would you do that?

asarhaddon: This PR intends to test the quasiquote changes. I have based it on #384 and #400 for convenience, but related discussions should happen at #384.
Anyway...
- this first change is replaced with a more correct ```(func-macro?-set! func #f)``` by a following commit.
- The way Guile and Scheme support metadata for macros breaks macros in the MAL implementation in #384. I have removed this optional feature, and introduced a soft test in case someone wants to work on a real fix. The intent is that a broken optional feature does not block the merge of #384.
However, #400 implements macros in MAL without metadata. If it is accepted, we could as well revert the changes in Guile and Scheme.

wasamasa: I see, thanks.

bbc-basic/step7_quote.bbc
```diff
@@ -53,25 +53,29 @@ END
...
-    IF FNis_symbol(caar%) THEN
-      IF FNunbox_symbol(caar%) = "splice-unquote" THEN
-        =FNalloc_list3(FNalloc_symbol("concat"), FNnth(car%, 1), FNquasiquote(FNrest(ast%)))
-      ENDIF
-    ENDIF
+DEF FNstarts_with(ast%, sym$)
+  LOCAL a0%
+  IF NOT FNis_seq(ast%) THEN =FALSE
+  a0% = FNfirst(arg%)
+  =FNis_symbol(a0%) AND FNunbox_symbol(a0%) = sym$
```
bjh21: I think this is unsafe: `AND` in BBC BASIC doesn't short-circuit, so this will call `FNunbox_symbol` even when `a0%` isn't a symbol, which will cause an error. I _think_ it should be possible to test for this by passing suitable input to `quasiquote` but I'm not in a position to experiment with that right now.

I think you need something like this instead:
```
  IF NOT FNis_symbol(a0%) THEN =FALSE
=FNunbox_symbol(a0$) = sym$
```

bbc-basic/step7_quote.bbc
```diff
@@ -53,25 +53,29 @@ END
...
+  a0% = FNfirst(ast%)
+  =FNis_symbol(a0%) AND FNunbox_symbol(a0%) = sym$
+
+REM  Map quasiquote, except to splice-unquoted elements.
+DEF FNqq_elts(seq%, env%)
+  LOCAL elt%, acc%
+  IF FNis_empty(seq%) THEN =seq%
+  elt% = FNfirst(seq%)
+  acc% = FNqq_elts(FNrest(seq%), env%)
+  IF starts_with(elt%, "splice-unquote") THEN
```
bjh21:
```suggestion
  IF FNstarts_with(elt%, "splice-unquote") THEN
```

bbc-basic/step7_quote.bbc
```diff
@@ -53,25 +53,29 @@ END
...
+    =FNcore_concat(FNEVAL(FNnth(elt%, 1), env%), acc%)
```
bjh21: `FNcore_concat` takes one argument, which is a list of all the sequences to concatenate, because BBC BASIC doesn't have variadic functions. You could assemble an argument list using `FNalloc_list2`, but it's simpler just to use `FNcore_concat1`:
```suggestion
    =FNcore_concat1(FNEVAL(FNnth(elt%, 1), env%), acc%)
```
Really, though, if `FNcore_concat1` is to be called from outside the `core` library it should be moved to `types` and renamed.

asarhaddon: Thanks. Please consider sending me a patch moving `more_concat1` to types, at least if you can test BBC-Basic locally.
I have created the branch `quasiquote-review.` in the `asarhaddon/mal` repository. I will merge it into the PR and trigger all tests again once each build failure has been investigated.

asarhaddon:
Out-of-date implementations are shown by soft-tests, but no hard test fails, and no test at all crashes.
I have rewritten history so that unrelated changes figure in distinct commits.

kanaka:
I had a chance to review this a bit more just now. I'm on board with all the changes except for the "Change quasiquote algorithm" and "soft-test quasiquote more extensively". If you want to send a PR with those separately, I would be happy to merge them in (assuming they pass Travis in separate form).

For the quasiquote change, the behavior change is the right direction. However, my conception of quasiquote is that it should be purely a term rewriting mechanism and not be coupled to the evaluation logic/lexical environments. Also, pushing EVAL down into quasiquote means that it can't participate in the TCO mechanism so it could blow the stack more quickly when quoting large data structures.

asarhaddon:
I understand that `quasiquote` can conviently be interpreted as a rewriting recipe, followed by an interaction with `eval` limited to `quote`, and that implementing it differently can cause confusion. However, it makes the eventual execution path more explicit, and this has already allowed to detect and fix some bugs.
For the performance issue, can you describe a form where the direct interpretation consumes more stack than rewriting followed by evaluation? As I undertand things, my implementation only calls `EVAL` from `quasiquote` in places where the previous implementation would have done the same, when it encounters `unquote` or `splice-unquote`.

kanaka:
Here is a quick hack to measure the stack depth for the two functions of interest (EVAL and quasiquote). The stack is deeper than this but those only two functions that are changed so this should capture the basic difference in stack usage. The diff to the original code/algorithm is basically the same except for the quasiquote arguments.

```patch
@@ -13,7 +13,7 @@ def READ(str):
 def starts_with (ast, sym):
     return types._list_Q (ast) and ast [0] == sym
 
-def quasiquote (ast, env):
+def _quasiquote (ast, env):
     if not types._sequential_Q (ast):
         return ast
     elif starts_with (ast, 'unquote'):
@@ -55,7 +55,7 @@ def eval_ast(ast, env):
     else:
         return ast  # primitive value, return unchanged
 
-def EVAL(ast, env):
+def _EVAL(ast, env):
     while True:
         #print("EVAL %s" % printer._pr_str(ast))
         if not types._list_Q(ast):
@@ -140,6 +140,27 @@ def EVAL(ast, env):
             else:
                 return f(*el[1:])
 
+stack_level = 0
+stack_max_level = 0
+def EVAL(ast, env):
+    global stack_level, stack_max_level
+    stack_level += 1
+    if stack_level > stack_max_level:
+        stack_max_level = stack_level
+    ret = _EVAL(ast, env)
+    stack_level -= 1
+    return ret
+
+def quasiquote(ast, env):
+    global stack_level, stack_max_level
+    stack_level += 1
+    if stack_level > stack_max_level:
+        stack_max_level = stack_level
+    ret = _quasiquote(ast, env)
+    stack_level -= 1
+    return ret
+
+
 # print
 def PRINT(exp):
     return printer._pr_str(exp)
@@ -167,6 +188,7 @@ if len(sys.argv) >= 2:
 # repl loop
 REP("(println (str \"Mal [\" *host-language* \"]\"))")
 while True:
+    print("stack_max_level %d" % stack_max_level)
     try:
         line = mal_readline.readline("user> ")
         if line == None: break
```

And here is the result. While the specific testing code below (nested quasiquote code) is not realistic, due to the pure interpreted nature of mal (no pre-evaluation of quoting, etc), this same effect could happen across functions and macros calls that make use of quasiquoting.

```clojure
$ python python/stepA_mal.original.py
Mal [python]
stack_max_level 3
user> `(1 ~(list 2))
(1 (2))
stack_max_level 4
user> `(1 ~(list `(2 ~(list 3))))
(1 ((2 (3))))
stack_max_level 7
user> `(1 ~(list `(2 ~(list `(3 ~(list 4))))))
(1 ((2 ((3 (4))))))
stack_max_level 10
user> `(1 ~(list `(2 ~(list `(3 ~(list `(4 ~(list 5))))))))
(1 ((2 ((3 ((4 (5))))))))
stack_max_level 13
```

```clojure
$ python python/stepA_mal.with-change.py
Mal [python]
stack_max_level 3
user> `(1 ~(list 2))
(1 (2))
stack_max_level 5
user> `(1 ~(list `(2 ~(list 3))))
(1 ((2 (3))))
stack_max_level 9
user> `(1 ~(list `(2 ~(list `(3 ~(list 4))))))
(1 ((2 ((3 (4))))))
stack_max_level 13
user> `(1 ~(list `(2 ~(list `(3 ~(list `(4 ~(list 5))))))))
(1 ((2 ((3 ((4 (5))))))))
stack_max_level 17
```

Any EVAL that occurs outside of the TCO loop in EVAL itself grows the stack and prevents TCO of that path.

asarhaddon:
OK, I understand the worst case. I would argue that the change is worth the while in normal circumstances, even without a word about speed or memory management.

The two diffs attached below (as .txt because github refuses .diff) add your instrumentation, fix the old behaviour on vectors, and implements `cond` with `quasiquote` in order to make the test more significant. Then

```sh
(cd python && ./run ../tests/perf1.mal)
```
- rewrite   17ms, stack depth 15
- interpret 10ms, stack depth 11

```sh
echo | make MAL_IMPL=python repl^mal
```
- rewrite    max stack level 19
- interpret  17
```sh
time (cd python && for i in alias-hacks equality pprint protocols reducers test_cascade threading trivial; do ./run ../tests/lib/$i.mal; done)
```
- rewrite    7 20 31 16 7 14 9 8 in 0.73s
- interpret 7 18 31 14 7 11 9 8 in 0.56s

[interpreting.txt](https://github.com/kanaka/mal/files/3407198/interpreting.txt)
[rewriting.txt](https://github.com/kanaka/mal/files/3407199/rewriting.txt)

kanaka:
That's a good test and does show that more normal usage patterns are a win for the early interpreting version of quasiquote. We might even want to capture that somewhere as an example for implementors of how to instrument to characterize TCO behavior.

However, I don't want to get too far into the performance weeds because that's not my main concern. My main concern is if this diverges from the typical quasiquote pattern. I did a cursory search of some compilers and interpreters:

- [Racket](https://github.com/racket/racket/blob/5bb837661c12a9752c6a99f952c0e1b267645b33/racket/src/racket/src/qqsrc.rkt#L2) - term rewriting
- [Clojure](https://github.com/clojure/clojure/blob/06097b1369c502090be6489be27cc280633cb1bd/src/jvm/clojure/lang/LispReader.java#L988) - term rewriting with namespace resolution (syntax quote)
- [Pixie](https://github.com/pixie-lang/pixie/blob/d76adb041a4968906bf22575fee7a572596e5796/pixie/vm/reader.py#L448) - term rewriting with namespace resolution (syntax quote)
- [Norvig's lispy.py](https://github.com/norvig/pytudes/blob/master/py/lispy.py#L280) - term rewriting
- [Guile](https://github.com/cky/guile/blob/89ce9fb31b00f1f243fe6f2450db50372cc0b86d/module/ice-9/quasisyntax.scm#L67) - term rewriting
- [picrin-scheme](https://github.com/picrin-scheme/picrin/blob/2af16bc88fbfca2efafd27ca9e5006d73c06f6fb/piclib/eval.scm#L447) - term rewriting
- [SBCL](https://github.com/sbcl/sbcl/blob/825d4d00705eed3a9dc676bd12fa49d384269233/src/code/backq.lisp#L111) - eval referenced but always quoted to evaluation later, so still effectively term rewrite

Most the definitions I'm familiar with treat quasiquote as something that is implemented as part of the reader code base and exposed as a function/macro to user code. But effectively it's reader functionality. The syntax-quote variants of quasiquote are effectively a phase between reader and compilation so that namespaced symbol resolution can occur. In mal, it's implemented in the step files because it's part of the incremental process and also closely related to macros, but logically it could have been in the reader.

To be clear, I'm not dead set on this change but I will need more convincing. Finding a Lisp (that's not just some toy project) that uses this same pattern of early eval in quasiquote would go a long way towards convincing me.

Also, one thing I'm concerned about this approach is that it would make it more complicated to build on mal to create a compiler. I haven't thought through that aspect in depth but I know that the more places where EVAL is called, the more complicated compilation is. Perhaps @dubek could weigh in on that since he created https://github.com/dubek/malc. @dubek I know it's been a while, but can you share any thoughts about what impact direct invocation of EVAL inside quasiquote would have on malc?

dubek:
In malc I took the simplest approach I could think of and just included a whole Mal interpreter just for defining and expanding macros; indeed it uses EVAL in it. The compiled code is not aware of any such thing as macros (also remember that malc doesn't support EVAL in compiled code).

This works as long as the macros are self contained and don't reference any "runtime" state. For example I believe this will fail in malc:

```
(def! f (fn* [x] (+ 1 x)))
(defmacro! m f)
```

Because the `defmacro!` is will be evaluated during compile-time inside a special `*macros-eval-env*` but `f` doesn't exist in that env.

So, regarding the current discussion, malc current approach could use EVAL instead quasiquote; but not sure we can learn from this as for the better approach.

kanaka:
@dubek thanks, I had forgotten the details of how it worked.

asarhaddon:
I have rebased the commits.
First commit introduces new tests. It cannot be applied separately because some implementations crash instead of just failing the new soft tests, but do you agree that the added tests describe the expected behaviour?
Third commit attempts to describe an algorithm based on rewriting but with the correct behaviour. The only solution that I have found requires `apply` in step7 instead of step9. Is this important? Do you have a better idea?

asarhaddon:
Hello. What is blocking this merge request?

kanaka:
@asarhaddon I had interpreted your "most implementations" to mean that you were still working through the implementations. Then my life got extra busy and I never circled around to check in on how things are going. The changes to the current set of implementations look good to me. If you are planning to work through the other implementations at some point (no hurry, just want to make sure things are eventually consistent), then I'm fine with merging it (after rebase and a successful travis).

asarhaddon:
I miss motivation, skill and/or a development environment for the remaining implementations, but I was not expecting this to be a problem. Some tests are already tagged as soft or commented out only because some existing implementations need a fix.

asarhaddon:
Hello. For elisp, emacs25 has introduced a convenient pattern matching system. Looking at other implementations, I guess that it is possible to use something more recent than emacs24 on unbuntu wily/15. Do you agree to change the Docker settings?
Note: in Debian/Ubuntu, versioned packages like emacs25-nox have disappeared in favor of emacs-nox. Ideally, Docker should installl the unversioned package.

asarhaddon:
Instead, I have removed the pattern matching stuff. It was causing too many recursions during self-host tests.

asarhaddon:
Hello.
I have introduced a `quasiquoteexpand` special, showing the result of `quasiquote` before evaluation. This is quite convenient while debugging step7, and I suggest it as an optional feature for new implementors. This may also help understanding the logic of step7. If you disagree, removing it will be trivial.

Most implementations are updated, but some issues remain out of my reach.

- The closure/clj implementation fails during installation of the build system, but the other build system passes all tests so this is probably unrelated.

- The factor implementation fails on `(let* (x 0) (quasiquote (unquote x)))`, despite `(quasiquoteexpand (unquote a))` correctly returning `a`. I fail to understand how my changes interact with environments.

The remaining issues only trigger soft tests. I suggest to merge nevertheless, so that new implementations are correct from the beginning (xslt, jq,swift5, zig, wren, python.2 and pike have been added since this PR is pending).

- The only implementations not using the new algorithm yet are: basic, forth, nasm and wasm.

- The lua implementation of `vec` mutates the original list, though the algorithm is copied from `vector`. The same problem affects `defmacro!`, so it is probably unrelated with this pull request.

asarhaddon:
All implementations are updated. The bug in factor is fixed.

The clojure issue has disappeared, but 3 of 4 wasm docker images now fail during the performance test. This is probably unrelated, as they were OK a few days ago.

Except this, all seems for a merge.

kanaka:
Wow, that was quite the monumental undertaking! I'm reviewing it now. So far looks good.

The wasm failures are probably due to an underlying change in the wasm library that is shared by some of the upstream wasm processors. I'll work on fixing those later.

kanaka:
`quasiquoteexpand` is a great idea (I wish I had though of it earlier, would have save me some headache). The name is wordy, but it's probably best to maintain the parallel naming with `macroexpand`.

Now implemented this in every implementation, you can probably drop this sentence from the process guide:
"In case you compare with existing implementations, be warned that previous versions of this guide were describing `quasiquote` quite differently."

I noticed that you enabled the quasiquote quine test. Apparently enough has changed in the test harness that the line length issue no longer applies. :-)

It's looking good so far after skimming the changes and reviewing a few in more depth. I don't expect to find any blockers and we can address those later. For example, I have some manual tests for finding GC leaks in basic, but those are tedious and I don't have the time to run them right now. When I get around to it, I'll enlist your help if I find any issues and the fixes aren't obvious :-). So if you make the minor change I mentioned above, I'll merge it.

kanaka:
So, I realized that there is a reasonable chance that you may hold the record for the greatest number of unique programming languages that you have made a non-trivial code change to. I would be curious to know any takeaways that you had while doing this. Any new favorite languages? Any languages that were particularly tedious to update? What's your impression of postscript and stack based/concatenative languages? Etc?

Actually, come to think of it, there are probably others that would enjoy hearing about your experience (good or bad). Maybe write a few thoughts in a public post of some sort? I suspect there might be a few conferences (PL, Strange Loop, etc) that might consider a presentation on this type of work interesting. No pressure though. Just thinking out loud.

Anyways, thank you for taking the time to do this. It was definitely one of the more crufty areas of the mal process/implementations. Thank you for improving it! I hope you got more out the process than it took from you. ;-)

kanaka:
Merged!

PR #402, opened 2019-06-22
Unterminated string fix: basic

bjh21:
One small step for #359.

I've essentially translated the BBC BASIC `FNunquote_string` into these more primitive dialects, including writing an implementation of `INSTR` in CBM BASIC because it's not natively present. Having discovered the lack of `INSTR` in CBM BASIC I considered rewriting the code to copy one character at a time into `R$`, but I decided that was likely to be much slower, so I stuck with the ersatz `INSTR`.


kanaka:

Looks good. Merged.

Issue #404, opened 2019-06-25
Typescript implementation error on IE and Edge


sandinosaso:
Hi, thank for sharing this I found a very hard to detect bug in IE10, IE11 and all the Edge versions.

In the typescript implementation for the Env it is used a Map() is used, and the whole MalSymbol object is used as a key for saving on that Map.

When doing `this.data.has(key)` being key an object (the MalSymbol) it returns **false** in IE, Edge even when I see the key is present in the Map object.

I think this may be a difference between the IE and Edge implementations when using objects as keys of a Map object, in Chrome and Firefox it works ok.

Using just the MalSymbol value of a key can fix the problem (this is how the pure JS implementation resolves this).

This implementation has the bug (because it uses a Map with object keys): https://github.com/kanaka/mal/blob/master/ts/env.ts
This is a good implementation: https://github.com/kanaka/mal/blob/master/js/env.js

Best regards.

sandinosaso:
This could be fixed adding a Map() polyfill

PR #405, opened 2019-06-25
Fix error of saving the whole MalSymbol object as a key on the Env

sandinosaso:
Instead of using a Map use a pure object with key values and use the MalSymbol.v as a key (the value) instead of the whole object

This PR fixes a bug reported here: https://github.com/kanaka/mal/issues/404

### DIFF
--- a/ts/env.ts
+++ b/ts/env.ts
@@ -1,10 +1,10 @@
 import { MalType, MalSymbol, MalList } from "./types";
 
 export class Env {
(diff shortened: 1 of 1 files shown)


sandinosaso:
Closing this as I fixed the issue by adding a Map() polyfill

kanaka:

Are you sure you don't want to merge this so that other people get the fix automatically?

PR #407, opened 2019-07-03
Test that macros use closures like functions. Fix ada.2 accordingly.

asarhaddon:


kanaka:

Looks good. Merged.

PR #411, opened 2019-07-07
step 1: Test reading some simple nested collections.

bjh21:
These tests are intended to catch the cases where the routine for reading a list, vector, or hashmap fails to consume the trailing token (`)`, `]`, or `}` as the case may be) from the input.  None of the existing step 1 tests detected this, and I only found it by playing around in the REPL.


kanaka:

Looks good. Merged.

PR #413, opened 2019-07-07
step 1: Test a hash-map with more than one entry.

bjh21:
All the existing tests in step 1 use empty or single-element hash-maps. That avoids any dependency on the ordering of printed hash-maps, but it does mean that some code paths (e.g. printing the space between hash-map elements) aren't tested.  Add a test that creates a three-element hash-map and uses some regexp trickery to check that all the right keys and values come out.


dubek:
Looks good to me. Can we add a comment to the test file explaining this regexp sorcery? (the intention, not its inner implementation details)

bjh21:
Hmm. I've no idea why four of the tests reported problems with parsing the new test. Running the common-lisp test under Docker on my laptop worked fine.

bjh21:
I worked it out on the way to work this morning: Travis pulled from `refs/pull/413/merge`, which isn't a constant commit.  So while the test run is shown against d6383c8, it was actually tracking the commits to the branch, and so it picked up the short-lived commit where I screwed up the comment syntax in test files. Should be nothing to worry about, and re-running those tests should succeed.

kanaka:
@bjh21 this sort of test is part of the reason why I added/fixed the regex support a few months ago. But that regexp is definitely deep magic :-) Merged.

PR #414, opened 2019-07-07
php: Make sure that hashmap keys get returned as strings

bjh21:
This PR fixes a bug I alluded to in #413 whereby when printing a hash-map with keys like `"1"`, or when calling `keys` on it, mal would return `1` rather than `"1"`. This is because of the unique and special way that PHP handles array keys. I also provide tests for the two cases and fix a minor error in a comment that I happened to spot.


kanaka:
Good catch. Change looks fine but due to other merges this now has a conflict. Can you rebase and resolve? Thanks.

bjh21:
Oops. I carefully made sure this didn't conflict with #413, but I forgot about #411. Fixed now (using the GitHub conflict resolver because I was curious).

kanaka:
Thanks. Merged.

PR #415, opened 2019-07-09
load-file: accept final comment without new line, return nil

asarhaddon:
Here is a first attempt to implement the changes discussed at #399 .


kanaka:
Looks like the change is screwing with runtest.py somehow.

kanaka:
I cancelled the Haskell merge that was delaying this. Travis is now executing and it looks like some impls still have issues but at least it's not 100% fails now.

asarhaddon:
My changes crash BASIC, someone more competent is required here.
All other implementations are updated.

kanaka:
For basic, I think it will need to be something like this:

```patch
--- a/basic/reader.in.bas
+++ b/basic/reader.in.bas
@@ -57,8 +57,11 @@ READ_FILE_CHAR:
   IF D$<>"" THEN C$=D$:IF RJ=0 THEN RETURN
   IF D$<>"" AND RJ=1 THEN D$="":RETURN
   D$=""
-  IF EZ>2 THEN C$=""
-  IF EZ=2 THEN C$=")"
+  IF EZ>5 THEN C$=""
+  IF EZ=5 THEN C$=")"
+  IF EZ=4 THEN C$="l"
+  IF EZ=3 THEN C$="i"
+  IF EZ=2 THEN C$="n"
   IF EZ=1 THEN C$=CHR$(10)
   IF EZ>0 THEN EZ=EZ+RJ:RETURN
   #cbm GET#2,C$
```

kanaka:
The actual problem is that the way that basic was getting the command line arguments is via an arguments file that is created by the run script. And it relied on the previous behavior of last value being returned. I just pushed a fix that moves the "(def! -*ARGS*- ...)" down into the args file. If you rebase on master and resolve the merge conflicts I'll merge (assuming everything now passes).

kanaka:
Looks good. Merged.

PR #416, opened 2019-07-13
step 4: Test that (= (list nil) (list)) returns false.

bjh21:
Certain naïve implementations of `=` (like the one I just wrote) will get this wrong, and no existing test at step 4 will catch it.


kanaka:
Thanks. Merged.

Issue #418, opened 2019-07-15
Self-host issues related to 'not' or '=' (fantom, haxe, io, nasm, powershell, r, rexx and vala)


kanaka:
@asarhaddon had a modification to how self-hosting works (captured here:
https://github.com/kanaka/mal/pull/400/commits/809d74cba78643b1a452fe80e810cc3c0ce4f4e6) that triggered some issues in fantom, haxe, io, nasm, powershell, r, rexx and vala. Here is the Travis self-hosted test where the issue manifested:
https://travis-ci.org/kanaka/mal/builds/558757626

asarhaddon:
The MAL implementation has been changed in order to hide these issues. A proper fix has been found for:
- [ ] fantom
- [ ] haxe
- [ ] io
- [ ] nasm
- [x] powershell (34de12452a979a69e0a6ba120922e84d3411862b)
- [ ] r
- [x] rexx (82bc78eb4342737a719c30b402b5ce24d1e010e8)
- [ ] vala

kanaka:
Strangely, the vala issue appears to go away if you make this change:

```
diff --git a/mal/step4_if_fn_do.mal b/mal/step4_if_fn_do.mal
index 05297be..0580ec3 100644
--- a/mal/step4_if_fn_do.mal
+++ b/mal/step4_if_fn_do.mal
@@ -74,6 +74,7 @@
 
 ;; core.mal: defined directly using mal
 (map (fn* [data] (apply env-set repl-env data)) core_ns)
+(env-set repl-env '*ABC* [1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19])
 
 ;; core.mal: defined using the new language itself
 (rep "(def! not (fn* [a] (if a false true)))")
```

If you remove any number from the list then the assertion errors re-appear and things don't work correctly. It appears to me that the issue appears when the number of mal memory objects allocated is below a certain threshold. @sgtatham any insight you can give on why this is happening and how to fix it? Here is what happens if you run without the addition above:

```
$ make DOCKERIZE=1 MAL_IMPL=vala repl^mal^step4
docker run -it --rm -u 1000 -v /data/joelm/personal/programming/mal-other/:/mal -w /mal/mal kanaka/mal-test-mal make step4_if_fn_do.mal
make: 'step4_if_fn_do.mal' is up to date.
REPL implementation mal, step file: mal/step4_if_fn_do.mal
[...]

PR #421, opened 2019-07-16
vala: avoid half-constructed vectors being garbage-collected.

sgtatham:
The clause in eval_ast() which evaluates each element of an input
vector into an output vector was holding the intermediate results in
an ordinary GLib.List, and putting them all into a vector at the end
of the evaluation. But that meant that nothing was preventing all
those values from being garbage-collected half way through.

Now we make an output Mal.Vector at the start of the process, and
point a GC.Root at it to ensure it stays around until we've finished
putting items in it.

This fixes the vala part of #418, I think.

### DIFF
--- a/vala/step2_eval.vala
+++ b/vala/step2_eval.vala
@@ -101,11 +101,12 @@ class Mal.Main : GLib.Object {
             return result;
         }
         if (ast is Mal.Vector) {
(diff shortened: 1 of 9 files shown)


kanaka:
Regular Travis build passed for Vala and I tested self-host locally and it passed. So Merged.

PR #422, opened 2019-07-17
NASM: Fix bug in str when appending empty arrays

bendudson:
The `slurp` function can produce strings which end with an empty Array, if the input file is a multiple of the Array size. When appending such a string, `string_append_string` would keep reading past the end of the string, and continue until it ran out of memory. This fix adds a check for empty Array.

Partly fixes issue #418


kanaka:
Thanks @bendudson. Merged.

PR #423, opened 2019-07-17
Trivial commits from 401

asarhaddon:


kanaka:
Thanks for splitting those out. Looks good. Merged.

PR #425, opened 2019-07-21
Io fixes

dubek:
Few fixes for the Io implementation (in 4 commits for easier review):

* Fix list and vector equality (this is the issue that broke the self-hosting in #418 ); I'll add the new step4 optional tests in a separate PR.
* keyword and symbol accept a keyword/symbol argument (soft fail in step9)
* Don't mutate existing function when defining a macro (soft fail in stepA)
* Atoms can have meta (soft fail in stepA)

### DIFF
--- a/io/MalTypes.io
+++ b/io/MalTypes.io
@@ -30,6 +30,14 @@ MalKeyword := Object clone do (
 
 MalSequential := Object clone do(
     isSequential := method(true)
+    equalSequence := method(other,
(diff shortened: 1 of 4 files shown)


dubek:
I cancelled the full travis build and then restarted the io tests, and they pass: https://travis-ci.org/kanaka/mal/jobs/561626579 .

kanaka:
Thanks. Merged.

PR #426, opened 2019-07-21
tests: step4: Add optional list and vector equality tests

dubek:
These extra optional step4 tests would have caught the io implementation equality bug solved in #425 (specifically solved in commit 8e4de1ed14edf0c980643852d0663afa214e1202 ).

Note that this might reveal other implementations issues, but we won't see it clearly as Travis failures because these are part of the optional tests of step4 (should be hard failures from step9 onwards, IMO).

### DIFF
--- a/tests/step4_if_fn_do.mal
+++ b/tests/step4_if_fn_do.mal
@@ -431,6 +431,8 @@ nil
 ;=>false
 (= :abc ":abc")
 ;=>false
+(= (list :abc) (list :abc))
+;=>true
 
 ;; Testing vector truthiness
 (if [] 7 8)
@@ -464,6 +466,8 @@ nil
 (= [] (list))
 ;=>true
 (= [7 8] [7 8])
+;=>true
+(= [:abc] [:abc])
 ;=>true
 (= (list 1 2) [1 2])
 ;=>true


kanaka:
@dubek Looks good. I've merged it. I really wish that Travis provided a way to easily start a build with custom options. I have a script that I can use for my own branches, but there doesn't appear to be a way to do this for pull requests (either by me or the submitter) :-( Then we could easily test self-host, regress, enable hard tests, etc. Right now what I do for optional/deferrable tests is use the Travis gem to download the logs for a build like this:

```
BUILD=983
for x in $(seq 1 97); do echo ${BUILD}/${x}; mkdir -p logs/${BUILD}; while ! travis logs ${BUILD}.${x} > logs/${BUILD}/${x}; do true; done; done
```

And then I can grep for soft failures.

dubek:
I wonder: all the optional and deferred tests from steps 2-9 should actually be fully implemented by the time stepA is complete. The only optional features for a full mal implementation are the optional features of stepA itself. I suggest adding to `.travis_test.sh` a regression test with the stepA implementation that runs all the step2-9 tests with hard failures on all optional and deferred tests.

kanaka:
That's a good idea. I've pushed a branch to add regression testing of deferrables with hard failures: here is the Travis test: https://travis-ci.org/kanaka/mal/builds/562807161 There are other optionals in prior steps that are also not required for self-hosting. I've just omitted all optionals from the regression test phase.

kanaka:
Assuming it doesn't increase the Travis runtime by too much then I'll merge it in once I've gotten everything to pass. I re-organized a few things and that's probably responsible for most of the failures that are showing so far.

dubek:
I think the timeout for the `io` implementation should be increased to avoid these sporadic failures in step5:

```
TEST: '(foo 10000)' -> ['',0] -> FAIL (line 23):
    Expected : '\\(foo\\ 10000\\)\r\n0'
    Got      : '(def! bar (fn* (n) (if (= n 0) 0 (foo (- n 1)))))\r\n(foo 10000)\r\n#<Function:params=(n)>'
FAILURES:
FAILED TEST (line 23): (foo 10000) -> ['',0]:
    Expected : '\\(foo\\ 10000\\)\r\n0'
    Got      : '(def! bar (fn* (n) (if (= n 0) 0 (foo (- n 1)))))\r\n(foo 10000)\r\n#<Function:params=(n)>'
```

kanaka:
Okay, I think the only remaining implementation with REGRESS issues on that branch is nasm.

@bendudson can you take a look at nasm? It appears that zero argument function calls work in step4 and stop working in step5. E.g. the following passes in step4 but fails in step5 (and onwards):

```
( (fn* () 4) )
```

kanaka:
@dubek Since nasm is the only implementation with an issue, I've gone ahead and pushed this regression change. I update the test timeout for io (and also guile which I think I've seen timeout before). I also fixed a couple of bugs (features that weren't forward ported to later steps) discovered in dart and factor because of the regression testing. I file a separate ticket for the nasm issue.

PR #428, opened 2019-07-23
io: Fix unneeded escaping of forward slash

dubek:
Instead of using Io's `asJson` method which escapes forward slashes, implement our own string escaping code so it fits the Mal requirements (escape only double-quotes, newline, and backslash).

The relevant step1 test was modified from soft to hard.


kanaka:
Thanks. Merged.

PR #430, opened 2019-07-25
perl: Idiomaticity improvements (and some bug fixes too)

bjh21:
This PR attempts to make the Perl implementation a little more idiomatic. For many years, Perl was my main programming language, and the implementation gave me the feeling of being written by a good programmer but not a native speaker of Perl. I've tried to make it more like I would have written it.

Major changes include:
* Core functions now get their arguments as arguments rather than as a List stuffed into the first argument.
* Various classes of mal value now have operator overloads so that they can be treated as though they were hash, array, code, or scalar references as appropriate.
* Core functions are now blessed coderefs rather than unblessed ones.
* The types of objects are checked using the `isa` method rather than `ref`.

There are still things to be cleaned up, but I think this PR is already scary enough.


kanaka:
@bjh21 "...the implementation gave me the feeling of being written by a good programmer but not a native speaker of Perl". Thanks, I'll take that as a compliment :-D TMTOWTDI. When I did the perl implementation it had probably been a good 12 years since I had written perl in anger. 

Looks much better than it did. Thanks! And congratulations on accepting the role as the primary maintainer of the perl implementation ;-)

PR #432, opened 2019-07-27
More perl improvements

bjh21:
My appointment in #430 has gone to my head, so here's another batch of improvements to the perl implementation.  The highlights:
* All class names now begin with `Mal::`, so won't conflict with perl's built-in `Symbol` and `Env`.
* More `foreach`, less `for (;;)`.
* I've discovered the `//` operator and used it in place of `||` where it makes sense.
* Better interop, including exception handling.
* All scalar types now share an implementation.
* The `try*` implementation is rather simpler.


kanaka:
Nice work. Merged.

Issue #429, opened 2019-07-25
Nasm fails on zero argument function call after step4


kanaka:
This was discussed in https://github.com/kanaka/mal/pull/426 but since that issue is unrelated and already closed I'm opening this to track the issue.

The following succeeds in nasm step4 but fails in step5-A:

```
( (fn* () 4) )
```

@bendudson, can you take a look at this? Thanks.

bendudson:
I've had a look and can't see yet where the problem is, but am getting close (I think). The following works, for example:
```
(def! f (fn* () 4))
(f)
```
I think something is being freed which shouldn't be (connected to the TCO optimisation), with maybe another bug complicating things. I'll try to work this out soon.

kanaka:
@bendudson thanks. That's sounds plausible. And yeah the interaction between TCO and GC is definitely a pain.

kanaka:
Resolved by #433

PR #433, opened 2019-07-28
Fix nasm zero argument function call

bendudson:
Releasing function object prematurely led to segfaults after step 4 when evaluating:
```
((fn* () 4))
```
Now works on steps 5 to A.

Fixes issue #429


kanaka:
Thanks! Merged.

PR #435, opened 2019-07-30
perl: New metadata and TCO and other changes

bjh21:
The whole reason why I was looking at the Perl implementation was that I wondered whether it used `goto &name`. Here I finally get around to making it do so.

# New metadata

The first major change here is to replace the metadata implementation, symbolised by 28bff72. This uses a `Hash::Util::fieldhash` to hold all the metadata in a hash indexed by object ID, with automatic deletion when objects are garbage-collected. Along with this, most of the composite types lose a layer of indirection (771045a, d64a552, 329c9c2). This has a couple of down-sides:

* Turning a function into a macro causes it to lose its metadata
* Attaching metadata to a sequence or hash-map causes the whole array or hash to be copied, which might be expensive.

On the other hand, this removes a hash lookup from every access to a list, array, hash-map, or atom, which I suspect is a win.

# New TCO

Perl has native support for tail calls. `goto &NAME` calls the subroutine `NAME` in place of the currently-executing subroutine without consuming additional stack. Using this feature (c0d61c2)  makes TCO in mal a lot simpler and slightly more powerful:

* There's no need for a loop in `EVAL`: tail recursion is just `goto &EVAL`.
* All functions can be Perl closures, since it's possible to tail-call into one (`goto &$f`) and back out again (`goto &EVAL`).
* TCO can work in `apply`, which it didn't before.

This does mean that the perl implementation doesn't quite follow the process guide. I think this is reasonable for two reasons:

* The overall structure of the code hasn't changed: everything's still in the same place as in other implementations, so it's easy to find the code you're looking for, it just works a bit differently.
* Support for explicit tail calls is a rare enough language feature that showing it off is interesting.

# Others

There are a few other changes here, some of which depend on the above. The major ones are:

* The `ismacro` flag is now represented by having separate `Mal::Function` and `Mal::Macro` classes.
* All checks for class membership now use `->isa` directly, which is the native Perl way to do that.

### DIFF
--- a/perl/core.pm
+++ b/perl/core.pm
@@ -152,8 +152,7 @@ sub with_meta {
 # Atom functions
 sub swap_BANG {
     my ($atm,$f,@args) = @_;
-    unshift @args, $$atm;
-    return $$atm = &$f(@args);
+    return $$atm = &$f($$atm, @args);
 }
 
 
--- a/perl/core.pm
+++ b/perl/core.pm
@@ -3,6 +3,7 @@ use strict;
 use warnings;
 
 use Data::Dumper;
+use Hash::Util qw(fieldhash);
 use List::Util qw(pairmap);
 use Time::HiRes qw(time);
 
@@ -140,11 +141,12 @@ sub seq {
     }
 }
 
+fieldhash my %meta;
+
 # Metadata functions
 sub with_meta {
-    no overloading '%{}';
     my $new_obj = _clone($_[0]);
-    $new_obj->{meta} = $_[1];
+    $meta{$new_obj} = $_[1];
     return $new_obj;
 }
 
@@ -226,7 +228,7 @@ sub pl_STAR {
     'seq'         => \&seq,
 
     'with-meta'   => \&with_meta,
-    'meta'        => sub { $_[0]->meta },
+    'meta'        => sub { $meta{$_[0]} // $nil },
     'atom'        => sub { Mal::Atom->new($_[0]) },
     'atom?'       => sub { _atom_Q($_[0]) ? $true : $false },
     'deref'       => sub { ${$_[0]} },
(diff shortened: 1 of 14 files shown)


kanaka:
I think that's perfectly acceptable to use native TCO to accomplish TCO in the implementation. There are a number of implementations that do that. I would much rather have an idiomatic implementation than one that follows the minute details of the process guide. My main concern is the high-level code organization (env, core, step files, eval_ast, EVAL, etc).  FYI, the main thing in the past that I have asked implementors to change in order to align with the process structure is that there is a temptation to move parts of the EVAL/eval_ast function out into a separate file that is shared by multiple steps. There is nothing wrong with that approach, but it makes implementations much harder to compare and so it's one of the areas where I will definitely request changes before merging.

Anyways, the tests pass, I reviewed the changes and they look great. Merged. Thanks for your work on this. Any other languages you know really well that you would like to clean up and make more idiomatic? ;-)

kanaka:
Oh, and thanks for the excellent write-up. That helps a lot when I am reviewing the code. It might even be worth putting some of this in perl/README.md file if you're willing.

bjh21:
I think the only other language I know well enough is C, and my kind of C (mostly in the NetBSD kernel and the portable parts of PuTTY) is quite a long way from the existing C implementation, which is more of a Glib implementation. I have been thinking about the possibility of a new plain ISO C implementation though. Maybe after I finish BCPL (at which point C will seem luxurious)...

I'd been thinking that the Perl implementation could do with some more comments and documentation. Maybe my next PR will include some.

kanaka:
BCPL, nice. How far along are you? Yes, the C implementation is definitely fairly high-level and heavily dependent on both glib and Boehm for automatic GC. I'm not even sure it builds anymore if you update the libraries due to a conflict between the Boehm GC and glib. I would be happy to replace it with a more standalone ISO C implementation. I actually have a VERY low-level C implementation that is 90% complete that I may get around to finishing at some point. But it wouldn't make a good representative C implementation because it's so low-level (it's designed to be linked with github.com/kanaka/fooboot and booted as a standalone kernel) so something more standard would be great if you get around to it.

PR #439, opened 2019-08-04
Final (?) perl improvements

bjh21:
I think I've reached the point where the changes I'm making to the perl implementation aren't actually making it any better, so here are the lest few that I think are actually good. I found a nicer way to implement keywords, eliminated the numbered variables in the special form switch, and documented things a little.


kanaka:
Looks good. Merged.

PR #440, opened 2019-08-07
step 4: Test (if nil 1) returns nil

aasimk2000:
Certain implementations like the one I wrote can get this wrong.

### DIFF
--- a/tests/step4_if_fn_do.mal
+++ b/tests/step4_if_fn_do.mal
@@ -50,6 +50,8 @@
 ;; Testing 1-way if form
 (if false (+ 1 7))
 ;=>nil
+(if nil 8)
+;=>nil
 (if nil 8 7)
 ;=>7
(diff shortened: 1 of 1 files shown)


kanaka:
Looks like all the existing implementations pass the test. But it's certainly a reasonable test to have so I've merged it.

aasimk2000:
Yeah, I'm aware. I didn't catch this bug in my implementation till self hosting. It took forever to debug. Thought I would add this in to help avoid some frustration for others.

kanaka:
@aasimk2000 yes, there are at least 20 tests cases that are the result of me going through that exact same pain. It's very hard to debug when self-hosting. So I tried to be diligent about making sure that once I debugged the issue, I captured the simple version of the problem (and it's almost always a simple issue) as a test so that it's caught before you reach the self-hosting stage. So thanks!

PR #441, opened 2019-08-11
mal implementation in Red

ALANVF:
This took a while to make, but it turned out pretty well. Also, I used Red's "parse dialect" (which is similar to PEG parsers) because Red doesn't support regex. I'm not sure if that is ok, but it ended up working anyways.


wasamasa:
Hey Alan! Using the parse dialect is a great idea, there's several implementations going for similar solutions like using a hand-rolled lexer (Scheme, Picolisp), or a library equivalent that's as powerful as regular expressions (Haskell). I'm not terribly familiar with REBOL and Red, from a brief glance the implementation looks fine to me.

kanaka:
@ALANVF awesome! I'll take a look tomorrow.

ALANVF:
Sounds good 👍

dubek:
Looks very good. Thanks for the contribution.

I see that `or` and `gensym` macros are still there, though we moved them to a library or something like that (there were several changes, I'm a bit confused). Maybe we forgot to update the guide accordingly?

red/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+RUN mkdir -p /mal
+WORKDIR /mal
+
+##########################################################
+# Specific implementation requirements
+##########################################################
+
+RUN dpkg --add-architecture i386
+RUN apt-get update
```
dubek: I don't think another `apt-get update` is needed after the one in line 8 above.

red/Makefile
```diff
@@ -0,0 +1,11 @@
...
+		step5_tco.red step6_file.red step7_quote.red step8_macros.red step9_try.red stepA_mal.red
+
+all:
+	$(patsubst %.red,%,$(STEPS))
+
+step%: step%.red
+	red-064 -c $<
+
+clean:
+	rm -f $(STEPS:%.red=%) *.red mal
```
dubek: clean removes `*.red`? that sounds bad.

red/Makefile
```diff
@@ -0,0 +1,11 @@
+STEPS = step0_rep.redl step1_read_print.red step2_eval.red step3_env.red step4_if_fn_do.red \
+		step5_tco.red step6_file.red step7_quote.red step8_macros.red step9_try.red stepA_mal.red
+
+all:
+	$(patsubst %.red,%,$(STEPS))
+
```
dubek: Consider adding the `stats*` and `mal` targets as in other implementations. @kanaka do we have instructions for these?

red/Dockerfile
dubek: Ah, maybe needed after the `dpkg` to load the i386 packages?

red/Makefile
ALANVF: I wasn't quite sure how this was supposed to work, so I used the one from pascal as reference

red/Makefile
kanaka: @dubek the stats mechanism now happens at the top-level using loccount so implementations no longer have `stats*` targets. The `mal` targets were really more of an experiment on my part to see how many implementations could be fully packaged into standalone executables. It can be really difficult depending on the target language. So I never documented it anywhere.

red/Makefile
kanaka: @ALANVF `*.red` there means clean will remove all your source files. The intention of clean is to remove all generated files, not source files.

red/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN apt-get -y install curl libreadline-dev libedit-dev
+
+RUN mkdir -p /mal
+WORKDIR /mal
+
+##########################################################
+# Specific implementation requirements
+##########################################################
+
+RUN dpkg --add-architecture i386
```
kanaka: Does this mean the red system is only 32-bit at the moment? That's fine, just seems a bit odd that a modern programming language would not have 64-bit support.

ALANVF: Yes, Red is only 32-bit for now, but the devs have said that it'll get 64-bit support by the end of the year.

red/Makefile
ALANVF: Ah that makes more sense. I saw it in the pascal makefile so I assumed that I needed it.

kanaka:
@ALANVF it's looking really close. I attempted to build a docker image but ran into issues. Here are the changes I needed to make to the Dockerfile to make it build:

```
--- a/red/Dockerfile
+++ b/red/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu
+FROM ubuntu:bionic
 
 # General requirements for testing or common across many
@@ -26,7 +26,7 @@ RUN apt-get install -y \
        libc6:i386 \
        libcurl3:i386 \
        wget
-RUN wget http://static.red-lang.org/dl/linux/red-064
-RUN chmod +x red-064
+RUN wget http://static.red-lang.org/dl/linux/red-064 -O /usr/bin/red-064
+RUN chmod +x /usr/bin/red-064
```

Also, I noticed that there are Windows line endings in the Dockerfile. I've pushed a docker image based on my changes so that Travis tests for red will run.

I ran into build issues when I tried to build. The first was this:

```
$ make step0_repl     
red-064 -c step0_repl.red
** Script Error: Invalid compressed data - problem: -3
** Near: script: decapsulate 
if none? script
```

Apparently red-064 doesn't like running from make. If I run the same command standalone things work better but then I run into what looks like a missing dependency issue:

```
$ red-064 -c step0_repl.red

-=== Red Compiler 0.6.4 ===- 

Compiling /mal/red/step0_repl.red ...
Compiling libRedRT...
...compilation time : 1250 ms

Compiling to native code...
...compilation time : 36851 ms
...linking time     : 489 ms
...output file size : 656932 bytes
...output file      : /mal/red/libRedRT.so 

*** Compilation Error: undefined word ask 
*** in file: /mal/red/step0_repl.red
*** near: [ask "user> " [
        print rep ui
    ] [
        quit/return 0
    ]]
```

So those issues will need to be resolved before merging.

The use of alternate parsing methods than the regex method described in the guide is totally fine. Even recommended if the language has a good idiomatic way of doing it. The main thing I require before merging (apart from all the tests passing) is that the code is generally organized according to the step-by-step process so that each step cleanly builds on the previous step and introduces on the functionality covered in the guide for that step. The reason this is important is that one of primary goals of mal is that somebody who wants to learn a language can look at the mal implementation for a language and starting from each step file can see the logical progression of more and more functionality added. The red implementation is pretty close but I notice that `printer.red` is including `malfunc.red`. The code in `malfunc.red` is really step 5 functionality (or step 4 at the earliest depending on limitations of the underlying implementation language). However, `printer.red` includes `malfunc.red` which means that steps 1, 2, and 3 and now depending on step 4/5 functionality. I think my preference would be if `malfunc.red` code was included in `types.red`. This doesn't technically change the actual code dependencies, but reader and printer are much more strongly associated with step 1 functionality whereas the types namespace is where the low-level details of how the underlying type system is elevated to fit the mal model reside.

So I think the remaining work before this is merged:
* update the Dockerfile
* fix make/build bugs and get it building/running successfully via docker/travis
* remove the `or` and `gensym` code as mentioned by @dubek 
* see if there is a more modular way to handle `malfunc.red` and if not move it into `types.red`
* update the implementation count at the top of the README.

kanaka:
I just noticed the Makefile needs a fix:

```patch
--- a/red/Makefile
+++ b/red/Makefile
@@ -1,8 +1,7 @@
 STEPS = step0_repl.red step1_read_print.red step2_eval.red step3_env.red step4_if_fn_do.red \
                step5_tco.red step6_file.red step7_quote.red step8_macros.red step9_try.red stepA_mal.red
 
-all:
-       $(patsubst %.red,%,$(STEPS))
+all: $(patsubst %.red,%,$(STEPS))
 
 step%: step%.red
        red-064 -c $<
```

ALANVF:
I should also probably just have it run in eval mode, now realizing that there are some issues when compiling and running red.

kanaka:
Using interpreted/eval mode is fine although it does that mean the red implementation will suffer in terms of the microbenchmarks (i.e. this https://kanaka.github.io/mal/graph/) Or perhaps running it that way just means red will compile the program every time instead of just once.

Also, I think you intended to modify the `run` script rather than the Makefile. The change to the makefile just makes each step of the red process run (and fail) during the build process and doesn't change the runtime behavior. If your Makefile change was attempting to disable builds, then what you really want is this:

```
--- a/red/Makefile
+++ b/red/Makefile
@@ -1,11 +1,10 @@
 STEPS = step0_repl.red step1_read_print.red step2_eval.red step3_env.red step4_if_fn_do.red \
                step5_tco.red step6_file.red step7_quote.red step8_macros.red step9_try.red stepA_mal.red
 
all: $(patsubst %.red,%,$(STEPS))
 
-step%: step%.red
-       red-064 -c $<
+$(STEPS):
+       true
 
 clean:
        rm -f $(STEPS:%.red=%) mal
```

That will allow the top-level Makefile (at the mal root) to test to make sure the step files are there but won't actually build anything.

kanaka:
Here is the Travis result of the current changes: https://travis-ci.org/kanaka/mal/jobs/570966567

kanaka:
@ALANVF do you have a Linux VM you can test this with? I noticed a couple other problems that are likely related to build/testing on Windows. The `red/run` script has two problem that keep it from running: a Windows linux ending on the shebang line and it is also not marked executable.

In addition, it seems like the red-064 program might have issues with UNIX job control. When I remove the "exec" from the run script it allows step starts up correctly although removing the exec means every instance now has two processes (the parent bash script and the red-064 instance itself) which is not ideal.

Also, I don't seem to be able to exit from a running implementation. Neither Ctrl-D nor Ctrl-D terminates the program. I suspect this just means that the repl loop isn't checking for an EOF situation.

ALANVF:
@kanaka no I do not, but I can set one up or use a computer I already have. I did make this on Windows, so I was expecting there to be an issue or two (not this many though). As for the EOF issue, I must have forgotten to have it exit after running the file. I apologize for how many issues this has caused. It worked fine for me, so I assumed it would work here without much of a problem

kanaka:
@ALANVF no need to apologize. This actually isn't all that uncommon with new implementation to have to deal with differences in the development environment between the implementor, my own dev environment, and Travis. Keep pushing through, we'll figure it out eventually.

I think the 'quit' that you added is a good fix, but I think it's actually not what I was running into. I was using the REPL and unable to exit and wasn't running a mal program with the implementation (which is the case you fixed). I think the ask loop needs to detect an EOF condition and exit rather than continuing to loop. BTW, since this is your pull request I think you should be able to stop the Travis build and the restart just the red task. That way you don't have to wait for all the other implementations in front of red.

ALANVF:
@kanaka thanks, I was worried that it looked really bad at first. As for the repl thing, I wasn't actually sure whether or not to break the loop, so that's an easy fix. (and yeah, I didn't even think about only running the red task)

ALANVF:
As a side note, it appears that I'm not able to stop the build nor restart any of the tasks

kanaka:
@ALANVF okay, I thought the creator of the pull request could start/stop the tests but perhaps it requires some sort of access to the upstream mal repo.

The following changes still need to be applied.

```
--- a/Makefile
+++ b/Makefile
@@ -243,7 +243,7 @@ ps_STEP_TO_PROG =      ps/$($(1)).ps
 python_STEP_TO_PROG =  python/$($(1)).py
 r_STEP_TO_PROG =       r/$($(1)).r
 racket_STEP_TO_PROG =  racket/$($(1)).rkt
-red_STEP_TO_PROG =     red/$($(1))
+red_STEP_TO_PROG =     red/$($(1)).red
 rexx_STEP_TO_PROG =    rexx/$($(1)).rexxpp
 rpython_STEP_TO_PROG = rpython/$($(1))
 ruby_STEP_TO_PROG =    ruby/$($(1)).rb
diff --git a/red/Makefile b/red/Makefile
index c6162e1..24a5bc1 100644
--- a/red/Makefile
+++ b/red/Makefile
@@ -3,8 +3,8 @@ STEPS = step0_repl.red step1_read_print.red step2_eval.red step3_env.red step4_i
 
 all: $(patsubst %.red,%,$(STEPS))
 
-step%: step%.red
-       red-064 $<
+$(STEPS):
+       true
 
 clean:
        rm -f mal
diff --git a/red/run b/red/run
old mode 100644
new mode 100755
index 1df9ad6..16ad3af
--- a/red/run
+++ b/red/run
@@ -1,2 +1,4 @@
 #!/bin/bash
-exec red-064 $(dirname $0)/${STEP:-stepA_mal}.red "${@}"
+red-064 $(dirname $0)/${STEP:-stepA_mal}.red "${@}"
```

If you are able to fire up a Linux system with docker then I think that would help. If you can get the following command to work from the top-level then I think you will be 99% of the way to getting things working Travis:

```
make DOCKERIZE=1 test^red
```

ALANVF:
I'll attempt to set up docker and make those changes. Hopefully this fixes it

kanaka:
@ALANVF BTW, if you have multiple changes, if you push them together it will only trigger a single Travis build. I've cancelled the first Travis build so that we can see the full results sooner.

kanaka:
So there are two failures in the most recent Travis build: https://travis-ci.org/kanaka/mal/jobs/571365821

The first problem is because there is actually another change needed to the Makefile (the all rule should point to the sources, not the old build targets):

```patch
--- a/red/Makefile
+++ b/red/Makefile
@@ -1,10 +1,10 @@
 STEPS = step0_rep.redl step1_read_print.red step2_eval.red step3_env.red step4_if_fn_do.red \
                step5_tco.red step6_file.red step7_quote.red step8_macros.red step9_try.red stepA_mal.red
 
-all: $(patsubst %.red,%,$(STEPS))
+all: $(STEPS)^M
 
 $(STEPS):
        true
 
 clean:
```

The second problem is because the run script is still not executable. IIRC some git clients on Windows have trouble properly marking a file as executable.

Once you resolve those, it will still fail because the test runner won't get output from the program and it will just hang. When interactive input works but the test runner doesn't, that's usually an indication that the underlying implementation hasn't implemented readline functionality correctly. You might need to support a command line option that disables readline/line editing functionality. Other implementations that need this hack include: cs, fsharp, tcl, vb.

Also, I should note that Ctrl-D and Ctrl-C still don't terminate the interactive REPL. It prints "q" and then loops and prints the prompt again. I do notice though that hitting enter without typing anything does actually terminate the REPL. That's not correct behavior either. An empty line should not terminate the REPL.

You should be able to see all these problems if you are able to test using docker on Linux.

ALANVF:
I realized that the "all" rule needed to be changed after I pushed, and I'm not sure how to mark it as executable. Red has readline functionality, but I just realized that I forgot to add the `--cli` option, which makes it run in the console instead of Red's custom console. I'm not sure why it prints "q" instead of quitting, but I'll investigate and update once everything works correctly

ALANVF:
Until version 0.7.0, Red will not have proper io support, so the best I can do is have it exit on `"q"`, although that means that typing a normal "q" also exits the repl

kanaka:
That's super unfortunate. I don't consider that a blocker to merge into the tree as long as you'll  update/fix it once the I/O support is in red. Any chance you want to explore direct FFI calls into readline or libedit? A number of implementation do that. The main complication I think would be supporting both Windows and Linux (although perhaps you could detect Windows and just use ask in that case). If one of your goals is learning Red more deeply then FFI is certainly a worthwhile area to learn. For example, here is an example of FFI calling into the curses library: https://github.com/red/code/tree/master/Library/Curses.

ALANVF:
I could definitely try to do that, however it probably wouldn't happen right away (as I have a busy schedule coming up). For now, however, would the suggest I gave work for now?

kanaka:
Sorry, to be clear, it doesn't need to respond to EOF to be merged, but it does need to pass Travis tests and it's hanging right now even for step0. In the debug log I only see this:

```
user> abcABC123<no newline>
```

It should look like this if it's working correctly:

```
user> abcABC123
abcABC123
user> hello mal world
hello mal world
...
```

This is indicative of a terminal handling issue in red. The test runner tries to simulate an interactive terminal, but poorly implemented readline implementations can break the test runner/simulated pty. As I mentioned above, the .NET based implementations and TCL have this issue and required a --raw command line argument to switch into plain input/output (instead of line editing mode). It doesn't sound like that option exists with red (since there is no lower-level raw I/O), so implementing FFI to call readline/lineedit might actually be the easiest route to getting this working.

ALANVF:
Well part of the problem may have also been that it wasn't running in the console, but Red's custom console. I'll attempt to make FFI bindings for readline fot now

wasamasa:
You might find my code with Red useful for that: https://github.com/wasamasa/mal-candidates/tree/master/red

kanaka:
@wasamasa nice! I didn't even think to look at your candidates repo to see if you had solved that issue already.

ALANVF:
@wasamasa Yes that works, thanks!

kanaka:
Looks like you may have manually pulled in some changes that I made to the README and that's causing a merge conflict which also seems to be preventing Travis from running.

I tried using the current changes and Ctrl-D/EOF now works to exit the interactive REPL, however it still hangs when running the tests. Even with the `--cli` mode, the ask command still does something weird that breaks `runtest.py`. Are you working on integrating the readline support as demonstrated in @wasamasa's candidates repo?

wasamasa:
I'm afraid you can't just use these as is, if you write Red/system code (which interfaces with readline) it can only be called from compiled Red code, not from the interpreter. This makes experimentation considerably harder, last time I've tried the compiler was super slow.

ALANVF:
Well that's a bit disappointing. Maybe I can keep this as a separate implementation, and possibly rewrite this in Rebol (since it is similar to Red, but has less issues)

wasamasa:
I don't expect REBOL to be much different. Foreign functions are in the paid version, user input returns "q" for nothing and IO, no idea.

ALANVF:
REBOL 3 is open source, but if I don't find a solution, I may have to rewrite this in Red/System

ALANVF:
I'm closing this because I don't think any more progress will be made here (and I'd like to update my fork lol). I've moved the code to [a separate repository](https://github.com/ALANVF/mal-Red) if anyone would like to use it as reference or would like to try adding it to the main repository themselves.

PR #442, opened 2019-08-18
Mal: cosmetic improvements for envs and macros

asarhaddon:


process/guide.md
```diff
@@ -1259,6 +1259,10 @@ simple.
   `env` environment and that function has the `is_macro` attribute set
   to true. Otherwise, it returns false.
 
+  Feel free to adapt this description. For example, it may be more
+  efficient to return the macro itself instead of `true`, or more
+  expressive to fully replace a call with pattern matching.
+
```
kanaka: This is true of lots of parts of the guide. I would be okay with saying something more generic at the top of the "Make-a-Lisp Process". There isn't anything unique about this particularly part. Something along the lines of "Feel free to follow the guide as literally or as loosely as you like. You are here to learn; wandering off the beaten path may be the way you learn best. However, each step builds on the previous steps, so if you are new to Lisp or new to your implementation language then you may want to stick more closely to the guide your first time through to avoid frustration at later steps."

kanaka:
Putting the outer env into a cons cell decreases code clarity and makes it less consistent with most other implementations which have some sort of named field for outer. Unless the efficiency gain was super compelling I don't think it's worth the cost in clarity and comparability of implementations. If having the :outer field contained in the data (i.e. accessible from user code) is particularly bothersome to you, I would consider a pull request that changed envs to be a map with an :outer and :data field where the :data field is the env values contained in an atom. Having :outer in data itself doesn't bother me personally though. I've actually considered several times making env objects first class objects in mal. The main reason I've not gone done that path is that it disrupts the incremental progression because you now need some sort of associative map and atom-like functionality early in step 3 (or make env part of the implementation steps and change it multiple times).

If you make the change to the guide I would be happy to merge the other commits.

asarhaddon:
Your paragraph is inserted into `process/guide.md`.
My intent was to shorten the frequently executed `env-find`. Since c20b346487601ce46683cc80735663684e661951, the key used for the outer environment is a keyword, which cannot interfer with normal keys (strings converted from symbols).
However, after some local tests, implementing environments with cons cells does not seem to perform significatively better so I have removed it.
I have kept the part moving the symbol to string conversion out of the iteration, more for style than for efficiency.

kanaka:
Change looks good.

I triggered self-host testing using the branch: https://travis-ci.org/kanaka/mal/builds/574924697 So far there are failures in awk and bbc-basic in step6. However, doing a little debugging it appears the failure was actually introduced with e6d41de (the load-file/nil change). I thought I had run self-host tests with that change but apparently not.

Once it finishes, if looks pretty clear that e6d41de is the culprit, then I'll merge this and open a ticket to track fixing issues with load file/nil.

kanaka:
The failures all look related to e6d41de. Merged.

Issue #443, opened 2019-08-21
load-file/nil change (e6d41de) breaks self-host for a few implementations


kanaka:
During testing for #442, I discovered that e6d41de broke a few implementations in step6 when self-hosting.

Here is the travis run: https://travis-ci.org/kanaka/mal/builds/574924697

Initial investigation suggests that it is probably reader tokenization issues related to the new newline in the load-file definition.

The following implementations need fixing:

```
- [ ] awk
- [ ] bbc-basic
- [ ] tcl
- [ ] yorick
```

Triggering the problem locally:

```
make DOCKERIZE=1 MAL_IMPL=awk test^mal^step6
```

I have a fix in progress for awk. @bjh21, can you take a look at bbc-basic. @dubek want to take a look at tcl and yorick?

bjh21:
Investigating...

bjh21:
I think this is arguably a bug in the mal implementation, or at least an ambiguity in the guide. The guide in step 6 says that `load-file` should be defined like this:
```
(def! load-file (fn* (f) (eval (read-string (str "(do " (slurp f) "\nnil)")))))
```
but the mal implementation defines it like this:
```
(rep "(def! load-file (fn* (f) (eval (read-string (str \"(do \" (slurp f) \"\nnil)\")))))")
```
so the string that actually gets passed to `rep` is:
```
(def! load-file (fn* (f) (eval (read-string (str "(do " (slurp f) "
nil)")))))
```
with a newline in the middle of  a string. That's permitted by the official regexp for string literals (so there's also a bug in the BBC BASIC implementation), but not what I think the guide calls for.

kanaka:
@bjh21 you're right, the guide is vague. I think having it vague in the guide but having a test that makes sure it works. And I think bare and escaped newlines should both work. Especially since only 5 implementations have issues.

kanaka:
Fixes for awk and bash: https://github.com/kanaka/mal/commit/c9b7b019eb9847648eb53a029f4e6e6d1e1ad585

PR #444, opened 2019-08-21
step 6: Test reading strings containing bare newlines.

bjh21:
This adds a test to detect the underlying problem reported in #443.Depending on how that's resolved, this test might want to be downgraded to optional.


kanaka:
Fixes for awk and bash: https://github.com/kanaka/mal/commit/c9b7b019eb9847648eb53a029f4e6e6d1e1ad585

PR #447, opened 2019-09-17
rust: remove readline.rs

rofrol:


kanaka:
Merged. Thanks for catching that.

PR #448, opened 2019-09-17
rust: rustfmt

rofrol:
Use rustfmt instead of hardcoded vim instructions.


kanaka:
Thanks. I personally prefer being able to vertically line up related things within a line (i.e. short match statement bodies), but if this is the canonical style now then we'll go with it.

PR #449, opened 2019-09-18
Pike implementation

dubek:
Add a Mal implementation in Pike.

According to [its website](https://pike.lysator.liu.se/),

> Pike is a dynamic programming language with a syntax similar to Java and C. It is simple to learn, does not require long compilation passes and has powerful built-in data types allowing simple and really fast data manipulation.

The Dockerfile uses Pike v8.0 from Ubuntu bionic apt repo (8.0.498); I also tested locally with latest (8.0.702) compiled from source.

The implemenation passes all the tests and self-hosting tests. It also supports basic Pike interop with `pike-eval` (see `pike/tests/stepA_mal.mal`).

Performance:

```
$ make perf^pike
Performance test for pike:
Running: env STEP=stepA_mal MAL_IMPL=js ../pike/run ../tests/perf1.mal
Elapsed time: 3 msecs
Running: env STEP=stepA_mal MAL_IMPL=js ../pike/run ../tests/perf2.mal
Elapsed time: 9 msecs
Running: env STEP=stepA_mal MAL_IMPL=js ../pike/run ../tests/perf3.mal
iters over 10 seconds: 3572
```

To compare, running python3 performance tests on the same macine gives: 6 msecs, 23 msecs, 1240 iters.

After finishing the implemenation I discovered that Pike comes with a [small Lisp implemenation](https://github.com/pikelang/Pike/blob/master/lib/modules/Languages.pmod/PLIS.pmod) in its modules library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,7 @@ matrix:
     - {env: IMPL=perl6,                       services: [docker]}
     - {env: IMPL=php,                         services: [docker]}
     - {env: IMPL=picolisp,                    services: [docker]}
+    - {env: IMPL=pike,                        services: [docker]}
     - {env: IMPL=plpgsql NO_SELF_HOST=1,      services: [docker]} # step3 timeout
 #    - {env: IMPL=plsql,                       services: [docker]}
     - {env: IMPL=ps,                          services: [docker]}
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ DOCKERIZE =
 IMPLS = ada ada.2 awk bash basic bbc-basic c chuck clojure coffee common-lisp cpp crystal cs d dart \
 	elisp elixir elm erlang es6 factor fantom forth fsharp go groovy gnu-smalltalk \
 	guile haskell haxe hy io java js julia kotlin livescript logo lua make mal \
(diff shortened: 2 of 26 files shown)


dubek:
I cancelled the build because a Docker image should be built and pushed first.

kanaka:
I built and pushed an image and restarted the build. All green. Merged. Want to tweet it?

That brings you to 11 implementations (12 if you include malc). So what's next on the docket? :-)

dubek:
Pike tweet: https://twitter.com/dubek/status/1174363929738395650

Next is probably Wren.

I started to look at Zig a few months ago (we don't yet have an implementation that starts with Z) but memory management there is more complex, so I need some more free time (which I don't have now) to learn that. I guess its similar to the Rust issues (@kanaka  I believe you mentioned that was one of the surprisingly harder ones).

kanaka:
Yes, the memory ownership model in rust along with trying to do it fairly early in the development of rust (rust error messages are now excellent but it wasn't always so) made it fairly difficult of an implementation. If I were to rate implementation difficulty for the ones I've done personally it would be: GNU Make, Postscript, Bash, Rust, Basic, Haskell. If I were to redo them from scratch, Rust and Haskell would fall out of the top because it wasn't limitations of the language that made them difficult, just the fact that I was picking them up for the first time to do the implementation. The others would remain difficult because of the languages themselves.

PR #455, opened 2019-10-14
Add python.2 implementation

epylar:
This commit adds a new Python 3 implementation of MAL. This implementation passes all non-optional tests, with the exception that the Travis test script doesn't like the period in the directory name. During development, this was called "py3" and it passed Travis tests.

This implementation is different from the main python implementation. First, it heavily uses type annotations making for a code style more similar to Java. Second, it uses the Arpeggio parser library to parse MAL syntax.

This commit also includes unit tests that show roughly the order of things that were interactively tested and debugged during development.


epylar:
Please let me know if you have any comments or suggestions on this.  I learned mypy/py3 type annotation/Arpeggio from scratch while doing this implementation. It was a lot of fun! Thanks for creating this project.  I'm already thinking about maybe doing this again in a language I'd like to learn (maybe Rust or Haskell).

kanaka:
@epylar The dot in the name shouldn't be a problem. It might be the fact that the run script isn't executable. I've build and pushed an image to docker hub so you should now be able to activate and test this in Travis.

BTW, I do notice that the this implementation is slower than the regular python3 implementation. I guess mypy doesn't do any performance optimizations with the typing information yet. I wonder if there are any plans to support mypy style annotations with rpython. Anyways, it's not a blocker, just an observation.

I've briefly reviewed the code and don't see any major issues. I think the remaining things are:
- fix ```./run``` executable bit
- add to ```.travis.yml``` and make sure it passes in Travis.
- add section to ```README.md``` and bump the implementation and mode counts at the top.
- I do final review and merge
- You tweet (if you want to) and I retweet to announce the implementation.

epylar:
@kanaka Cool. I notice disabling asserts with -O makes the code about 21% faster. Also my code calls constructors and accessors quite a bit compared to the Python native version which doesn't use this layer of indirection.

Mypy as far as I am aware today only performs static analysis and does not affect runtime behavior.

I'll do the above steps plus rename my 'tests' to 'unittests.'  I'd like Travis to run through the unit tests also but that could be a future PR.  Another future PR could be changing 'run' to pass in -O to speed things up a bit.

kanaka:
Travis is passing. I think the only thing left before merging is to update the README.

epylar:
@kanaka `README.md` is updated

kanaka:
Merged! Only thing left is the announcement. If you're on twitter and want to tweet it, throw me a link to retweet. Otherwise, I can just send out a tweet myself.

dubek:
I see the Docker image is based on alpine; @kanaka , I thought you'd like all images to be based on some standard base layers (maybe publish a `kanaka/mal-base` image and use that in the `FROM` line for all impls)?

kanaka:
@dubek True, thanks for reminding me. @epylar There are a couple of reasons for that are for my own long term sanity. If images share the same base layer and especially if they share the same first few layers, then this both is faster for me to rebuild multiple images and also takes up substantially less disk space. For the disk space issue alpine is not such a big issue because it's a small base layer to start from. But the speed of rebuild is still affected by the number of different base layers that images require. If all 78 implementations had a different base layer then it would take me a week just to rebuild everything and I would like have over 100GB of images. @epylar if you have a chance, can you update the repo with a Dockerfile that uses the same base layer and initial layers as the python Dockerfile? I'll rebuild it and then restart Travis to make sure it still works.

epylar:
@kanaka Sure thing! One wrinkle though--the existing python Dockerfile is based on vivid, which is 15.04 and not an LTS. When I build it from scratch, I get a bunch of 404 Not Found errors. Perhaps everything could be upgraded to at least xenial (16.04)? That's the closest LTS to 15.04 and it builds fine when substituted into the python Dockerfile.  Also make DOCKERIZE=1 test^python passes with xenial.

epylar:
@kanaka Also, I tweeted here: https://twitter.com/epylar/status/1184507835511525376

kanaka:
@epylar My normal process is that when I have a need to update an existing image, I bump the base layer to whatever the current latest LTS image is. I really should do a pass through to bring the vivid images up to date at some point. Either xenial or bionic is fine (they are the second and third most common base layer).

I retweeted you. And welcome to twitter ;-)

PR #458, opened 2019-10-17
Standardize python.2 dockerfile

epylar:
Match other Dockerfiles to re-use layers.


kanaka:
Merged! I built and pushed an update to the image based on your change.

PR #459, opened 2019-10-17
Fix Travis test script to work with Bash 5

epylar:
Bash 5 doesn't like periods in variable names so this commit performs a substitution to double underscores to avoid them. Fixes #456.

PR #460, opened 2019-10-19
Improve python.2 performance on perf3 benchmark by 77%

epylar:
Performance is improved about 77% on my system
    
Optimizations:
* remove use of `ABC`/`abstractmethod`
* start new environments with copy of `outer`; no need to `find(key)`
* run with asserts off

### DIFF
--- a/python.2/env.py
+++ b/python.2/env.py
@@ -12,8 +12,10 @@ class Env(object):
         binds: Optional[List[MalExpression]] = None,
         exprs: Optional[List[MalExpression]] = None,
     ) -> None:
-        self._data: Dict[str, MalExpression] = {}
-        self._outer = outer
+        if outer:
+            self._data: Dict[str, MalExpression] = outer._data.copy()
+        else:
+            self._data = {}
         if binds is not None and exprs is not None:
             for x in range(0, len(binds)):
                 assert isinstance(binds[x], MalSymbol)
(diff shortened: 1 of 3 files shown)


epylar:
I'm not sure if the `outer` optimization is a good idea (uses more space?).

epylar:
Also unsure why GitHub is showing the #459 merge here even after resetting my origin to upstream and rebasing.

epylar:
OK, I'll stop experimenting on this PR and figure out how to do proper PRs across forks on a dummy repo.

kanaka:
Looks like you have it entangled with some of the bash-5 work you were doing. I was able to fix the epylar:perf-python.2 branch you have pushed like this:

```
git rebase 2fe64cd
```

The hash ```2fe64cd``` is the one for the bash-5 fix that is upstream in master.

epylar:
@kanaka rebase 2fe64cd and then push --force-with-lease seems to have resolved it

epylar:
@kanaka I also noticed running either of the Python implementations in pypy produced a huge speedup for perf3. For the main `python` implementation, it ran about 6.6 times faster.

kanaka:
Merged. From the travis results it does seem that the implementation now has the same or better performance than the regular python3 implement. Yes, pypy is very fast. Even faster is mal implementation in rpython (pypy itself is implemented using rpython). I would be happy to take a PR that added an additional run mode for the python and python.2 implementation that runs with pypy. This would involve adding pypy to the Docker images, a mode variable for python.2 (python__2_MODE with the recent path), and adding a Travis run mode for each.

PR #462, opened 2019-10-23
Add test for (def!)

epylar:
I'm curious to see how many implementations pass this.

### DIFF
--- a/tests/step3_env.mal
+++ b/tests/step3_env.mal
@@ -19,7 +19,7 @@ x
 y
 ;=>8
 (def!)
-;/.*(syntax|Syntax).*
+;=>nil
 
 
 ;; Verifying symbols are case-sensitive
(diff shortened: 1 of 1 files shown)


epylar:
Rust MAL implementation panics; python MAL implementation returns nil.  Clojure (actual Clojure, not MAL-in-Clojure) produces a syntax error.

dubek:
Error handling like this is intentionally not specified for the implementations, in order to keep the "exercise" focused. I assume you'll get a variety of issues if you try `fn*` or `let*` or `macroexpand` without arguments; but handling all those cases would make the code of many implementations even more complex. Maybe we should add it to the end of `process/guide.md` as a possible addition for a production-grade programming language.

epylar:
@dubek makes sense! Yeah, adding all those checks would make language implementations much more difficult.

PR #463, opened 2019-10-24
Update Rust implementation to Rust 2018

epylar:
- Use 'cargo fix --edition' to perform upgrade
- Update Dockerfile to use latest Rust stable: 1.38.0
- Add Cargo.toml as a dependency for Rust steps in Makefile


epylar:
Is this something that would be better done as a second Rust implementation? I've been writing my own Rust implementation but aside from using Rust 2018 and later library versions, it's not much different from the existing one, so I'm thinking just push any obvious improvements without changing the underlying style.

kanaka:
Unless the implementation is different in some notably interesting way, I think sticking with a single implementation and updating it is best.

I noticed that there are bunch of warnings during the build. Updating lazy_static fixes them:

```
diff --git a/rust/Cargo.toml b/rust/Cargo.toml
index 3eb30b0..a262b9b 100644
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -4,8 +4,9 @@ version = "0.1.0"
 authors = ["root"]
 
 [dependencies]
 rustyline = "1.0.0"
-lazy_static = "1.0.1"
+lazy_static = "1.4.0"
 
 regex = "1.0.0"
 itertools = "0.7.4"
```

If you make that change then I'll update the docker image and merge.

Interestingly, I also tried updating the version of rustyline and the stepA test that invokes readline directly broke. Not a blocker since the old version seems to work fine for now.

epylar:
@kanaka done.

epylar:
This should fix the rustyline issue.

kanaka:
Looks good! Merged.

PR #464, opened 2019-10-30
tests: step8: Verify nth doesn't throw on nil elements

dubek:
I had a bug in the Wren implementation of `nth` that slipped all the way to self-hosting.

Adding these tests caught it early in step8 when we introduce `nth`.


kanaka:
Looks good.

PR #465, opened 2019-11-02
Wren implementation

dubek:
A new Mal implementation in [Wren](http://wren.io/), tested with Wren 0.2.0. It includes Mal-Wren interop via the `wren-eval` function; see `tests/stepA_mal.mal` for examples.

The implementation passes all tests and self-hosting tests. On my machine it's faster than python3 and slower than Ruby 2.5.

Wren doesn't have a regexp module, so I ended up tokenizing the input string char-by-char. I found no easy way to add Wren modules with native code at runtime (it's pretty easy to add modules when you compile the whole Wren interpreter).

For `time-ms` I found no workaround and ended-up adding a C call to the system's `gettimeofday` function. The modification to the interpreter is in `wren/wren-add-gettimeofday.patch` (the `Dockerfile` applies this patch before building the interpreter). I'll try to submit it upstream.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -111,6 +111,7 @@ matrix:
     - {env: IMPL=wasm wasm_MODE=node      NO_SELF_HOST_PERF=1, services: [docker]}
     - {env: IMPL=wasm wasm_MODE=warpy     NO_SELF_HOST_PERF=1, services: [docker]}
     - {env: IMPL=wasm wasm_MODE=wace_libc NO_SELF_HOST_PERF=1, services: [docker]}
+    - {env: IMPL=wren,                        services: [docker]}
     - {env: IMPL=yorick,                      services: [docker]}
 
 script:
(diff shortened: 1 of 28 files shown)


kanaka:
wren test is failing but looks like everything is working except the initial build step. I think it just needs a stub Makefile in the directory.

dubek:
I pushed a dummy Makefile and now it passes:  https://travis-ci.org/kanaka/mal/jobs/606545197

dubek:
I also added a `dist` target to the Makefile (create a standalone file).

kanaka:
Awesome. Merged. Just 11 more until 100!

kanaka:
Oh, you might want to mention in the README that full functionality requires the patch to wren. Can you use it with unpatched wren (and it just fails when you try to invoke time-ms)?

dubek:
https://twitter.com/dubek/status/1190733522790694913

I think that's 21 more till 100.

I'll check if pure wren runs OK.

dubek:
Yep, it's fine:

```
 $ ./run
Mal [wren]
user> (+ 1 1)
2
user> (time-ms)
Error: System metaclass does not implement 'gettimeofday'.
user>
```

dubek:
I mentioned the wren patching in `wren/README.md` but not in the main `README.md`.

kanaka:
Oh yeah, 21. I must be tired. And since it works except for time-ms, no need to mention it in the main README.

PR #467, opened 2019-11-08
Swift5 implementation

MontakOleg:
Hello, @kanaka 
Thank you for great tutorial!

This is swift5 implementation.

- [x] make "test^[IMPL_NAME]"
- [x] make MAL_IMPL=[IMPL_NAME] "test^mal"
- [x] ./.travis_test.sh test [IMPL_NAME]

Main differences:
1. swift package manager to build/run steps
2. reader implemented using parser combinators

Dockerfile from swift4


kanaka:
@MontakOleg sorry for the slow reply. Last week I was focusing on completing my PhD dissertation which is now finally done. Yay!

Anyways, you're implementation looks good. I built the docker image, pushed it and restart the Travis build and it passes. I think I only have one thing I would like changed (or at least understand your motivation) about the implementation before it is merged: the run script invoke the swift command on the implementation step rather than just executing the compiled program like other compiled implementations. My preference is that the run script does not require compilation if the implementation language supports compilation. That way, the compiled implementation could be run using the run script even if the swift program is not available.

MontakOleg:
`swift run` makes two things:
1. if necessary compiles code
2. find executable for current platform (.build/macosx-x86_64/debug or .build/linux-x86_64/debug) and run it

So there is no additional compilation if we run:
```
swift build
swift run
```

For clarification I added `--skip-build` option to swift run invocation

So we using `swift run` only to find location of executable.

>That way, the compiled implementation could be run using the run script even if the swift program is not available.

We can avoid using `swift run` by using `swift build --show-bin-path` on build step and saving that path to file. Run script can use that file to figure out where search binary.
Do you think we should go that way?

kanaka:
@MontakOleg The concern I have is not so much with the build, but the dependency on the swift executable. My preference would be for the Makefile to use the swift command to copy the build executable to a well known name and then for the run script to use that directly. Basically, I would like the run file to be usable in a context with only the built executable (i.e. no swift tools). For example, sometimes I use a docker image (which has the swift tools) to build the executables but then run tests outside the image which doesn't have the docker tools. I don't consider this a blocker though so I went ahead and merged it. But would like to see that cleaned up when you have the chance.

Thanks for the updated implementation. If you have a twitter account and would like to "announc" it, post a link to you tweet here and I'll retweet it.

PR #469, opened 2019-11-23
Use a real (if primitive) terminal emulator for tests

bjh21:
One problem I ran into in my BCPL implementation was that the BCPL runtime optimised away certain CR characters from its output, which meant that `runtest.py` thought there was something wrong because CRs that it was expecting weren't there. Initially I solved this by adding extra complexity to the regular expressions in `runtest.py`, but this whole approach felt wrong.

So here's my new approach.  I've written a simple terminal emulator called "olte" that implements an ANSI terminal one line high and infinitely wide. `runtest.py` then looks at what appears on the display (and scrolled off the top) of this terminal to decide if the implementation output the correct thing.  This has the advantage that it doesn't care how the implementation gets its output right: it can muck about with CR, BS, and (supported) escape sequences and as long as the display ends up right, that's acceptable.

This has left `runtest.py` rather cleaner, albeit at the expense of adding a a great lump of additional code. I think it's probably a good idea, but I could well be wrong.

### DIFF
--- a/olte.py
+++ b/olte.py
@@ -1,0 +1,125 @@
+# OLTE: One-Line Terminal Emulator
+
+# OLTE is a simple terminal emulator library that emulates a terminal
+# one line high and infinitely wide.  It's intended for test suites
+# that want to drive programs with command-line editing and the like.
+
+from __future__ import unicode_literals
+
+import re
+import sys
+
+IS_PY_3 = sys.version_info[0] == 3
+
+if not IS_PY_3:
+    chr = unichr
+
+def isprintable(s):
+    if IS_PY_3: return s.isprintable()
+    return 32 <= ord(s) < 127 or 160 <= ord(s)
+
+class OneLineTerminalEmulator(object):
+    # There are three regular expressions here for each kind of magic
+    # character sequence that we recognise.  The first recognises a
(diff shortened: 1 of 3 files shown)


bjh21:
I see that I've got some bugs to look at.  Tomorrow.

dubek:
@bjh21 note that in step 6 we run `run_argv_test.sh` which checks the stdout of simple Mal programs (basically to test `*ARGV*` implementation details). I'm not sure how this will behave if an implementation will output a bunch of BS chars or escape sequences.

bjh21:
`runtest.py` runs the implementation inside a pseudo-terminal (by default), while `run_argv_test.sh` runs it without a psuedo-terminal. Any implementation that emits control characters other than LF should really only do so in an interactive context, so I think it's reasonable for the two test wrappers to have different behaviours.

bjh21:
But actually I do now wonder about an alternative approach: the time when we're most likely to get weird escape sequences is when the implementation is reading an input form, and I don't think `runtest.py` really has any cause to care whether the implementation displays input forms correctly.  So it could just ignore everything between the prompt and the next newline, ignore all CRs, and process what's left. I'll try some experiments.

bjh21:
This is superseded by #510.

PR #470, opened 2019-11-29
Fix asterisk replace in string in Bash.

chr15m:
I think this is the fix for https://github.com/chr15m/flk/issues/1
Tests passing on my local, will see if Travis complains.


chr15m:
Looks like the bash test has passed.

Just to provide a bit more info: I ran into this when doing `(sh* "grep -oE '.*'")` and having that return the wrong results. When I dug in I realised that this was actually executing `grep -oE '.__STAR__'`. So this patch stops the `string` type in the bash implementation from replacing all asterisks with `__STAR__`. Hopefully that makes sense.

chr15m:
Pushed another commit in here which fixes an issue where asterisks which are single quoted inside an `(sh*...)` invocation are parsed into a wildcard by bash when they shouldn't be. Added a test which will fail without 08340fb and pass with it.

chr15m:
Hm these tests are passing on my local. Digging in to figure out why.

chr15m:
I've worked out a better fix for this which I'll PR.

PR #471, opened 2019-11-30
Fix Bash implementation interop carriage returns.

chr15m:
Before this fix the 'sh*' function was joining the lines of the output of whatever command was run in a way that resulted in '\n' instead of a literal carriage return.

Without this fix:

```
$ ./mal
Mal [bash]
user> (println (sh* "head mal"))
#!/usr/bin/env bash\n#\n# mal (Make a Lisp) object types\n#\n\nif [ -z "${__mal_types_included__}" ]; then\n__mal_types_included=true\n\ndeclare -A ANON\n
nil
```

After this fix (carriage returns correctly inserted by `sh*`):

```
$ ./mal
Mal [bash]
user> (println (sh* "head mal"))
#!/usr/bin/env bash
#
# mal (Make a Lisp) object types
#

if [ -z "${__mal_types_included__}" ]; then
__mal_types_included=true

declare -A ANON


nil
```

### DIFF
--- a/bash/stepA_mal.sh
+++ b/bash/stepA_mal.sh
@@ -166,7 +166,7 @@ EVAL () {
               local output=""
               local line=""
               while read line; do
-                  output="${output}${line}\n"
+                  output="${output}${line}"$'\n'
               done < <(eval ${ANON["${r}"]})
               _string "${output%\\n}"
               return ;;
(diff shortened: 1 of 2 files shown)


chr15m:
Added more tests, fixed the stripping of the final `\n`, and allow `sh*` to capture output which has no `\n` at all within it.

PR #473, opened 2019-12-02
Fix bash interop asterisk quoting

chr15m:
This supercedes kanaka#470 and includes kanaka#471.
This fixes some edge cases when passing an asterisk to an sh* interop call.
Before this fix mal would pass the string with any wildcard character converted to __STAR__ when passing to the subcommand.
For example `(sh* "ls *")` would become `ls __STAR__` when executed.
See the tests for examples of where this would fail to do what the user expects.
See also: chr15m/flk#1


kanaka:
@chr15m thanks for tracking this down and for this easy to review commits and issue comments! Merged.

PR #475, opened 2019-12-04
npm update should be npm install

gautamkrishnar:
`npm update` is used to update the deps use `npm install` instead


kanaka:
Thanks. Merged.

PR #476, opened 2019-12-07
Fix python.2 Env.find implementation.

rectalogic:
It should search the outer Env

Currently unit tests fail:
```
$ python -m unittest discover -s tests
.....F...............................4
..........................................................."exc is:" "'abc' not found"
.............
======================================================================
FAIL: test_env_find_outer (test_step3.TestStep3)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/aw/Projects/cureatr/dev/cureatr/experiments/lispy/mal/python.2/tests/test_step3.py", line 23, in test_env_find_outer
    self.assertTrue(result is outer)
AssertionError: False is not true

----------------------------------------------------------------------
Ran 109 tests in 0.545s

FAILED (failures=1)
```

### DIFF
--- a/python.2/env.py
+++ b/python.2/env.py
@@ -12,10 +12,8 @@ class Env(object):
         binds: Optional[List[MalExpression]] = None,
         exprs: Optional[List[MalExpression]] = None,
     ) -> None:
-        if outer:
-            self._data: Dict[str, MalExpression] = outer._data.copy()
-        else:
-            self._data = {}
+        self._outer = outer
+        self._data: Dict[str, MalExpression] = {}
         if binds is not None and exprs is not None:
             for x in range(0, len(binds)):
                 assert isinstance(binds[x], MalSymbol)
@@ -32,6 +30,8 @@ class Env(object):
     def find(self, key: MalExpression) -> Optional["Env"]:
         if str(key) in self._data:
             return self
+        if self._outer is not None:
+            return self._outer.find(key)
         return None
 
     def get(self, key: MalExpression) -> MalExpression:
@@ -49,4 +49,4 @@ class Env(object):
         for d in self._data:
             env_str += str(d) + ": " + str(self._data[d]) + ", "
         env_str += "}"
-        return "environment: (data: " + env_str + ")"
+        return f"environment: (data: {env_str} outer: {repr(self._outer) if self._outer is not None else 'None'})"
--- a/tests/step6_file.mal
+++ b/tests/step6_file.mal
@@ -126,6 +126,11 @@
 *ARGV*
 ;=>()
 
+;;
+;; Testing that eval sets aa in root scope, and that it is found in nested scope
+(let* (b 12) (do (eval (read-string "(def! aa 7)")) aa ))
+;=>7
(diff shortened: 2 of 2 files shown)


kanaka:
@rectalogic These are implementation specific tests and not the main mal tests. I tried running them myself and got an error both with and without the changes. The main mal tests pass with and without the change.

@epylar can you review this change provide feedback?

epylar:
@rectalogic @kanaka I'm inclined to either include the unit tests in the 'build' or simply remove them. I'll take a look at this.

epylar:
@rectalogic It should be sufficient just to remove `test_env_find_outer` from `test_step3.py` -- this is no longer a valid test (including all the outer keys is a performance enhancement and doesn't seem to affect mal functionality).  A separate PR could run add 'run unit tests' to the build.

Unit tests will currently only pass if run from the `python.2/tests` directory (due to relative paths being used).

rectalogic:
I don't think it's correct to copy the outer keys instead of recursively calling `find` on the outer scope.  e.g. this fails but works with my patch.

Fails on python.2 master:
```lisp
user> (let* (b 12) (do (eval '(def! a 7)) a))
'a' not found
```
Works on python.2-env branch:
```lisp
user> (let* (b 12) (do (eval '(def! a 7)) a))
7
```

epylar:
@rectalogic Good catch.. I wondered if this would change any observable mal behavior but couldn't find anything. I didn't think of eval. Maybe you could introduce this as a mal test case for eval? I think most implementations should pass.  

The unit tests do pass for me as long as I run them from `python.2/tests`, on `0c31a28`.

@kanaka Do you think the above example `(let* (b 12) (do (eval '(def! a 7)) a ))` => `7` would be a good test case to add to mal?

kanaka:
@epylar yes, that seems like a reasonable test to add to step6. It will be interesting to see how many other implementations break.

rectalogic:
I added a test, all implementations passed running `MAKE=make make DOCKERIZE=1 "test^step6"` except plsql and plpgsql both hung starting the docker container. They attempt to sudo but the tests run `docker -u` as the local user who does not exist in the container so sudo fails.

```
Status: Downloaded newer image for kanaka/mal-test-plsql:latest
Starting Oracle XE
sudo: unknown uid 2003: who are you?
```

kanaka:
@rectalogic oh, I wasn't very clear. I meant it will be interesting to see what Travis reports. Everything passed here with that test: https://travis-ci.org/kanaka/mal/builds/623853075. You probably have several dozen gigabytes of docker images on your system now. Sorry I wasn't more clear.

@epylar I think I understand from your comments that you're okay with this change so I have merged it. If I misunderstood and it's not quite what you want then send me another PR and I'll merge it.

kanaka:
@rectalogic the plsql and plpgsql implementations are a bit sensitive to the outside environment user configuration due to the fact that the implementations are built on relation database engines. I don't know of a cleaner way to do it (open to suggestions) but the current setup does work with my home user ID layout and the Travis user ID layout.

PR #480, opened 2019-12-14
swift5: get rid of swift dependency to run tests

MontakOleg:
Get rid of swift dependency to run tests as discussed in #467


kanaka:
Thanks! Merged.

PR #481, opened 2019-12-17
js, es6: meta-data on atoms

dubek:
Support the (optional, soft-fail) meta-data on atoms (added to the guide).

Before the fix:

```
Mal [javascript]
user> (meta (with-meta (atom 7) {"a" 1}))
Error: clone of non-collection: atom
    at Object._clone (/home/dubek/world/mal/js/types.js:77:15)
    at Function.with_meta (/home/dubek/world/mal/js/core.js:153:25)
    at _EVAL (/home/dubek/world/mal/js/stepA_mal.js:143:22)
    at EVAL (/home/dubek/world/mal/js/stepA_mal.js:151:18)
    at /home/dubek/world/mal/js/stepA_mal.js:55:45
    at Array.map (<anonymous>)
    at eval_ast (/home/dubek/world/mal/js/stepA_mal.js:55:20)
    at _EVAL (/home/dubek/world/mal/js/stepA_mal.js:138:18)
    at EVAL (/home/dubek/world/mal/js/stepA_mal.js:151:18)
    at rep (/home/dubek/world/mal/js/stepA_mal.js:162:40)
```

After the fix:

```
Mal [javascript]
user> (meta (with-meta (atom 7) {"a" 1}))
{"a" 1}
```

(fixed similarly for the es6 implementation.)

### DIFF
--- a/es6/types.mjs
+++ b/es6/types.mjs
@@ -28,6 +28,8 @@ export function _clone(obj, new_meta) {
     } else if (obj instanceof Function) {
         let f = (...a) => obj.apply(f, a)  // new function instance
         new_obj = Object.assign(f, obj)    // copy original properties
+    } else if (obj instanceof Atom) {
+        new_obj = new Atom(obj.val)
     } else {
         throw Error('Unsupported type for clone')
     }
--- a/js/core.js
+++ b/js/core.js
@@ -156,9 +156,9 @@ function with_meta(obj, m) {
 }
 
 function meta(obj) {
-    // TODO: support symbols and atoms
     if ((!types._sequential_Q(obj)) &&
         (!(types._hash_map_Q(obj))) &&
+        (!(types._atom_Q(obj))) &&
         (!(types._function_Q(obj)))) {
         throw new Error("attempt to get metadata from: " + types._obj_type(obj));
     }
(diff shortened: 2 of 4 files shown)


kanaka:
Metadata isn't supported on Clojure atoms. That doesn't mean we can't support it in mal but is there a use-case that motivates it? There is also a question of what the behavior should be. Most objects are cloned when meta-data is added. It's a little weird to clone an atom because then you have two references that can diverge (i.e. they are different atoms but start out pointing to the same underlying object). But mutating an atom also doesn't seem like the right behavior either (i.e. `with-meta` would be a pseudo form of reset! just for the metadata part).

dubek:
not in front of PC now; I wonder why that final test was added in stepA.

kanaka:
I added it commented out in b079f51028571bc603b8d43761c29ff56273bffc and without thinking about it. I think I just added metadata tests for all non-scalar types. It got uncommented a few months ago as a soft test in 77057e69200e1bdd1e6e3704787906324c031420 and I didn't think about it carefully at the time. Also the comment about it not being support by Clojure was dropped which made it look more official. Anyways, you're proposal was the first time I really though about it in any depth. It should probably just be dropped as a test and idea. What do you think?

dubek:
yes, remove this test completely and close this PR without merging.

kanaka:
Resolved in b75dc79.

PR #482, opened 2019-12-17
es6, js, python, ruby: defmacro! doesn't mutate original function

dubek:
Modify `defmacro!` to clone the given function before setting the `ismacro` flag.

This fixes a soft-fail test in stepA.

(there are other implementations which still don't support it.)


kanaka:
Merged. Thanks!

PR #486, opened 2020-01-05
Citron implementation

alimpfard:
wrote this for fun, figured I'd make a PR of it :man_shrugging: 

tested against trunk (https://github.com/alimpfard/citron). sometimes `step4` crashes for no apparent reason (only on native?)

readline-like capability is provided through linenoise (https://github.com/antirez/linenoise)


kanaka:
I pushed a build of the docker image to docker hub (kanaka/mal-test-citron) so you should be able to test it in Travis now. Running step4 tests crashes every time for me. Interestingly, it mostly but not entirely consistently when running the following test:

```
( (fn* [f x] (f x)) (fn* [a] (+ 1 a)) 7)
```

It's not that test along because firing up the REPL and running just that test does not cause a failure. Also, where exactly if fails seems to have more to do with whatever I just did on the host machine. This smells like a GC issue. My suspicion is that there is some weird interaction with BoehmGC. Just a hunch though. The program just seems to crash with no output. Might need to capture a crash dump to be able to debug this.

alimpfard:
Figures, really.
the GC on that pet language of mine has quite a few problems...

_one day_ I'll get around to fixing it.

alimpfard:
Your hunch seems to be right, as disabling the automatic collection _does_ cause the test to pass. consistently.

PR #487, opened 2020-01-09
Add test for interaction of let* local atoms and captures

alimpfard:
Test the interaction between (possibly shadowing) atoms from a let* scope

this is used in load-file-once, would've been nice to know about the fault in advance


alimpfard:
Perhaps this is more suited to be an issue, I'll open an issue instead.

PR #488, opened 2020-01-10
Add Jq implementation

alimpfard:
This is a [jq](https://github.com/stedolan/jq) implementation, _can_ self-host but reading files seems to take forever (I'd love to work on that further)

Since jq has extremely limited io (in the form of reading from stdin and writing json to stderr), I had to make a python wrapper that emulated those (hope this doesn't count as too much cheating~)


jq/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM archlinux/base
```
dubek: So far we use a few base images to save space (and shorten pull time) when testing several implementations. @kanaka - do we still keep this practice?

I'm sure there's a `jq` package in Ubuntu as well.

Makefile
```diff
@@ -94,7 +94,7 @@ IMPLS = ada ada.2 awk bash basic bbc-basic c chuck clojure coffee common-lisp cp
 	guile haskell haxe hy io java js julia kotlin livescript logo lua make mal \
 	matlab miniMAL nasm nim objc objpascal ocaml perl perl6 php picolisp pike plpgsql \
 	plsql powershell ps python python.2 r racket rexx rpython ruby rust scala scheme skew \
-	swift swift3 swift4 swift5 tcl ts vala vb vhdl vimscript wasm wren yorick zig
+	swift swift3 swift4 swift5 tcl ts vala vb vhdl vimscript wasm wren yorick zig jq
```
dubek: Can you please put `jq` in the correct place according to alphabetic order?

README.md
```diff
@@ -44,6 +44,7 @@
 | [Hy](#hy) | [Joel Martin](https://github.com/kanaka)  |
 | [Io](#io) | [Dov Murik](https://github.com/dubek) |
 | [Java](#java-17) | [Joel Martin](https://github.com/kanaka)  |
+| [Jq](#jq) | [Ali MohammadPur](https://github.com/alimpfard) |
```
dubek: I think Jq should be just before Julia according to alphabetic order.
Also please update the total numbers at the top of the table.

jq/Dockerfile
alimpfard: That's a fair point.

Though the vivid LTS has a lot of outdated packages (the jq in ubuntu's repositories is ver 1.5, I think that one tends to segfault for some reason)

I'll try to use the same base image, let's see how that goes

jq/step4_if_fn_do.jq
```diff
@@ -0,0 +1,220 @@
...
+
+def read_line:
+    . as $in
+    | label $top
+    | _readline;
+
+def READ:
+    read_str | read_form | .value;
+
+# def eval_ast(env):
```
wasamasa: Why that big block of commented out code?

jq/Dockerfile
dubek: vivid is pretty old; use `FROM ubuntu:18.04` (bionic - the latest LTS release) as newer implementations do.

jq/step4_if_fn_do.jq
alimpfard: Ah, that reminds me to clean up.

jq/Dockerfile
alimpfard: Ended up using ubuntu:bionic (zig also uses this, so theoretically, no loss?)

dubek: Looks OK. I guess you could use curl instead of installing wget, but not sure it matters. Anyway need to wait for @kanaka to build this image and push it to docker-hub so the tests in Travis can run; we'll hear his opinion.

kanaka:
Sorry for the slow reply. I was in the Caribbean last week and not checking email at all. Seems like every time I go on vacation there are multiple mal implementations completed and waiting on me. :-)

Great work on the implementation. Languages without full I/O and/or FFI support are always an interesting challenge. And yes, for that sort of language an I/O driver/harness is acceptable. This was necessary for the SQL implementations.

I have pushed a build of the docker image and restarted the most recent Travis build: https://travis-ci.org/kanaka/mal/jobs/635939712 It currently fails the overall test but the only reason is because the travis test is expecting to be able to do a "make" inside the directory and the implementation is missing a stub Makefile. I haven't done an in depth review of the code but should be able to do so in the next couple of days.

alimpfard:
I'll address that makefile issue.

kanaka:
@alimpfard the overall implementation looks good. There are a few structural changes that need to be made before merging. A key make-a-lisp goal is that each implementation is serves as an incremental reference (learning tool) for how the different features of a mal interpreter. So if I do a diff of two adjacent steps I can see what was changed to implement that step. E.g.

```
diff -urp python/step1_read_print.py python/step3_eval.py
```

While the `jq/interp.jq` file makes a lot of sense for reducing code duplication and modularity, it's actually in conflict with the incremental comparability goal. step2 includes `interp.jq` but only makes use of a small portion of that code. Also, step2 shouldn't transitively depend on env and core, and step3 shouldn't depend on core. Although duplication between the steps feels gross, I don't consider that a problem here. The basic principle is that if a function changes in multiple steps, it should be in the step files themselves. If the code is added once, and not changed after that, it can go in a library/included file. This is violated slightly with the `env` module which is added in step3 and modified in step4 to add bindings support (although depending on how you implement env the bindings might be separate from the default env).

Can you please refactor interp so that the step based changes are in the step files (with minimal subset needed for each step)? Other than that the code looks good. Thanks for working on this. I looked at jq a while back and it looked pretty challenging so I'm glad somebody tackled it.

alimpfard:
I'll work on it.
Expect to hear about it in a day or so if everything goes smoothly

alimpfard:
@kanaka am I to assume that keeping `env.jq`, `interp.jq` and `core.jq` as-is for the later steps (i.e. 4,5,...) is fine?

If so, that seems to have went faster than I expected...

kanaka:
Yes, atoms should be refactored out of 4 and 5.

kanaka:
Awesome! I went ahead and merged it. Do you have a twitter account to announce it. If so post a link here and I'll retweet.

kanaka:
@alimpfard Oh, while this is still fresh on your mind, can you add your voice to https://github.com/stedolan/jq/issues/925 explaining what functionality from jq would allow this to be done directly without the need for `rts.py`? I suspect a concrete example of the workarounds necessary for this might spur @stedolan to implement more complete I/O support.

alimpfard:
Here's a tweet:
https://twitter.com/the_semicolon_/status/1216858336236314624?s=20

(not exactly a decent account, but oh well)

kanaka:
@alimpfard thanks for weighing in on: https://github.com/stedolan/jq/issues/925. If you have the time, it might be interesting to create a version of the jq implementation that uses @nicowilliams' dlopen branch (https://github.com/stedolan/jq/pull/1843). It would only test a subset of the new I/O features, but it would be real world testing (in some sense of "real world") that might be helpful as feedback. Also, then it would be ready to update when/if the dlopen branch lands. If you have time and are motivated to do so.

alimpfard:
I was planning to look at some other language when my next batch of free time hit (xpl0 seemed interesting...) but I guess I could fit this 'new' version of jq into my weekends until then

In any case, let's see if I can build that branch to begin with

kanaka:
Understood. Please don't feel any pressure from me. Also, xpl0 would be cool. It would hit one of the three letters that we are missing still: "q", "u", "x". ;-)

alimpfard:
I know, I picked xpl0 since there was a bix X-shaped hole in the list of languages

no pressure, you've only managed to make me addicted to implementing MAL in weird languages

PR #493, opened 2020-01-15
Add test for atoms "leaking" out of let* binds

### LINKED ISSUE #490: Tests for step6 neglect the requirement that atoms are not to be referenced only by name
local atoms to `let*` are to be 'leaked' if they are closed over.
This might seem like an obvious requirement, but some languages can not store things by reference

```
(def! g (let* (atm (atom 0)) (fn* () (deref atm))))
(def! atm (atom 1))
(g)
;=>0
```

alimpfard:
(hopefully) closes #490

### DIFF
--- a/tests/step6_file.mal
+++ b/tests/step6_file.mal
@@ -93,6 +93,12 @@
 (f)
 ;=>9
 
+;; Testing whether closures can retain atoms
+(def! g (let* (atm (atom 0)) (fn* () (deref atm))))
+(def! atm (atom 1))
+(g)
+;=>0
+
 ;>>> deferrable=True
 ;;
 ;; -------- Deferrable Functionality --------


kanaka:
So ... the jq implementation is the only one that fails the test. :-)

alimpfard:
And _that_, is called irony.

Weird though, if it fails that test, how does load-file-once work at all?

kanaka:
It looks like it is getting the value from the prior definition of `atm` rather than the closed over one and that if there is no current definition it works (stepA):

```clojure
user> (def! g (let* (atm (atom 0)) (fn* () (deref atm))))
#<function :anon>
user> (g)
0
user> (def! atm (atom 17))
(atom 17)
user> (def! g (let* (atm (atom 0)) (fn* () (deref atm))))
#<function :anon>
user> (g)
17
```

I also notice that the behavior of step6 (through 9) is different from stepA:

```clojure
user> (def! g (let* (atm (atom 0)) (fn* () (deref atm))))
#<function :anon>
user> (g)
JqMAL Exception :: 'atm' not found
user> (def! atm (atom 17))
(atom 17)
user> (def! g (let* (atm (atom 0)) (fn* () (deref atm))))
#<function :anon>
user> (g)
```

alimpfard:
I'll have to take a look at that...I don't remember changing anything related to atoms in stepA...

alimpfard:
It could be that the locally bound atoms are being detected as free references, which could be why it throws a "not found" exception up until stepA, as that one is not removing the references to free variables from the closure environment.

This could either be trivial, or a mess.

kanaka:
"This could either be trivial, or a mess." Yep, in my experience this sort of bug has no middle ground.

kanaka:
With #494, this now passes for my locally so I'm merging it and trusting that nothing additional will break in Travis :crossed_fingers:

kanaka:
Looks good: https://travis-ci.org/kanaka/mal/jobs/637988326

Also, the good news is that jq isn't the slowest implementation: https://kanaka.github.io/mal/graph/?ckey=syntax&xkey=perf3&ykey=perf3&skey=sloc&xlog=true&ylog=true Looks like it's ahead of 9 languages for the perf3 micro-benchmark :tada:

alimpfard:
Speaking of speed, I do wonder how the status quo would change once `eval` is added, since I could theoretically JIT to Jq itself...

*closed as too ambitious* tee-hee

kanaka:
@alimpfard heh, cool idea.

PR #494, opened 2020-01-16
[jq impl] fix atoms not leaking out of let* (related to pr #493)

alimpfard:
Legacy code, just sitting there...


alimpfard:
yay that broke a perf test

PR #497, opened 2020-02-25
XSLT impl

alimpfard:
XSLT 3 impl

Still incomplete, but hoping someone crazy enough decides to take a look at it and hopefully figure out issues. (and also testing the dockerfile :P )

performance is absolute crap (env is serialised and deserialised between every EVAL call), and something unknown is broken.
runs steps 0 thru 2, but blows up after those.

passes all tests except one, and also runs the perf tests (perf3: 2 iters/10 sec)

in advance, I'm sorry, for anyone that decides to help or review.


dubek:
Impressive feat @alimpfard . I hope I'll have some time to review this (partially, at least) in the next days.

alimpfard:
@dubek Looking forward to having new eyes on the code, mayhaps you can see something I'm not seeing.

_please exercise caution when handling fragile xml_ (teehee)

kanaka:
I built and pushed a image so that should help with testing. I also restarted the Travis test: https://travis-ci.org/kanaka/mal/jobs/655052741

kanaka:
I think something was missed from the Dockerfile. I did a little bit of guessing and came up with the following change:

```patch
index 367cdc2..8715785 100644
--- a/impls/xslt/Dockerfile
+++ b/impls/xslt/Dockerfile
@@ -29,5 +29,6 @@ RUN curl https://repo1.maven.org/maven2/net/sf/saxon/Saxon-HE/9.9.1-6/Saxon-HE-9
 
 RUN mv Saxon-HE-9.9.1-6.jar /bin/Saxon-HE-9.9.1-6.jar
 
-RUN echo -n "#!/bin/sh\njava -cp /bin/Saxon-HE-9.9.1-6.jar net.sf.saxon.Transform \"\$@\""
+RUN echo -n "#!/bin/sh\njava -cp /bin/Saxon-HE-9.9.1-6.jar net.sf.saxon.Transform \"\$@\"" > /usr/bin/saxon && chmod +x /usr/bin/saxon
+
 
```

I pushed an image with that change although you should make it to your PR too. I also restarted the Travis build link above.

alimpfard:
well...the results of this test say: back to the drawing board.

I suppose I cannot get away with (de)serialising the environment every step afterall.
will patch the Dockerfile too, thanks for catching that.

alimpfard:
oh dear...I pushed my graalvm experiment by mistake

alimpfard:
I hope I manage to find some extra time in the coming week to figure out how to serialise the environment better (14 seconds of the 20 total are spent serialising the environment when running perf3).

I am not even sure why it takes so long to do this serialisation tbh

alimpfard:
Unfortunately I don't think we can test self-hosting here, Saxon assumes the stylesheet is in an infinite loop after some 100 tests are run...

not to mention the fact that bootstrapping simply takes too much time (stepA timeouts here)

alimpfard:
stepA's repl launches in around 100 seconds, and Saxon aborts if we throw a lot of tests at it...other than those issues, everything seems to be in working order (at least on my machine).

The code is not going to undergo any more big changes, If anyone was considering reviewing (@dubek?).

wasamasa:
I've done a quick code review and pointed out some style things that have been fixed.

@kanaka @dubek Anything from your side preventing a merge? Performance seems good enough now to not time out the build.

wasamasa:
Looks like something broke here:

```
TEST: '(= true true)' -> ['',true]
Exception: TypeError('expected string or bytes-like object')
Output before exception:
(= true true)
Makefile:409: recipe for target 'test^xslt^step4' failed
make: *** [test^xslt^step4] Error 1
```

alimpfard:
Oddly enough the test passes on my machine :thinking:

wasamasa:
Could it be #510?

alimpfard:
I could try appending an extra carriage return to see if it works :P

but let me pull the docker image and test on that, maybe my python version is behaving differently somehow

kanaka:
@wasamasa thanks for doing the review.

I re-started the Travis build and it passed. Then I ran the test two more times in a row just to make sure it's not temperamentalness in the XSLT implementation itself. It failed again on the third run. This time it was the final equality test in step4 that failed. I do notice when I run it locally that the memory grows over time. It especially balloons up in the middle of the step4 tests with function closures and recursion. Also, it seems to slow down as more tests are executed. Simple equality comparisons later in the file are much slower than the ones earlier in the file. I don't see random failures running the tests locally, but I wonder if the memory usage/growth is causing it to get throttled occasionally on Travis which times out the test case.

Any low-hanging fruit that could reduce memory consumption or perhaps a way to force GC more often? The memory usage itself isn't a blocker for merging (although it would be good to not have obvious leakage) but if this causes Travis to fail every few test runs that will be annoying. So I would want to get the reliability on Travis increased a bit before merging.

Two minor additional requests:
- First, can you move the xslt Travis task down into alphabetical order (I agree it's useful to have it at the top when you are doing repeated testing)?
- Second, a question: will you (@alimpfard) be available to fix issues in the future or update with changes that effect multiple implementations? I don't normally ask this because with most implementations I (or somebody else on the team) can understand the implementation well enough to make small changes/fixes when necessary. This one is a bit more unusual than normal, so I want to make sure it won't rot over time :-)

alimpfard:
> will you (@alimpfard) be available to fix issues in the future or update with changes that effect multiple implementations?

I'll take a look at the second one, but I have a strong feeling that it'd be locked up behind EE/PE.
So far as low-hanging fruit, the whole json environment architecture is a _big_ memory eater, that also takes up around 50% of the runtime. Last time I tried to do this with XML, I failed miserably.

Other than that, I don't think anything would have a significant enough memory footprint.

Sidenote, disabling the profiler might speed up the whole process a little bit. I left it on to get stats, I planned to turn it off once everything was ok anyway


> will you be available to fix issues in the future or update with changes that effect multiple implementations?

I'm always available here, unless I'm dead or something lol.

alimpfard:
We _cannot_ force a GC, but we can ask java to think long and hard to not go over 2GB.
I'm afraid that will require a Docker image rebuild on your part.

kanaka:
I rebuilt the image with the 2G limit on the JVM. Unfortunately, the temperamentalness persists. But it doesn't seem to be related to memory usage. I've re-run the test 8 times on Travis (using the 2G limited versions) and I've gotten 3 failures:

```
            step1:
                TEST: '1;!' -> ['',1]
                Exception: TypeError('expected string or bytes-like object')
                Output before exception:
                1;!

            step1:
                TEST: '({})' -> ['',({})]
                Exception: TypeError('expected string or bytes-like object')
                Output before exception:
                ({})

            step9:
                TEST: '(count (keys hm3))' -> ['',2]
                Exception: TypeError('expected string or bytes-like object')
                Output before exception:
                (count (keys hm3))
```

I think before each of those failures there is a pretty long pause. So I think this may be a timeout issue (although it's pretty clear that the output is there at least for the 2 step 1 failures.

Unfortunately, I can't reproduce this locally so there is something unique about the Travis platform. The on disk bits aren't different because the whole test runs within a container. The fact that the output appears to be there in two of the cases is really weird. Maybe it's missing a newline and not matching? Or maybe there is some sort of buffering going on that prevents matching? Not sure, but maybe instrumenting runtest.py to print out more diagnostics on failure could be helpful: like `repr()` around the output and perhaps timing info for how long the individual test actually took, etc.

But the really high failure rate does mean that this needs to be tracked down and fixed/mitigated before merging so that the Travis testing remains useful and not noisy.

alimpfard:
IIRC step1 just echoed stuff back right? the output is probably not there.
I wonder if it's at all related to the serving thread in my harness?
does travis allow concurrent processes, or is it just one vcpu serving things...?

If it's the later, it could be the case that runtest.py gets a timeslice sooner than the harness thread and then just blocks until it times out...

alimpfard:
If that's the case though, why doesn't any other harness that does this sort of thing fail the same way?

alimpfard:
tried passing `--cpus=0.8` to docker trying to emulate a single vcpu, didn't work out.
I'll try messing with runtest.py then.

kanaka:
Yes, step1 is basically just reading/parsing and printing. So it does seem like something in the harness combined with something about timing/pauses on Travis seems like the most likely culprit at this point.

alimpfard:
Hmmm...I'm not sure how to proceed with this, I could rewrite the harness and try using select, but I don't think that would help much.
maybe I could make it less race-y by adding some sleeps around? that could maybe work...

alimpfard:
I'm gonna drop all the jobs in travis' list temporarily, I imagine cancelling them is a pain anyway

alimpfard:
Travis pass counter: 1

alimpfard:
I think that's it.
if this build passes, I'm gonna call it ok.

kanaka:
Awesome! I ran it 8 more times myself on Travis and it passed every time. Also, it seems to be a fair bit faster. The overall Travis test takes a bit over 3 minutes and used to take about 5 minutes. There is still the significant slowdown the happens with closures/recursion but everything else seems quite a bit more snappy.

I think the only thing left before I merge is to add instructions to the README and bump the counts.

alimpfard:
Yeah, I should've learned not to do busy waits from the jq impl, but...yeah.
I'll patch the README :)

kanaka:
Okay, understood about the performance.

One final thing I just noticed before merging: it looks like the debug in `runtest.py` accidentally is part of the PR. The extra logs and timing could stay in if they were behind a verbose flag but the `%r` printing should be removed. Can you fix that and then I'll merge?

alimpfard:
Should I just drop the commit?
unless we have any particular reason to keep the extra logs?

alimpfard:
I dropped that commit, if runtest debugging is needed, it can be done locally, I guess.

kanaka:
Yeah, dropping the commit is fine. The timing info is definitely generally useful on occasion but it doesn't need to be addressed now in order to merge the implementation.

alimpfard:
:tada:

kanaka:
Merged! Thanks for following up and doing the tedious work to get this to the point where it can be merged. Want to tweet for this?

kanaka:
With XSLT taking the "x" slot, there are now only two starting letters available: "q" and "u". Care to try your hand at Q or Unicon? ;-)

alimpfard:
Never heard of Q, but Unicon reads like gibberish to me lol.
I can certainly take a look!

alimpfard:
Q as in https://code.kx.com/q/ ?

alimpfard:
> Want to tweet for this?

Sure, one sec.

wasamasa:
You may find https://depp.brause.cc/mal-candidates/unicon/ useful for making sense of Unicon.

alimpfard:
Here's the tweet: https://twitter.com/the_semicolon_/status/1267545685437661185?s=20

@wasamasa I'll take a look, thanks.

alimpfard:
@wasamasa the more I look at it the less sense it makes lol
I'll try messing with it sometime in the near future, who knows, it might be very intuitive for all I know

kanaka:
Yes, https://code.kx.com/q/ is the one I was thinking of. Although if you know another language starting with "q" that you prefer, I'm all for that. There just aren't that many :-)

alimpfard:
There was another q which was apparently renamed to "Pure".

Alternative: MAL on quantum computers because why not, q# lol.

PR #499, opened 2020-02-27
update from mal

fredrik-jansson-se:


kanaka:
@fredrik-jansson-se I'm going to close this PR. Apologies for not giving feedback sooner. If you're still interested in this, can you clean it up and re-submit? I'm willing to accept alternate implementations for the same language. However, it should have some unique or interesting difference from an existing implementation. And the naming should be `impl.2` for the second implementation of `impl`.

PR #500, opened 2020-03-02
Fix with-meta and meta description in guide

xixixao:
As is this was confusing, but from tests it's clear that all the types should support meta/with-meta.


dubek:
Looks good; thanks for your contribution @xixixao .

PR #501, opened 2020-03-02
[RFC] Add Hack implementation

xixixao:
Hi, I've made a type-based implementation for [Hack](https://docs.hhvm.com/hack/).

It should be idiomatic Hack.

I haven't created the Docker image, as that's not a fun step, so I'd like to get a go-ahead on this PR before I invest my time in that.

I also plan to add a non-type-based implementation, which would be less educational but more performant (it wouldn't wrap Hack values in objects, it would likely be close to the PHP implementation).

Also I have tried to keep a clear history of commits, so that someone could follow step by step (not just via files), but I can squash them if that's preferred.


impls/hack/bin/step0_repl.hack
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+
+function rep(string $input): string {
+  return print_out(evaluate(read($input)));
+}
+
+// READ
+function read(string $input): string {
```
xixixao: I didn't capitalize these before the Hack linter would complain about the names.

wasamasa: That's fine, the reason these are typed in allcaps is because `read`, `eval` and `print` tend to be already used function names in programming languages and avoids the conflict between the built-in name and our own.

xixixao:
Heya, let me know if there's a will to merge this and I'll work on the docker image. Thanks

kanaka:
@xixixao Yes, I would be interested in merging this.

There is some refactoring/cleanup that will need to happen first. The commit history is fine, but structure of the files needs to show the progression. Somebody wanting to learn from the implementation should be able to look at step3 and easily be able to identify the code that contributes directly to implementing step3 functionality. The use of `autoload.hack` and the common eval.hack file obscure this. The clear delineation is less important for some of the lower-level target implementation specific details (the type of stuff in `exceptions.hack` and `types.hack`). However, the code in `eval.hack` has a direct link to the implementation steps so it needs to be captured with its implementation. This does feel counter to normal develop practices (to not repeat identical code) but the goal of mal is educational for future reference and not just the implementer. More rationale is given in the FAQ https://github.com/kanaka/mal/blob/master/docs/FAQ.md#will-you-add-my-new-implementation You done this to some degree in the commit history, but that gets messy over time as fixes are applied and there aren't good processes/tools for running regressions tests and/or comparing differently named files across a git history. For similar reasons I think mal files should depend on other mal files explicitly so the code dependency are explicit to somebody trying to understand the code instead of automatic (again, a bit contrary to what you would do for a production non-teaching project).

Does the implementation self-host successfully?

Let me know when you have a Dockerfile that is ready for me to be build and pushed to docker hub. Once that happens then the test for this can be activated in Travis. With the changes I described above and passing tests in Travis then I'll do a bit deeper code review to see if it's ready for merging.

Thanks!

kanaka:
Hi @xixixao It's been a while. There is still no Hack implementation created for mal. Any interest in getting this finished up and merged?

Let me know if you have interest in continuing and finishing this up. If so, here is what needs to be completed:

* Rework the steps so that it follows the standard incremental process (no separate shared eval) as described above.
* The recommended process has also changed slightly: the eval_ast is no longer seperate from eval. It would be good if you could update to follow the new structure.
* Add a Dockerfile that installs the tools need to compile/run your implementation
* Move the implementation from `Makefile` to `Makefile.impls`. Also, add your implementation to `IMPLS.yml` in order to trigger CI for you implementation.
* Make sure that the Github CI tests all pass for your implementation. The process is now more automated and doesn't require my manual intervention to build docker images (part of the reason to rebase).

Thanks!

xixixao:
Realistically I won't have the time to work on this.

PR #505, opened 2020-05-10
Step 6: Test read-string of a single atom at the end of input

bjh21:
All of the existing `read-string` tests test forms whose end can be detected before the end of the input string, either because they're naturally self-terminating (lists, strings) or because there's white space after them.  All of the uses of the reader from the REPL are similarly likely to have input with a newline or similar at the end.

My BCPL implementation turned out to read one character past the end of the input when the input ended inside an atom, and to incorporate that one byte into the output.  The consequences of a buffer over-read are a little unpredictable, so I've just written a test that was good enough to catch the bug for me.

### DIFF
--- a/impls/tests/step6_file.mal
+++ b/impls/tests/step6_file.mal
@@ -9,6 +9,9 @@
 (read-string "(1 2 (3 4) nil)")
 ;=>(1 2 (3 4) nil)
 
+(= nil (read-string "nil"))
+;=>true
+
 (read-string "(+ 2 3)")
 ;=>(+ 2 3)
 


wasamasa:
Hm, it seems the change broke nothing, but I don't understand the test anyway. Wouldn't it make more sense to write a test where an implementation might read more than one value?

bjh21:
The point of this test is to cover the case where an atom is at the end of the input string.  The end of an atom (unlike other tokens) is detected by looking one character ahead in the input string to see if there's a non-atom character there.  This means that if you're writing an _ad hoc_ tokenizer in a language without bounds-checking and without null-terminated strings, you might end up reading off the end of the string into uninitialised memory.  That's what I did.

The bug in my code meant that when reading "nil", my implementation ended up with a four-character symbol whose first three characters were "nil" and whose fourth was whatever byte happened to be after "nil" in memory.  I found the bug while experimenting in the REPL trying to track down a completely different bug.  None of the existing tests up to step 6 found it.  It's not detectable before step 6 because up without `read-string` all reader input has a trailing newline, at least in my implementation.

It's possible there are other tests that will catch other bugs, and those tests might make more sense, but this test catches the bug I wrote, so I thought it might be useful to other people as well.

kanaka:
The test certainly doesn't hurt and could conceivably catch other types of bugs too. Merged.

PR #506, opened 2020-05-16
Update and fixes for the lua implementation.

gabrielmadruga:
I tried the lua implementation with lua 5.3.5 and found a various issues.

- Readme said it was tested with 5.2 but the makefile is looking for the libraries in a 5.1 directory. I updated the makefile to be version agnostic. I also added a step to install rex_pcre.so through luarocks.
- It seems it also needs LuaSocket. To make it easier I just commented the require and changed `ms-time` to use `os.clock` (won't give time in ms).
- `loadstring` is now `load` in newer versions of lua.
- Lua now supports unicode, so I changed `\177` to `\u{029e}` to be more like some of the other implementations.
- Some tests were failing, reason was that the `keyword` implementation was wrong.
- There is still one test soft failing that I haven't looked at in detail yet (I don't think it's lua version related):
```
Test that defining a macro does not mutate an existing function.
TEST: '(f (+ 1 1))' -> ['',true] -> SOFT FAIL (line 295):
    Expected : '\\(f\\ \\(\\+\\ 1\\ 1\\)\\)\r\ntrue'
    Got      : '(f (+ 1 1))\r\nfalse'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Each implementation of mal is separated into
 11 incremental, self-contained (and testable) steps that demonstrate
 core concepts of Lisp. The last step is capable of self-hosting
 (running the mal implementation of mal). See the [make-a-lisp process
-guide](process/guide.md). 
+guide](process/guide.md).
 
 The make-a-lisp steps are:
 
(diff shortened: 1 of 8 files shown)


kanaka:
Can you update the Dockerfile to install lua 5.3.5? Once you do that I'll test it locally and push out a docker image. After a final check in Travis then I'll merge. 

One comment, ms-time is suppose to be a millisecond counter. Couldn't you just multiple os.clock() by 1000000 to get milliseconds?

```
print(math.floor(os.clock()*1000000))
```

gabrielmadruga:
Updated the Dockerfile, I'm not 100% my update is ok, but I got it working on my machine.
I also multiplied os.clock() by 1000000. 
Is it ok that travis fails? I don't know where to check why.

Will keep working on my mal implementation. Thank you for the work you have done on this project!

kanaka:
I built and pushed a new image build and restarted the Travis test. Passed. Merged. Thanks for the update!

PR #507, opened 2020-05-18
julia 1.4 upgrade

ctzurcanu:
symbol -> Symbol
type -> struct / mutable struct
ismatch -> match  ... !== nothing


ctzurcanu:
probably Travis uses an older version of Julia where STDOUT was still capitalized... in the present version:
STD* -> std*

dubek:
Travis runs the tests in docker images which are built by @kanaka . When he gets to rebuild this image, it'll probably re-run the apt update thingy and get the latest julia release (and will break the current code in master?).

kanaka:
@ctzurcanu can you figure out what updates need to be made to the Dockerfile to be able to install julia-1.4? I did a trivial attempt to update the image to bionic, but doesn't seem like the julia ppa has bionic support. If you add that to the changeset then I'll build and push a new julia image.

ctzurcanu:
@kanaka image has been updated and `julia --version` was executing correctly after installation. However, I was unable to run `make DOCKERIZE=1 "test^julia"`, to see if tests pass. The read tests are the only ones that do not work, but maybe it is for causes external to Julia (readline).

also: check out: https://github.com/ctzurcanu/mali that runs Mal in an Interact.jl setup and the link to Observablehq
I would love to see a short manual of the Mal functions for ppl who forgot their lisp :D

kanaka:
The `/mal` path is where the mal sources get mounted so julia can't be installed there in the docker image. Also, julia-1.4 wants to write to a DEPOT_PATH on startup which by default is within the home directory. So that needs to be set (otherwise it defaults to an unwritable path). Here is a modification to the Dockerfile to get julia-1.4 to startup:

```
diff --git a/impls/julia/Dockerfile b/impls/julia/Dockerfile
index f345b72..442baff 100644
--- a/impls/julia/Dockerfile
+++ b/impls/julia/Dockerfile
@@ -24,6 +24,9 @@ WORKDIR /mal
 # Julia
 RUN apt-get -y install wget unzip
 RUN \
+  mkdir -p /opt && cd /opt && \
   wget https://julialang-s3.julialang.org/bin/linux/x64/1.4/julia-1.4.1-linux-x86_64.tar.gz && \
   tar zxpf julia-1.4.1-linux-x86_64.tar.gz && \
-  ln -s /mal/julia-1.4.1/bin/julia /usr/local/bin/julia
\ No newline at end of file
+  ln -s /opt/julia-1.4.1/bin/julia /usr/bin/julia
+
+ENV HOME /mal
```

However, once I did that, `make DOCKERIZE=1 "test^julia"` works, but step1 has failures in trying to read strings (and maybe other things):

```
TEST: '"abc"' -> ['',"abc"] -> FAIL (line 76):
    Expected : '\\"abc\\"\r\n\\"abc\\"'
    Got      : '"abc"\r\nnil'
```

ctzurcanu:
presently I do not know how to quickly fix the remainder of tests. I may return to this when I will know Julia better (probably not soon)

kanaka:
@ctzurcanu Any interest in revisiting this? The Github CI process has been improved significantly so I am no longer a manual part of the loop (docker images are built by the CI process directly).

ctzurcanu:
> @ctzurcanu Any interest in revisiting this? The Github CI process has been improved significantly so I am no longer a manual part of the loop (docker images are built by the CI process directly).

I would like to do it, but I am trying to get mal work in Lean 4. Learning Lean now. If I can't get it within a week, I can return to Julia 1.4. (I think Lean 4 would be more important to mal.)

kanaka:
Sounds good. Lean 4 definitely looks like an interesting language.

PR #509, opened 2020-05-21
ps: Fix handling of exceptions thrown from `catch*` clause

bjh21:
Here's a proposed fix for the bug in the PostScript implementation exposed by #508. The explanation in the commit message is substantially longer then the commit itself.


kanaka:
@bjh21 wow, nice catch and fix. :fire: Were you familiar with postscript prior to this?

bjh21:
Yes, I've done quite a bit with PostScript before, but always for the things it was actually designed for. I had to resort to the Red Book (well, a PDF of the Red Book) to find out how `stopped` worked because I'd never used it before.

PR #510, opened 2020-05-23
Simplify runtest.py's input handling

bjh21:
This is a successor to #469. There, I found that my BCPL implementation (now almost finished!) was tripping up `runtest.py` by omitting carriage returns in its output where they would have no effect. This appears to be a feature of the BCPL standard library and hard to avoid.  More recently, I've started on an APL implementation and found that GNU APL's readline-alike is very enthusiastic about emitting escape sequences.

This PR modifies `runtest.py` so that it's more tolerant of this kind of behaviour without the extravagant overhead of building an entire terminal emulator into it.

The important feature is that I've stopped `runtest.py` caring about how its input is echoed.  Previously when testing a form, it would check not only that the form was evaluated correctly, but also that it was echoed back correctly on the REPL command line.  This is unnecessary and unhelpful, since it almost all cases that echoing will be provided by that host-language readline library or the operating system terminal driver.  Even if those are somewhat under the control of the programmer, Mal should be primarily about making a LISP, not wrangling readline libraries. `runtest.py` now assumes that everything up to the first line feed after the prompt is echoed input, and ignores it. It still allows for the possibility of a second such line, to work around #30.

Similarly, I've stopped `runtest.py` caring about carriage returns at all.  These are inserted either by the host-language runtime or by the terminal driver, so they don't reflect whether the implementation works properly.  The input routine now strips them out and none of the rest of the code expects them.

These changes are sufficient to allow `runtest.py` to handle my nascent APL implementation, and almost enough for BCPL (there's another problem there, but I think it might be the implementation's job to fix that).

### DIFF
--- a/runtest.py
+++ b/runtest.py
@@ -28,8 +28,7 @@ def log(data, end='\n'):
     print(data, end=end)
     sys.stdout.flush()
 
-# TODO: do we need to support '\n' too
-sep = "\r\n"
+sep = "\n"
 rundir = None
 
 parser = argparse.ArgumentParser(
@@ -127,11 +126,7 @@ class Runner():
                 #print("new_data: '%s'" % new_data)
                 debug(new_data)
                 # Perform newline cleanup
-                if self.no_pty:
-                    self.buf += new_data.replace("\n", "\r\n")
-                else:
-                    self.buf += new_data
-                self.buf = self.buf.replace("\r\r", "\r")
+                self.buf += new_data.replace("\r", "")
                 for prompt in prompts:
                     regexp = re.compile(prompt)
                     match = regexp.search(self.buf)
@@ -216,10 +211,10 @@ class TestReader:
                     break
             if self.ret != None: break
 
-        if self.out[-2:] == sep and not self.ret:
+        if self.out[-1:] == sep and not self.ret:
             # If there is no return value, output should not end in
             # separator
-            self.out = self.out[0:-2]
+            self.out = self.out[0:-1]
         return self.form
 
 args = parser.parse_args(sys.argv[1:])
--- a/runtest.py
+++ b/runtest.py
@@ -132,13 +132,6 @@ class Runner():
                 else:
                     self.buf += new_data
                 self.buf = self.buf.replace("\r\r", "\r")
-                # Remove ANSI codes generally
-                #ansi_escape = re.compile(r'\x1B\[[0-?]*[ -/]*[@-~]')
-                # Remove rustyline ANSI CSI codes:
-                #  - [6C - CR + cursor forward
-                #  - [6K - CR + erase in line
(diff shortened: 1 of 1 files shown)


bjh21:
The test failure is in the PL/PgSQL implementation when running the step 6 tests under step A. The failure is in the stand-alone `*ARGV*` test, which this PR doesn't touch. It works when I test it locally:

```
bjh21@sole:~/mal$ make DOCKERIZE=1 REGRESS=1 OPTIONAL=0 HARD=1 test^plpgsql^stepA
[lots of output]
Testing ARGV of test^plpgsql^stepA; step file: impls/plpgsql/stepA_mal.sql
Running: docker run -e STEP=stepA_mal -e MAL_IMPL=js -it --rm -u 1001 -v /home/bjh21/mal/:/mal -w /mal/impls/plpgsql kanaka/mal-test-plpgsql ../tests/run_argv_test.sh ../plpgsql/run 
ERROR: ld.so: object 'libgtk3-nocsd.so.0' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
Waiting for postgres to start
psql:init.sql:5: NOTICE:  schema "io" does not exist, skipping
psql:init.sql:5: NOTICE:  schema "types" does not exist, skipping
psql:init.sql:5: NOTICE:  schema "reader" does not exist, skipping
psql:init.sql:5: NOTICE:  schema "printer" does not exist, skipping
psql:init.sql:5: NOTICE:  schema "envs" does not exist, skipping
psql:init.sql:5: NOTICE:  schema "core" does not exist, skipping
psql:init.sql:5: NOTICE:  schema "mal" does not exist, skipping
psql:init.sql:8: NOTICE:  extension "hstore" does not exist, skipping
psql:init.sql:11: NOTICE:  extension "dblink" does not exist, skipping
OK: '("aaa" "bbb" "ccc")'
psql:init.sql:5: NOTICE:  drop cascades to 152 other objects
OK: '()'
Passed all *ARGV* tests

[more output]
TEST RESULTS (for ../tests/stepA_mal.mal):
    0: soft failing tests
    0: failing tests
   12: passing tests
   12: total tests

bjh21@sole:~/mal$ 
```

kanaka:
@bjh21 Looks like it was a transient failure. I restarted the test and it passed in Travis.

Great work! This is definitely an area that was in desperate need of improvement. Any chance you have Cygwin and could test that it works there?  A little of the line-ending complexity was due to supporting Windows. Not all the implementations will run/pass of course, but it would be good for people to be able to do the process on windows. I think something basic like python should be able run in Cygwin. I merged it because we can address Windows issues after, but I'm hoping somebody has the ability to test this on Windows (I don't easily).

bjh21:
I'm afraid I'm not currently in a position to test on Windows at the moment: I don't think there's even a Windows system in the house.

PR #511, opened 2020-05-27
BCPL implementation

bjh21:
Here is my BCPL implementation of mal.  It's _almost_ ready for merging, but there's one problem I'd like some guidance on.

The Cintsys BCPL implementation likes to print a banner at startup, and I've not yet found a way to stop it.  This means that starting Mal looks like this:
```
bjh21@sole:~/mal/impls/bcpl$ ./run

BCPL 32-bit Cintcode System (3 Jan 2019)
0.000> 
Mal [BCPL]
user> 
```
The main test script, `runtest.py` is quite happy to allow the implementation to output rubbish before the prompt (at least, once I'd fixed its prompt detection it was), but the ARGV test in step 6 is much stricter.  Arguably this discrepancy is reasonable: `runtest.py` is running the interpreter in non-interactive mode, while `run_argv_test.sh` is running it on non-interactive mode.

So my question is: should I amend `run_argv_test.sh` to permit extra output at startup, or should I find a way to stop BCPL producing that output?


wasamasa:
You could patch it to not display the message since you're building from source anyway or maybe even suggest the maintainer to add a `-q` option to supress the banner. Searching for "bit Cintcode System" shows a few code matches, might be sufficient to edit `cintcode/sysb/boot.b`.

impls/bcpl/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+##########################################################
+# Specific implementation requirements
+##########################################################
+
+RUN apt-get -y install gcc mesa-common-dev libglu1-mesa-dev freeglut3-dev
+
+RUN cd /usr/local && \
+    curl -o bcpl.tgz https://www.cl.cam.ac.uk/~mr10/BCPL/bcpl.tgz && \
+    tar zxf bcpl.tgz && chmod -R a+rX BCPL
```
kanaka, line 28: You could delete bcpl.tgz as a final step to shrink this layer.

bjh21:
OK. I'm quite happy to use mal as a means of making toy languages (or toy implementations) less so.  Also on the agenda: some way for a program to cause GNU APL to exit.

impls/bcpl/Dockerfile
bjh21: Ah yes. Too long spent writing Ansible playbooks for VMs, where a few magabytes either way don't really matter, I'm afraid.

bjh21: Should be fixed by 0149fc6.

kanaka:
Hi @bjh21. Any word on this? I know I've been incommunicado for the past year, but I should be partly back and paying attention now and I would still love to get BCPL into the main tree. I did a rebuild and re-test of the image and it looks like the banner is still a problem.

kanaka:
@bjh21 It's been a few years. Any interest in pushing this forward or should I close the ticket. It would be great to get a BCPL implementation merged, but I can understand if you are no longer interested/motivated.

PR #512, opened 2020-05-28
Step 2: Test empty vectors and hash-maps

bjh21:
An empty vector or hash-map should evaluate to itself, but I reached the test of `(empty? [])` in step 4 before I discovered that I'd got `[]` evaluating to `0`.  Add tests for both cases to step 2.  This shouldn't cause any new failures, unless I suppose there's an implementation where `(empty? 0)` is true. It just means that this bug in `eval_ast` can be caught before it propagates to two more stage files.
- [x] yorick

### DIFF
--- a/impls/tests/step2_eval.mal
+++ b/impls/tests/step2_eval.mal
@@ -41,3 +41,9 @@
 
 {:a (+ 7 8)}
 ;=>{:a 15}
+
+;; Check that evaluation hasn't broken empty collections
+[]
+;=>[]
+{}
+;=>{}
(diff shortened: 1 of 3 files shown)


bjh21:
Well I never! It fails on one existing implementation:
- [ ] yorick
```
TEST: '[]' -> ['',[]]
Exception: IOError(5, 'Input/output error')
Output before exception:
[]
ERROR (eval_ast) bad dimension list
  LINE: 29  FILE: /mal/impls/yorick/step2_eval.i
yorick: quitting on error in batch mode
```

bjh21:
Incidentally, the problem I had turned out to be a bug in GNU APL, now fixed:
https://lists.gnu.org/archive/html/bug-apl/2020-05/msg00045.html

bjh21:
I think the yorick fix is obvious enough that it doesn't deserve its own PR.

PR #513, opened 2020-05-29
step 4: Test that (= (list) ()) is true

bjh21:
Another  test of empty things, because those are special in APL.  Since #512 smoked out a bug elsewhere, I think this one is worth trying.

In APL, empty arrays have type, and the `≡` function pays attention to that, so if you're not careful empty lists produced by different means compare different.  This test detects that by comparing an empty list produced by `list` with one produced by evaluating an empty list.

My bug was found by a deferrable step 4 test, but the bug should be non-deferrable.

- [ ] fantom
- [x] jq

### DIFF
--- a/impls/tests/step4_if_fn_do.mal
+++ b/impls/tests/step4_if_fn_do.mal
@@ -119,6 +119,8 @@
 
 (= (list) (list))
 ;=>true
+(= (list) ())
+;=>true
 (= (list 1 2) (list 1 2))
 ;=>true
 (= (list 1) (list))


bjh21:
Two failures: both fantom and jq return `false` rather than `true`.

dubek:
I'm working on the fantom fix for this, and found a related one (in fantom):

```
user> (map str ())
()
user> (= (list) (map str ()))
false
```

It appears that fantom's internal map function returns a list of `Obj?` items unless explicitly typed to return a list of `MalVal` items. So the equality check fails because the two empty lists do not have same element type signature.

dubek:
It looks like all the Github Actions jobs are green: https://github.com/kanaka/mal/actions/runs/761581623

PR #514, opened 2020-05-30
Jq: Fix our share of #513

alimpfard:
Explicit state is such a pain :)


alimpfard:
Note: this issue is created in step3, by the addition of explicit env wraps, and removed in step5 (since the env wrap is redesigned in step5 to fit TCO, and the forgotten wrap is added back).

kanaka:
Travis passes and I tested it locally with the fix and it passes the #513 test case. Merged.

PR #515, opened 2020-06-01
runtest: Treat test timeouts as fatal

bjh21:
If a test times out, that means the implementation hasn't returned a prompt within reasonable time.  Continuing to type at it is unlikely to be helpful, since the lack of a prompt indicates that the subprocess has hung, so subsequent tests are likely to time out as well.

This is particularly annoying in step 5, where the line that usually times out, `(def! res2 (sum2 10000 0))` has its output ignored, so a timeout is treated as success, meaning that runtest proceeds to the next test, which of course also times out.

This PR converts timeouts into fatal errors (through the existing exception-handing mechanism) so that a timeout immediately aborts the entire test.

### DIFF
--- a/runtest.py
+++ b/runtest.py
@@ -264,6 +264,9 @@ fail_cnt = 0
 soft_fail_cnt = 0
 failures = []
 
+class TestTimeout(Exception):
+    pass
+
 while t.next():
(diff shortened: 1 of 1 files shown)


kanaka:
Yes, that's a good improvement. Thanks.

PR #517, opened 2020-06-07
C2

fungiblecog:
Hi Joel

You probably don't need another C implementation, but I'll offer it anyway.

Reasons you might want to include it:
- It passes all the tests (I had some problems with the existing c implementation in this regard)
- It has fewer dependencies (readline, garbage collector, FFI only). As this was a learning exercise I wrote my own simple linked list and used it for lists, vectors and hashmaps rather using libraries
- I think the style - while probably not idiomatic - is easier for beginners to understand (this is my first real c program so my priority was consistency and readability)
- It has extensive error handling

Cheers
Duncan


dubek:
Hi @fungiblecog, thanks for your contribution. Very impressive for a first C project.

Indeed the current C implementation has issues.

I think that @kanaka has another C implementation in the works; I believe he planned it with zero dependencies (just the standard C library) to allow running it without an operating system.

Even if the decision is not to merge, maybe we can link to your implementation.

fungiblecog:
Hi @dubek 

I'm happy with whatever you and Joel decide - merge or link. I really just wanted Joel to know there are people out there who really appreciate what he's done. I never would have taken on a project like this without the brilliant guide and especially the test framework. It has bee a fantastic learning experience and now that I'm addicted I intend to keep developing it. Next step will be to add real persistent data structures like clojure's...

wasamasa:
Some stylistic things you might want to fix:

- The Makefile could use less redundancies, with built-in rules you can set a few variables for compiler/linker flags, only declare source/destination files and omit the explicit compiler calls
- Indentation uses a mix of spaces and tabs, I'd choose either and stick to it

wasamasa:
To further clarify, I mean a Makefile as simple as this:

```make
LDLIBS= -lgc -lreadline
CFLAGS = -Wall -Wextra -Wpedantic -std=c99 -O2

all: test repl

test: test.o util.o
repl: repl.o
util.o: util.h

clean:
	$(RM) *.o test repl
```

The reason this works is the existence of a built-in rule compiling every .c to its matching .o file (while taking the variables `CFLAGS` and `LDLIBS` into account) and another one linking several .o files into an extensionless file. Try `make -p` to see these rules.

fungiblecog:
@wasamasa - Dude, I can spend my time learning the syntax for Makefiles, or I can keeping working on something interesting like my persistent hashmap implementation. I'm doing this for fun. I don't care if the Makefile is less than perfect! I'm also terrible with git but learning that stuff is not my priority.

kanaka:
Hi @fungiblecog . Apologies for the very slow response getting back to you on this. I haven't been doing any mal/open source activity for the past year or so. I am actually interested in merging this into the main tree if you are willing to make a couple of updates:
* Rename the implementation to `c.2` (rather than `c2`) to disambiguate this as another implementation of mal in C as opposed to being an implementation in C version 2 (e.g. `es6`, `perl6`, `swift5`, etc).
* Rebase to pick up some new tests and make sure Travis/Github Actions pass with the new tests (I've built and pushed a docker build named `kanaka/mal-test-c.2` so that when you rename and update your pull request, the Travis/Github CI tests should be able to find the docker image to test with).

If you are willing to make those changes I would be happy to merge it. If you don't want to make the changes or don't otherwise have time to make the changes (I certainly understand life changes that may get in the way), I'd be happy to link to your repo/implementation in the README in lieu of merging into the main tree. Just let me know if you want to do that instead.

PR #518, opened 2020-06-13
impls/bash: Minor changes to make it run under OSH

andychu:
OSH is a bash-compatible shell: https://www.oilshell.org/

reader.sh:

- Put the constant regex pattern in a string literal.  This simplifies
  it by removing mixed regex vs. shell quoting, and implicit
  concatenation with $'\n'.

  This is suggested by the bash manual:
  https://www.gnu.org/software/bash/manual/bash.html#Conditional-Constructs

  "Storing the regular expression in a shell variable is often a useful
  way to avoid problems with quoting characters that are special to the
  shell."

- Initialize __reader_tokens as an array, not a string.

  https://www.oilshell.org/release/0.8.pre6/doc/known-differences.html#values-are-tagged-with-types-not-cells

env.sh:

Simplify quoting in 'eval' expressions.  This quotes associative array keys,
which is required by OSH to avoid dynamic parsing.

  https://www.oilshell.org/release/0.8.pre6/doc/known-differences.html#strings-vs-bare-words-in-array-indices
  http://www.oilshell.org/blog/2016/10/20.html

core.sh:

Quote associative array keys.  '<' and '>' are shell operators and OSH doesn't
have a special case when inside [].

----

With this change, OSH can run tests just like bash, e.g.:

$ osh impls/bash/stepA_mal.sh tests/step4_if_fn_do.mal

----

Test results are the same before and after this change:

$ NO_DOCKER=1 ./.travis_test.sh test bash

FAILURES:
SOFT FAILED TEST (line 295): (f (+ 1 1)) -> ['',true]:
    Expected : '.*\ntrue'
    Got      : '(f (+ 1 1))\nfalse'

TEST RESULTS (for ../tests/stepA_mal.mal):
    1: soft failing tests
    0: failing tests
  106: passing tests
  107: total tests

### DIFF
--- a/impls/bash/core.sh
+++ b/impls/bash/core.sh
@@ -376,10 +376,10 @@ declare -A core_ns=(
     [readline]=readline
     [read-string]=read_string
     [slurp]=slurp
-    [<]=num_lt
-    [<=]=num_lte
-    [>]=num_gt
-    [>=]=num_gte
+    ['<']=num_lt
+    ['<=']=num_lte
+    ['>']=num_gt
+    ['>=']=num_gte
     [+]=num_plus
     [-]=num_minus
     [__STAR__]=num_multiply
--- a/impls/bash/env.sh
+++ b/impls/bash/env.sh
@@ -48,7 +48,7 @@ ENV_FIND () {
         r="${1}"
     else
         local obj="${ANON["${1}"]}"
-        eval local outer="\${${obj}["__outer__"]}"
+        eval 'local outer=${'${obj}'["__outer__"]}'
         if [[ "${outer}" && "${outer}" != "${__nil}" ]]; then
             ENV_FIND "${outer}" "${2}"
         else
@@ -66,7 +66,7 @@ ENV_GET () {
     local key="${ANON["${2}"]}"
     if [[ "${r}" ]]; then
         local obj="${ANON["${env}"]}"
-        eval r="\${${obj}["${key}"]}"
+        eval 'r=${'${obj}'["'${key}'"]}'
     else
         _error "'${key}' not found"
     fi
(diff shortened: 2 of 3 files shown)


andychu:
Great thanks for the quick review :-)

kanaka:
Nice! I wasn't previously familiar with OSH. I've gone ahead and merged this PR but I wonder if you might want to add OSH to the test matrix as a bash mode? That way we won't accidentally break OSH in the future without realizing it. It could also serve as one form of regression test for new versions of OSH itself. It might be challenging to find a single purpose bash script that exercises more features of bash that the mal implementation :smile:. Also, it would be interesting to see the performance micro-benchmark comparison of OSH to regular bash.

Want to tackle that? It would require:
- adding an OSH install to the bash Dockerfile (after which I can build it and push to docker hub)
- adding a `bash_MODE` variable to the top-level Makefile and conditionalizing the bash `run` file 
- add to the Travis matrix in `.travis.yml` to split the bash modes

Python and clojure are model and you can refer to them for reference.

andychu:
Yes I think it would be a great test and it's very much the direction I've been thinking of.  I did run the perf tests and that's one of the things that made me want to run mal with OSH (it uncovered a couple OSH bugs like negative indices in string slices)

Although I think it makes more sense to wait until the C++ version is ready:

http://www.oilshell.org/blog/2020/05/translation-progress.html

Long story but Oil is written in Python and semi-automatically translated to C++.

The Python version is about 10x slower than bash now on the `step4` test file, e.g. 15 seconds vs. 1.5 seconds.  So I am very curious to see how fast the C++ version runs, and that's one of the reasons I tried mal.

A  relatively "pure" bash script is a nice benchmark to isolate the shell's speed, rather than one that does a lot of I/O.

I've also run a pure bash brainfuck interpreter and a JSON parser.  Along with mal, that will be a nice set of benchmarks.  I'll probably be back when OSH is further along :)

kanaka:
Sounds good :+1:

PR #519, opened 2020-06-14
ruby: (keyword :abc) should just return same keyword

dubek:
Before the fix Ruby had one soft fail in step9:

```
TEST: '(keyword :abc)' -> ['',:abc] -> SOFT FAIL (line 386):
    Expected : '.*\n\\:abc'
    Got      : '(keyword :abc)\n:\xca\x9eabc'
```

The fix causes `(keyword :abc)` to return its keyword argument without modification.


kanaka:
Thanks!

PR #520, opened 2020-06-21
Fantom fixes

dubek:
* Fix #513 in Fantom and similar problems with Fantom's interop (all related to empty lists and hash-maps)
* Fix broken step1 (and step6) test involving a `!` char - requires rebuilding the Docker image because I've added a java config option to disable jLine's expand events. I've also moved that test to be mandatory in step1 (because the comment said that fantom is the only offender)
* Fix step9 optional test - `(keyword :abc)`
* Fix stepA optional test - defmacro! doesn't mutate the original function

### DIFF
--- a/impls/fantom/Dockerfile
+++ b/impls/fantom/Dockerfile
@@ -31,7 +31,8 @@ RUN cd /tmp && curl -sfLO https://bitbucket.org/fantom/fan-1.0/downloads/fantom-
     && mv fantom-1.0.70 /opt/fantom \
     && cd /opt/fantom \
     && bash adm/unixsetup \
-    && curl -sfL -o /opt/fantom/lib/java/jline.jar https://repo1.maven.org/maven2/jline/jline/2.14.6/jline-2.14.6.jar
+    && curl -sfL -o /opt/fantom/lib/java/jline.jar https://repo1.maven.org/maven2/jline/jline/2.14.6/jline-2.14.6.jar \
+    && sed -i '/java.options/ s/$/ -Djline.expandevents=false/' /opt/fantom/etc/sys/config.props
 
 ENV PATH /opt/fantom/bin:$PATH
 ENV HOME /mal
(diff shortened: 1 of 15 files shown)


kanaka:
While I've been away, looks like the bitbucket hosting of fantom changed (and the props file changed a bit too). Here is the change I had to make to get it to build and pass the tests:

```
--- a/impls/fantom/Dockerfile
+++ b/impls/fantom/Dockerfile
@@ -25,14 +25,14 @@ WORKDIR /mal
 RUN apt-get -y install openjdk-8-jdk unzip
 
 # Fantom and JLine
-RUN cd /tmp && curl -sfLO https://bitbucket.org/fantom/fan-1.0/downloads/fantom-1.0.70.zip \
-    && unzip -q fantom-1.0.70.zip \
-    && rm fantom-1.0.70.zip \
-    && mv fantom-1.0.70 /opt/fantom \
+RUN cd /tmp && curl -sfLO https://github.com/fantom-lang/fantom/releases/download/v1.0.75/fantom-1.0.75.zip \
+    && unzip -q fantom-1.0.75.zip \
+    && rm fantom-1.0.75.zip \
+    && mv fantom-1.0.75 /opt/fantom \
     && cd /opt/fantom \
     && bash adm/unixsetup \
     && curl -sfL -o /opt/fantom/lib/java/jline.jar https://repo1.maven.org/maven2/jline/jline/2.14.6/jline-2.14.6.jar \
-    && sed -i '/java.options/ s/$/ -Djline.expandevents=false/' /opt/fantom/etc/sys/config.props
+    && sed -i '/java.options/ s/^\/\/ *\(.*\)$/\1 -Djline.expandevents=false/' /opt/fantom/etc/sys/config.props
 
 ENV PATH /opt/fantom/bin:$PATH
 ENV HOME /mal
```

I've updated the docker image based on the changes above so you should be able to test them in Travis/Github Actions now. Once you've applied the above to you PR and things pass I'll merge it in.

dubek:
I fixed fantom's Dockerfile in 5f18c618d1c71420fee888e09b4d557bd860b042 .

PR #521, opened 2020-06-21
Test equality of other kinds of empty list

bjh21:
In #513, @dubek noted two additional ways to make empty lists in fantom that compare different from an empty list made by evaluating an empty list.  This adds tests for those, along with another two that might fail similarly (but don't in fantom).

### DIFF
--- a/impls/tests/step7_quote.mal
+++ b/impls/tests/step7_quote.mal
@@ -27,6 +27,8 @@ a
 ;=>()
 (concat (list) (list))
 ;=>()
+(= () (concat))
+;=>true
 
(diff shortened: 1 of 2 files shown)


dubek:
Consider adding also: `(= {} (hash-map))` in step 9.

bjh21:
@dubek: That's a good plan.  I've added it and another obvious one.

dubek:
Travis looks OK except fantom (as expected). Fantom fixes are in #520 .

kanaka:
Tests look good. Merged.

PR #523, opened 2020-07-12
Should not ignore test_timeout

soiamsoNG:
Should not ignore test_timeout if set to none default value (20)
Fix bug on line 305, to guard the res to avoid Exception raise form re.search

### DIFF
--- a/runtest.py
+++ b/runtest.py
@@ -296,13 +296,13 @@ while t.next():
         res = r.read_to_prompt(['\r\n[^\s()<>]+> ', '\n[^\s()<>]+> '],
                                 timeout=args.test_timeout)
         #print "%s,%s,%s" % (idx, repr(p.before), repr(p.after))
-        if (res == None):
+        if (res == None and (args.test_timeout == 20)):
             log(" -> TIMEOUT (line %d)" % t.line_num)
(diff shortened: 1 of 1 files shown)


dubek:
Thanks for your contribution.
Can you please describe what scenario was failing before this suggested fix?

soiamsoNG:
I use runtest.py without make, to continue a long time stop learning in 2017

I use version 9d7fb0fd8d93a619855fac4e3d4503a1ce299fdf
./runtest.py --test-timeout 1 step0.mal ./step0_main

step0_main is application can not return any thing
then it raise Timeout Exception behaviour not like before 

So I go back one version ab7ecbbcad1eea71d3b1a807a256b65302fbba09
then raise TypeError('expected string or bytes-like object')
behaviour not like version before  f6f5d4f2a3b708d9f272dfef82d21c54a2642cc0

kanaka:
Hi @soiamsoNG. Apologies for the long delay in getting back to you. I'm not quite following your description of what this accomplishes. In what other contexts (besides the specific one you are working on) would this change be useful? In addition, having a hard-coded value of "20" for checking the timeout isn't the best way to do this check.

soiamsoNG:
A little bit long time， I also forget how to reproduce

PR #527, opened 2020-08-21
haskell: clarify reader

asarhaddon:
The grammar is not fully equivalent to the one in the process, but it
passes all tests.

I suggest that it becomes a formal/testable reference (are tabs
allowed as spaces? is "a~" valid? is "' a" valid?...).

As it is formuled here, it can directly be translated either to
another language with high-level parsers, or to a low-level language
as each production only switches after a look at the next input
character.

### DIFF
--- a/impls/haskell/Core.hs
+++ b/impls/haskell/Core.hs
@@ -5,7 +5,7 @@ where
 import System.IO (hFlush, stdout)
 import Control.Monad.Except (throwError)
 import Control.Monad.Trans (liftIO)
-import qualified Data.Map as Map
+import qualified Data.Map.Strict as Map
 import Data.Foldable (foldlM)
 import Data.Time.Clock.POSIX (getPOSIXTime)
 import Data.IORef (newIORef, readIORef, writeIORef)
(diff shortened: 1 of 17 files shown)


kanaka:
@asarhaddon code changes looks fine. Can you update the PR to not modify the travis test file and also rebase (to pull in the addition of github actions)?

asarhaddon:
I cannot see the logs of failed checks in the web interface.  On my machine, the new tests pass on awk, bash and perl. What should I do?

kanaka:
Finally had a few minutes to look at this. It's been long enough I need to push to a new branch to get Github Actions testing it again. There were a couple failures for unrelated issues so I'm merging this now.

In terms of formal specification, I think that would be reasonable documentation for a specific implementation (e.g. the Haskell  implementation), however, I want to stay away from a strict spec that implementors would need to adhere to in order to pass all the tests. Admittedly, this is contrary to the normal goal of language design where you want the formal specification to be as complete and specific as possible. In the general mal case, the goal is reasonably short path for implementation of a working interpreter in any target language. There are certainly places where being very specific helps this goal (because it eliminates subtle problems and headaches later). But there are also cases where locking down the edges of the language would mean more work for the casual mal implementator (the precise set of characters included and excluded from symbols/keywords, error message text/format, etc)

PR #530, opened 2020-09-07
Fix typos in step 7 - quoting

mpritham:


wasamasa:
While you're at it, could you please fix the unbalanced parentheses in the two lines above as well?

mpritham:
I've done so.

wasamasa:
Thanks!

PR #532, opened 2020-10-02
vhdl: Fix defmacro! so it doesn't modify existing functions

dubek:
Fix a soft failure in vhdl stepA tests.


dubek:
The `IMPL=vhdl` job in travis is green: https://travis-ci.org/github/kanaka/mal/jobs/732329504

PR #535, opened 2020-10-17
vimscript: defmacro! doesn't modify existing functions

dubek:
Fix soft failure in tests of stepA.


dubek:
The `IMPL=vimscript` job in travis is green: https://travis-ci.org/github/kanaka/mal/jobs/736686362

PR #537, opened 2020-10-26
tcl: Soft failure fixes

dubek:
Fixes of soft test failures in step 9 and step A:

- tcl: Fix (keyword) of keywords
- tcl: defmacro! doesn't modify existing functions


dubek:
The tcl job in travis is green: https://travis-ci.org/github/kanaka/mal/jobs/739057410

Issue #536, opened 2020-10-22
make doesn't work as expected


vrruiz:
Step 0 says: ``Create a step0_repl.qx file in quux/``. 

However, by doing a ``make quux``, Makefile calls ``impls/quux``. In my case:

```
IMPLS = rust 
rust_STEP_TO_PROG =          rust/$($(1)).rs
```

```
[rvr@localhost mal]$ make "test^rust^step0"
make -C impls/step0_repl.rs rust/step0_repl.rs
make[1]: *** impls/step0_repl.rs: No existe el fichero o el directorio.  Alto.
make: *** [Makefile:390: rust/step0_repl.rs] Error 2
```

So, even if the path on the Makefile has been changed to ``rust/`` instead of ``impls/rust``, it goes to ``impls``.

dubek:
Thanks @vrruiz  for noticing this and reporting. The Makefile is OK; the guide should be fixed.

Historically all the implementations were at the root of the project. At some point they were all moved to the `impls` directory and the Makefile was updated, but I assume there are missing places to fix in the guide.

If you wish, you can fix the guide in the relevant places and open a PR.

Again, thanks for reporting; hope you're having fun building your Mal implementation.

PR #539, opened 2020-10-31
Mention impls/quux instead of quux when needed

wasamasa:
See #536. Another thing I'm less sure about: Do the paths mentioning tests/ and process/ need to be adjusted, too?

### DIFF
--- a/process/guide.md
+++ b/process/guide.md
@@ -96,7 +96,7 @@ cd mal
 * Make a new directory for your implementation. For example, if your
 language is called "quux":
 ```
-mkdir quux
+mkdir impls/quux
 ```
 
 * Modify the top level Makefile to allow the tests to be run against
@@ -106,7 +106,7 @@ mkdir quux
 ```
 IMPLS = ... quux ...
 ...
-quux_STEP_TO_PROG = mylang/$($(1)).qx
+quux_STEP_TO_PROG = impls/quux/$($(1)).qx
 ```
 
 * Add a "run" script to your implementation directory that listens to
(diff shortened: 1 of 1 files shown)


process/guide.md
```diff
@@ -1690,14 +1690,15 @@ implementation.
...
+    and returns the result converted to the relevant Mal type. You may
+    also add other interop functions as you see fit; Clojure, for
     example, has a function called `.` which allows calling Java
     methods. If the target language is a static language, consider
     using FFI or some language-specific reflection mechanism, if
     available. The tests for `quux-eval` and any other interop
-    function should be added in `quux/tests/stepA_mal.mal` (see the
-    [tests for `lua-eval`](../lua/tests/stepA_mal.mal) as an example).
+    function should be added in `impls/quux/tests/stepA_mal.mal` (see
+    the [tests for `lua-eval`](../lua/tests/stepA_mal.mal) as an
```
dubek, line 1700:
```suggestion
    the [tests for `lua-eval`](../impls/lua/tests/stepA_mal.mal) as an
```

dubek:
It seems like `process` stays at the root dir.

As for `tests`, it was moved into `impls/`, so I guess we should update all the documentation and scripts and eventually remove the `tests` symlink from the root dir. 

@kanaka - any comments?

kanaka:
I merged the update. @dubek I wanted to keep the tests actually at the top-level, but there was some situation where something wouldn't follow the symlink. It's been too long now and I don't remember what it was. I think I would prefer to keep the `tests` symlink since that *should* be where it lives.

PR #541, opened 2020-11-25
Add Fennel implementation

sogaiu:
This is an implementation of mal in [Fennel](https://fennel-lang.org/).

All tests pass in both direct and self-hosted modes:

make "test^fennel"
make MAL_IMPL=fennel "test^mal"

I also used a peg in this implementation and not regular expressions.

Included in the PR is a Dockerfile I successfully tested in a VM.

I haven't changed the README yet -- thought I might wait to see what happens with the Janet PR.

### DIFF
--- a/IMPLS.yml
+++ b/IMPLS.yml
@@ -26,6 +26,7 @@ IMPL:
   - {IMPL: es6}
   - {IMPL: factor}
   - {IMPL: fantom}
+  - {IMPL: fennel}
   - {IMPL: forth}
   - {IMPL: fsharp}
   - {IMPL: go}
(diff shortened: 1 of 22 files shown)


kanaka:
@sogaiu I've pushed a docker image for this impl also. The fennel implementation appears to have the same hanging behavior as the janet implementation. I suspect the reason is because the `time-ms` function is not returning milliseconds but rather a small fractional number and so the perf3.mal code is not properly detecting the passage of 10 seconds of time.

kanaka:
The reason the fennel implementation failed just now is because it's not fully defined in the `Makefile.impls` file. At some point I may try and combine the data from `Makefile.impls` and `IMPLS.yml` but for now it needs to be specified in both places.

sogaiu:
@kanaka Sorry about that.  I added fennel to `Makefile.impls`.

PR #545, opened 2021-01-03
Add missing expected result for test case

jamesroutley:
Heya - I might be wrong, but I think this line needs to be added - it's the expected output from evaluating the line above - `1; &()*+,-./:;<=>?@[]^_{|}~`

### DIFF
--- a/impls/tests/step1_read_print.mal
+++ b/impls/tests/step1_read_print.mal
@@ -281,6 +281,7 @@ false
 ;=>1
 ;;; Hopefully less problematic characters
 1; &()*+,-./:;<=>?@[]^_{|}~
+;=>1
 
 ;; FIXME: These tests have no reasons to be optional, but...
 ;; fantom fails this one


dubek:
Thanks, the fix looks good to me.

Many of the travis builds fail due to:

```
+docker run -it -u 2000 -v /home/travis/build/kanaka/mal:/mal kanaka/mal-test-ts make -C impls/ts
Unable to find image 'kanaka/mal-test-ts:latest' locally
docker: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
See 'docker run --help'.
The command "./.travis_test.sh build ${IMPL}" exited with 125.
```

Can we tell Travis not to run more than N jobs in parallel?

Also, there's a message about moving from travis-ci.org to travis-ci.com, not sure what's the effect on mal's CI configuration. @kanaka ?

kanaka:
@dubek yeah, docker Hub started limiting pulls to 100 every 6 hours for non-authenticated pulls. Which is a pain when we have over 100 tests that want to run and each pull down a docker image. I've submitted an open source application just now to try and get the limit for that increased. We'll see if they get back to me and are willing (might have to convert to an github/docker organization to do so which might not be a bad idea anyways). We'll see what they say.

PR #546, opened 2021-01-04
Deno implementation of MAL

graeme-lockley:
Hi,

I have created a Deno implementation of MAL.  The code itself is Typescript and therefore you might elect to see it as another Typescript implementation however, the cool thing with Deno, is there is no package management - you simply type

  deno run --allow-all step<whatever>.ts 

it then it just runs.

All of the tests pass, self hosting, docker image and the travis script runs.

Very keen to hear your thoughts and comments.

Thank you so much for creating and maintaining MAL!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ DOCKERIZE =
 #
 
 IMPLS = ada ada.2 awk bash basic bbc-basic c chuck clojure coffee common-lisp cpp crystal cs d dart \
-	elisp elixir elm erlang es6 factor fantom forth fsharp go groovy gnu-smalltalk \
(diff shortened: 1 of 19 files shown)


kanaka:
@graeme-lockley sorry for the slow response. I don't think the use of deno is itself enough justification for adding the implementation to the main tree. However, if you can make an argument that your implementation is different in some interesting way then I would definitely consider merging it into the tree. Examples: more idiomatic, different but well accepted style of TypeScript code, better performance in some aspect, more portable, etc. Basically, the purpose of having implementations in the main repo is pedagogical; for reference and comparison (both between languages and between implementations in the same language). Alternately, I am always happy to mention your implementation in the README even if I don't end up merging it.

PR #547, opened 2021-01-12
Suggest use StdIn.ReadLine instend of Predef.ReadLine cause Predef.ReadLine has been deprecated

zsdz:
method ReadLine defined in predef has been deprecated since scala 2.11.0

I change the code with StdIn.ReadLine in step0_repl.scala,and it works fine

If this pull will be accepted,I will change other code in the files

### DIFF
--- a/impls/scala/step0_repl.scala
+++ b/impls/scala/step0_repl.scala
@@ -1,3 +1,5 @@
+import scala.io.StdIn
+
 object step0_repl {
   def READ(str: String): String = {
     str
@@ -17,7 +19,7 @@ object step0_repl {
 
   def main(args: Array[String]) {
     var line:String = null
-    while ({line = readLine("user> "); line != null}) {
+    while ({line = StdIn.readLine("user> "); line != null}) {
       try {
         println(REP(line))
       } catch {


kanaka:
@zsdz sorry for the very slow followup. I would definitely consider merging this if you changed all the files and also cleaned up the commit to have a better commit summary like "[scala] use StdIn.ReadLine instead of deprecated Predef.readLine" and if you can, add a link in the commit body to a notice of this being deprecated.

zsdz:
Thanks.But github can't be connected since 2021 from china.I will commit later



***@***.***
 
From: Joel Martin
Date: 2021-04-22 05:50
To: kanaka/mal
CC: zsdz; Mention
Subject: Re: [kanaka/mal] Suggest use StdIn.ReadLine instend of Predef.ReadLine cause Predef.ReadLine has been deprecated (#547)
@zsdz sorry for the very slow followup. I would definitely consider merging this if you changed all the files and also cleaned up the commit to have a better commit summary like "[scala] use StdIn.ReadLine instead of deprecated Predef.readLine" and if you can, add a link in the commit body to a notice of this being deprecated.

PR #549, opened 2021-02-09
fix: got syntax error when string is bigger than 500 chars

rosineygp:
When value of str or readline is bigger than 500 chars the parser failed.

```
 >_ ./mal 
Mal [bash]
user> (str "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
Error: "expected ')', got EOF"
```

### DIFF
--- a/impls/bash/reader.sh
+++ b/impls/bash/reader.sh
@@ -111,7 +111,7 @@ TOKENIZE () {
     local datalen=${#data}
     local idx=0
     local chunk=0
-    local chunksz=500
+    local chunksz=$(( datalen>500 ? datalen : 500 ))
     local token=
     local str=
 
--- a/impls/tests/step0_repl.mal
+++ b/impls/tests/step0_repl.mal
@@ -15,6 +15,9 @@ hello mal world
 hello world abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ 0123456789 (;:() []{}"'* ;:() []{}"'* ;:() []{}"'*)
 ;=>hello world abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ 0123456789 (;:() []{}"'* ;:() []{}"'* ;:() []{}"'*)
 
+;; over 500 characters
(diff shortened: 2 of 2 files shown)


kanaka:
@rosineygp apologies for the slow followup. That's a good catch. Can you actually you add a test case for this? It's possible bash might not be the only case with a problem. Also, can you rebase your commit onto master and squash your commits to clean things up a bit? Thanks.

rosineygp:
I just put the test for all impls now we got the same bug at other languages (sad but true).

PS: I don't know if I put the test in right location

dubek:
Thanks @rosineygp .   The CI failures are in:

1. basic: basic_MODE=cbm and qbasic: I see a limit `IF LEN(R$)<255` . Not sure if this can be arbitrarily increased because this impl has severe memory restrictions.
2. forth: defines `create buff 128 allot` and `buff 128 stdin read-line throw`; seems easy to increase the size.
3. prolog: the CI error is `make: *** No rule to make target 'test^prolog'.  Stop.` .  I assume this PR needs to be rebased on top of most current master branch.
4. wasm (all modes): it defines `(LET $line (STATIC_ARRAY 201))` and `(global.set $printf_buf (STATIC_ARRAY 256))` which need to be increased.

So maybe to accommodate basic, we can change the largest input string to be 254 chars? The bash fix is still good -- implementations that can handle longer strings should; but our official tests and self-hosting Mal code should all be under 254 chars. Maybe we can even state this limit in the guide (or the 254-char test string in step0 should be enough). Then we just need to fix wasm and forth.

PR #533, opened 2020-10-09
q impl

alimpfard:
Still has some work left, but self-hosting does work now.

- [x] unordered hashmap comparison (this is an unused optional, so I'm gonna pass on implementing it)
- [x] stepA optionals
- [x] a docker thingy (somehow...)
- [x] anything else that pops up

PR #554, opened 2021-04-05
fix typo in guide.md step 7

fabjan:
I'm pretty sure this is a typo at least.

### DIFF
--- a/process/guide.md
+++ b/process/guide.md
@@ -1261,7 +1261,7 @@ macros.
   return a list containing:
   the "vec" symbol,
   then the result of processing `ast` as if it were a list not
-  starting with `quote`.
+  starting with `unquote`.
 
 <a name="step8"></a>
 


alimpfard:
I noticed it too https://github.com/kanaka/mal/pull/533/commits/c3781c2438ddb15cf4cd401a70e6014c2a584de2, it's definitely a typo.

wasamasa:
I'm not 100% sure about it since there's been many changes to quasiquoting and I haven't looked into them yet. @asarhaddon Could you please take a quick look?

fabjan:
Here is why I think it's a typo:

1. This paragraph is about adding support for vectors, and with "unquote", it reflects how lists work.
2. Other implementations seem to treat vectors and lists without "unquote" the same (I looked art C, Clojure and OCaml).
3. My SML implementation failed tests until I used the same code as for lists not starting with "unquote".

but it would of course be great with input from someone who knows for sure.

asarhaddon:
You are right, this should be "a list not starting with `unquote`" instead of "a list not starting with `quote`".

kanaka:
Sorry for the slow reply everyone. Without actually planning to do this, I basically ended up taking a break from open-source/mal/free-time work for the past year. I'll blame it on the pandemic. I had a paid job shift recently so I may have some more time in the next few weeks to catch up on some mal stuff that I've been putting off for a while.

Issue #560, opened 2021-05-17
Reader macro tests could be more rigourous


wgmyers:
I had a go at implementing Mal in Ruby and somehow managed to get all the way to completion of Step A - including passing all the self-hosting tests - with a huge bug in the reader macro expansion code.

This only came to light when `make "perf^foo"` failed.

Turned out that the 'time' function was being truncated after the first `let*` in the quasiquote. Replacing the backtick with the quasiquote form typed out longhand showed me where the bug was.

I'm struggling to figure out what the best way to test this is because this is also my first exposure to any kind of Lisp, with which I am also struggling tbh, but something  like:

```
;; Reader Macro Expansion Test
(defmacro! expansiontest
  (fn* ()
    (let* [sym (symbol "s")]
      `(let* (~sym ("bol"))
         (do
           (println "sym: " ~sym))))))
```

would have caught this bug - the above should produce:

```
sym:  (bol)
```

but with incorrect quasiquote expansion, gets nothing at all, or just blows up.

kanaka:
That's a good catch. Can you send a pull request with a test addition to step7 tests that covers this? Then I'll activate Github CI against the PR to make sure that all the other implementations also pass and merge it if so (and if they don't pass then we'll work on fixing those implementations first).

wgmyers:
Just working on this now, and have hit an 'I don't really know what I am doing' problem.

In order to use the test above, I have put the test addition into step8 rather than step7, so I can use defmacro!.

The test has been modified slightly - the macro is as above but collapsed to one line (test suite doesn't seem to like multi-line expressions); also the spurious space at the end of the first arg to println has been removed, and looks like this:

```
;; Testing reader macro expansion in complex expressions
(defmacro! expansiontest (fn* () (let* [sym (symbol "s")] `(let* (~sym ("bol")) (do (println "sym:" ~sym))))))
(expansiontest)
;=>sym: (bol)
```

This passes fine in my own ruby-wgm implementation, but seems to fail for a variety of reasons wherever else I test it (ruby, c, rust, perl, go, python).

This causes me to suspect that my own implementation is doing the wrong thing. If so, I can't work out why. 

Bit stumped.

Not sure whether to go ahead and submit a pull request anyway just yet: please advise.

asarhaddon:
Hi.
The `ada.2` implementation answers:
```
[...]

PR #564, opened 2021-05-29
Reader Macro Expansion Test Step 8

wgmyers:
Pull request added per issue #560.

Test has been added to step 8 rather than step 7 in order to be able to use defmacro!.

My first pull request here: apologies in advance if I've done anything wrong: let me know and I'll try and fix it.

### DIFF
--- a/impls/tests/step8_macros.mal
+++ b/impls/tests/step8_macros.mal
@@ -43,6 +43,11 @@
 `(1)
 ;=>(1)
 
+;; Testing reader macro expansion in complex expressions
+(defmacro! expansiontest (fn* () (let* [sym (symbol "s")] `(let* (~sym (list "bol")) (do (println "sym:" ~sym))))))
+(expansiontest)
+;=>sym: (bol)
+
 ;>>> deferrable=True
 ;;
 ;; -------- Deferrable Functionality --------


asarhaddon:
Testing the reader in step1 is easyer because no evaluation is involved yet.
 ```
`(a (b) c)
;=> (quasiquote (a (b) c))
```
The faulty reader (in the commit before the fix referenced in issue #560) answers `(quasiquote (a (b)) c)`.  The same probably holds for other reader macros.

wgmyers:
I don't know what the criteria are for adding tests, so I'm not sure how to answer here.

It's absolutely true that the suggested step 1 test above would catch the exact implementation bug that I managed to perpetrate - my code was indeed over-aggressively closing brackets before it should have done, hence managing to turn \`(a (b) c) into `(quasiquote (a (b)) c)` rather than `(quasiquote (a (b) c))`. However,  this suggested test would only catch that and not other potential subtle misimplementations.

The test I have suggested in this PR would catch a wider range of potential implementation bugs: rather than testing for a specific implementation error it instead tests for being able to successfully decompose a complex expression with multiple instances of reader macros in the correct way.

Either would have saved me from getting to the end of step A without being able to run `make "perf^mal"`, and I can see arguments for both tests being used: the step 1 test suggested above tests specifically for the exact bug we have found in the wild as perpetrated by me; the step 8 test in my PR tests for the whole class of such bugs, or at least some hopefully large subset of it.

asarhaddon:
Investigating macro and quasiquote expansion is difficult.  I will not discuss the large test, but the short test should definitely be included in step1 for each reader macro.

PR #569, opened 2021-06-03
Move step9 test of map as exceptions from deferrable to optional

asarhaddon:
Self-hosting is not requiring this anymore.


kanaka:
That's an obvious improvement. Thanks!

PR #570, opened 2021-06-03
Print a stack trace for uncaught exceptions in self-hosting

asarhaddon:
An uncaught exception in the self-hosting implementation causes the
interpreter to display the EVALed ast for each EVAL level until the
exception, something similare to a stack trace.  This does not cost
much and may help a lot debugging the self-hosting step.

On the other hand, remove try* from step0 where it serves no useful
purpose.

Because of indentation level, this diff is better viewed with the git
-b command line option.

### DIFF
--- a/impls/mal/step0_repl.mal
+++ b/impls/mal/step0_repl.mal
@@ -18,10 +18,7 @@
   (if line
     (do
       (if (not (= "" line))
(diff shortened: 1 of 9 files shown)


kanaka:
That's a great improvement for very little cost.

PR #573, opened 2021-06-20
Add PureScript implementation

mrsekut:
This is a PureScript implementaion.
The test of step5 has not passed (implementing..)
Dockerfile etc. are not ready yet.


kanaka:
@mrsekut any progress on step5? I'm going to hold off on reviewing it until that is working since it could lead to refactoring of other steps.

mrsekut:
I'm sorry, please wait a little longer.

I learned that PureScript needs special measures to mitigate stack overflow, so I will implement it.

Also, the test in step A has failed, and that problem has been fixed, but I haven't pushed it yet. After fixing step 5, let me talk about how to fix the commit log.

mrsekut:
@kanaka 
It took a while, but I was able to fix it. The content of the correction was large, so I recreated the PR.

PR #578, opened 2021-07-18
python: various simplifications

asarhaddon:
python_MODE=python3 crashes on impls/python/tests/step5_tco.mal, but
it also does on the current head.

Replace some lambda constructs with more idiomatic generator
expressions, removing several temporary sequences (conversions,
contatenations, slices, arguments).

Remove obsolete coll_Q test and preservation of metadata by conj.

Use normal classes for functions in order to simplify cloning and
membership tests.

Close file open by 'slurp' after reading.

Derive List and Vector from tuple, which is more efficient than list
for immutable structures.  Remove the need for __getitem__ stuff.

Search in outer environments with a loop instead of a recursion.

Replace most python2/3 wrappers with code compatible with both
versions.  Use a non-printable ASCII characters instead of Unicode
characters (in order to avoid unicode in python2).

Follow Python style conventions:
* int(a/b)      -> a//b or floordiv(a, b)
* obj.__str__() -> str(obj)
* obj.__meta__  -> obj._meta
* f (args)      -> f(args)

Remove obviously backported stuff from first steps.

Let try* benefit from TCO.

Fix the detection of special forms (strings were accepted as well as
symbols).

Arguably improve readability:
* merge is_macro_call into macroexpand
* merge eval_ast into EVAL
(The two 'if not types._list_Q(ast)' tests in EVAL were redundant)

### DIFF
--- a/impls/python/step0_repl.py
+++ b/impls/python/step0_repl.py
@@ -6,8 +6,7 @@ def READ(str):
     return str
 
 # eval
-def EVAL(ast, env):
-        #print("EVAL %s" % printer._pr_str(ast))
+def EVAL(ast):
         return ast
 
 # print
@@ -16,14 +15,15 @@ def PRINT(exp):
 
 # repl
 def REP(str):
-    return PRINT(EVAL(READ(str), {}))
+    return PRINT(EVAL(READ(str)))
 
 # repl loop
 while True:
     try:
         line = mal_readline.readline("user> ")
-        if line == None: break
-        if line == "": continue
         print(REP(line))
-    except Exception as e:
+    except EOFError:
+        print()
+        break
+    except Exception:
         print("".join(traceback.format_exception(*sys.exc_info())))
--- a/impls/python/step2_eval.py
+++ b/impls/python/step2_eval.py
@@ -4,8 +4,7 @@ import mal_types as types
 import reader, printer
 
 # read
-def READ(str):
-    return reader.read_str(str)
+READ = reader.read_str
 
 # eval
 def EVAL(ast, env):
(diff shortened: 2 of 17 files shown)


impls/python/step0_repl.py
```diff
@@ -6,17 +6,16 @@ def READ(str):
     return str
 
 # eval
-def EVAL(ast, env):
-        #print("EVAL %s" % printer._pr_str(ast))
-        return ast
+def EVAL(ast):
+    return ast
```
dubek, line 10: So here indeed `env` is not used, but it's part of the evolution of the program (you'll see that in all impls). In step 0, 1 it's not used and then in later steps it is used (and later it becomes an `Env` object instead of a dict).

impls/python/step2_eval.py
```diff
@@ -4,53 +4,41 @@
 import reader, printer
 
 # read
-def READ(str):
-    return reader.read_str(str)
+READ = reader.read_str
 
 # eval
-def eval_ast(ast, env):
+def EVAL(ast, env):
```
dubek, line 10: Same thing: the guide explains you need a function called `eval_str` which does X,Y,Z, and another function called `EVAL` which does A,B,C . This modification steps away from this guide.

impls/python/step0_repl.py
asarhaddon: In `process/guide.md`, the `env` parameter explicitly appears in step2.  Several implementations, including the graphs in the same subdirectory, are not consistent, probably because of copy-paste between steps.

impls/python/step2_eval.py
asarhaddon: In the guide, `eval-ast` serves two very different purposes:
* to shorten the main conditional in `eval`.
* as a helper for  the apply phase and the `do` special form.

Mxing the two does not improve the learning experience. As a beginner discovering step2, I have had the impression that `eval-ast` describes the evaluation of lists as it does for symbols, vectors and maps.
I have suggested to remove `eval-ast` from the guide, but kanaka prefers this structure.
Note however that implementations are allowed to diverge here.  For example, several languages with pattern matching already do, as well as some exotic languages defining `eval` as an overridden object method.
In this merge request, the change is motivated because `do` and the apply phase can manipulate a python generator.. The current structure is the only reason why they construct an intermediate MAL list in memory.

kanaka:
Good improvements. Thanks!

PR #580, opened 2021-07-21
additional test

Garklein:
This was a bug that took me a while to find in the self-hosting step. In my case, it had to do with javascript's object copying system.


dubek:
The CI tests pass except `IMPL=elm`, which fails during `npm install`, so not related to this change.

kanaka:
@Garklein I've merged, but if you have a chance, it would be good to check the return value of the final call.

PR #583, opened 2021-08-05
Added expected result for last test

Garklein:


Garklein:
I don't think I did the pull request right but I don't know how to fix it.

PR #589, opened 2021-09-10
pprint properly prints hash-maps

jig:
Current implementation of **pprint** is not compatible with implementations using languages that do not guarantee that **keys** on a _hash-map_ keep ordering (e.g. Go language).

This PR corrects this.

### DIFF
--- a/impls/lib/pprint.mal
+++ b/impls/lib/pprint.mal
@@ -28,7 +28,7 @@
                                       (pp- k kindent)
                                       " "
                                       (pp- (get obj k) vindent)))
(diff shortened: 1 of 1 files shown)


jig:
With previous code, the effect was that some keys appear duplicated and others do not appear at all when printing a _hash-map_ with **pprint**. 

This fix extracts the **keys** list only once, so _hash-map_ keys order indetermination is not an issue.

Validated with Go language.

kanaka:
@jig Sorry for the really slow response. The change looks good and I have merged it. Thanks!

PR #599, opened 2021-12-11
Ruby.2

cookrn:
I'm not sure if you consider accepting additional implementations for already completed languages, but I recently finished one in Ruby that passes all the tests. I definitely couldn't have figured it all out without reviewing some of the tricky details of the current implementation! It was very fun to work through :sunglasses: 

A few aspects of this implementation as compares to the included one in Ruby:

- No global variables
- 2 space tabs
- No modification (monkey-patching) of core ruby classes
- All code contained within `Mal` namespace

This certainly isn't a criticism, but these are some techniques I used to try to make the implementation more idiomatic if one were to compare to other Ruby code in the wild. If you're open to it, I'd love to polish this more to help get it merged. Cheers!

### DIFF
--- a/IMPLS.yml
+++ b/IMPLS.yml
@@ -81,6 +81,7 @@ IMPL:
   - {IMPL: rexx}
   - {IMPL: rpython, SLOW: 1}
   - {IMPL: ruby}
+  - {IMPL: ruby.2}
   - {IMPL: rust}
   - {IMPL: scala}
   - {IMPL: scheme, scheme_MODE: chibi}
--- a/Makefile.impls
+++ b/Makefile.impls
@@ -38,7 +38,7 @@ IMPLS = ada ada.2 awk bash basic bbc-basic c c.2 chuck clojure coffee common-lis
 	elisp elixir elm erlang es6 factor fantom fennel forth fsharp go groovy gnu-smalltalk \
 	guile haskell haxe hy io janet java java-truffle js jq julia kotlin livescript logo lua make mal \
 	matlab miniMAL nasm nim objc objpascal ocaml perl perl6 php picolisp pike plpgsql \
-	plsql powershell prolog ps python python.2 r racket rexx rpython ruby rust scala scheme skew sml \
+	plsql powershell prolog ps python python.2 r racket rexx rpython ruby ruby.2 rust scala scheme skew sml \
 	swift swift3 swift4 swift5 tcl ts vala vb vhdl vimscript wasm wren yorick xslt zig
 
 step5_EXCLUDES += bash        # never completes at 10,000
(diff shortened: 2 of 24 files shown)


kanaka:
I'm willing to merge an additional ruby implementation (especially since it's fairly different). This was basically my second program using ruby so I'm sure it's not particularly idiomatic. One of the things I really enjoyed about ruby is that it seemed very "Lispy" under the hood so I tried to leverage that as much as possible. The ruby implementation is one of the smaller implementations as a result.

I think my only comment about the code itself is that `core.rb` seems fairly verbose. I think some of that comes from the fact that you have user-defined types for symbols, true, false, and nil. Wouldn't it still be fairly idiomatic to use ruby's built-ins for that? Not a blocker for merging, it just stood out to me as really large compared to my implementation.

For me to merge it, you'll need to get the CI stuff working. You'll need a stub Dockerfile and Makefile. The following commands need to work:

```
make "docker-build^ruby.2"
make DOCKERIZE=1 "test^ruby.2"
```

cookrn:
Great! I'll work on getting the CI build to pass.

Regarding the distinct types for `true`, `false`, `nil`, and symbols, I found that as I was working through the steps it was helpful to know in an obvious and visible way when I was using the types for something related to Mal interpretation separate from internal Ruby logic. Also, maybe an optimization too far, but it seemed like it could be helpful in a running system using Ruby and Mal w/ interop to know which data was coming from which language.

Anyways, I'm very open to changing to using the Ruby types for those value explicitly if it would be better for all current and future Mal users, so just let me know if it's important to change that in order to be merged :smiley:. A different way to improve the readability of `core.rb` could be to divide it into more files in a `core/` directory based on the function intent.

cookrn:
@kanaka is each implementation pushed as its own image to Dockerhub prior to running in CI? Apologies, I don't quite understand how to get the Docker image to be accessible during the build.

kanaka:
@cookrn I forgot to actually pull the trigger on pushing out the docker image. I've done that now and restarted the CI.

cookrn:
@kanaka no problem, looks like it passed :fireworks: 

Do you know whether it's expected that the other builds are failing?

kanaka:
Okay, tests look good. The elm and wasm failures can be ignored. The only thing left is to add your implementation to the README with a brief description of the differences from the first ruby implementation and then bump the implementations and runtime modes count values at the top of the README. Once you do that I'll merge it.

kanaka:
And don't forget to fix the FROM line of your Dockerfile to use ubuntu:21.04 (otherwise the next time I rebuild and push, it will unnecessarily bump the version and might cause subtle issues that are tough to track down).

cookrn:
@kanaka I just realized I forgot to complete the interop. Would it be best to finish that now before merging? Or update later if/when that's complete?

README change should be complete now.

kanaka:
Interop is completely optional so I went ahead and merged it. You can add a new PR when you have interop ready.

Congrats and thanks!

cookrn:
Awesome, thanks for merging!

PR #602, opened 2021-12-24
anchor hyperlink fix

The-Alchemist:


dubek:
Looks good. @The-Alchemist - thanks for the fix!

PR #605, opened 2022-01-26
factor: update for Factor 0.98

mrjbq7:
Updating syntax for small changes in Factor 0.98.

### DIFF
--- a/impls/factor/Dockerfile
+++ b/impls/factor/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:vivid
+FROM ubuntu:20.04
 MAINTAINER Joel Martin <github@martintribe.org>
 
 ##########################################################
@@ -24,7 +24,9 @@ WORKDIR /mal
 # Factor
 RUN apt-get -y install libgtkglext1
 RUN cd /usr/lib/x86_64-linux-gnu/ \
-    && curl -O http://downloads.factorcode.org/releases/0.98/factor-linux-x86-64-0.98.tar.gz \
+    && curl -O https://downloads.factorcode.org/releases/0.98/factor-linux-x86-64-0.98.tar.gz \
     && tar xvzf factor-linux-x86-64-0.98.tar.gz \
     && ln -sf /usr/lib/x86_64-linux-gnu/factor/factor /usr/bin/factor \
     && rm factor-linux-x86-64-0.98.tar.gz
+
+ENV HOME /mal
--- a/impls/factor/Dockerfile
+++ b/impls/factor/Dockerfile
@@ -24,8 +24,7 @@ WORKDIR /mal
 # Factor
 RUN apt-get -y install libgtkglext1
 RUN cd /usr/lib/x86_64-linux-gnu/ \
-    && curl -O http://downloads.factorcode.org/releases/0.97/factor-linux-x86-64-0.97.tar.gz \
-    && tar xvzf factor-linux-x86-64-0.97.tar.gz \
+    && curl -O http://downloads.factorcode.org/releases/0.98/factor-linux-x86-64-0.98.tar.gz \
+    && tar xvzf factor-linux-x86-64-0.98.tar.gz \
     && ln -sf /usr/lib/x86_64-linux-gnu/factor/factor /usr/bin/factor \
-    && rm factor-linux-x86-64-0.97.tar.gz
-
+    && rm factor-linux-x86-64-0.98.tar.gz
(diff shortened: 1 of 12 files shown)


mrjbq7:
The Factor 0.98 release was in 2018.  I expect a Factor 0.99 release in 2022.

Both should work with this slight change.

impls/factor/Dockerfile
```diff
@@ -24,8 +24,7 @@ WORKDIR /mal
 # Factor
 RUN apt-get -y install libgtkglext1
 RUN cd /usr/lib/x86_64-linux-gnu/ \
-    && curl -O http://downloads.factorcode.org/releases/0.97/factor-linux-x86-64-0.97.tar.gz \
-    && tar xvzf factor-linux-x86-64-0.97.tar.gz \
+    && curl -O http://downloads.factorcode.org/releases/0.98/factor-linux-x86-64-0.98.tar.gz \
```
dubek, line 27: Please change the URL from `http://` to `https://` .

impls/factor/Dockerfile
```diff
@@ -24,8 +24,7 @@ WORKDIR /mal
...
+    && tar xvzf factor-linux-x86-64-0.98.tar.gz \
     && ln -sf /usr/lib/x86_64-linux-gnu/factor/factor /usr/bin/factor \
-    && rm factor-linux-x86-64-0.97.tar.gz
-
+    && rm factor-linux-x86-64-0.98.tar.gz
```
dubek, line 30: Please add the following line at the end of Dockefile:

```

ENV HOME /mal
```

This will allow factor to create the `$HOME/.cache` directory.

impls/factor/Dockerfile
```diff
@@ -24,8 +24,7 @@ WORKDIR /mal
 # Factor
```
dubek, line 24: Please modify the top of the Dockerfile: change `FROM ubuntu:vivid` to `FROM ubuntu:20.04` (just to modernize the images).

impls/factor/step7_quote/step7_quote.factor
```diff
@@ -11,7 +11,7 @@ SYMBOL: repl-env
 
 DEFER: EVAL
 
-GENERIC# eval-ast 1 ( ast env -- ast )
+GENERIC#: eval-ast 1 ( ast env -- ast )
```
dubek, line 14: besides this, there's another GENERIC in line 52 which needs fixing (missing colon). I get an error:

```
Started with:
../factor/step7_quote/step7_quote.factor

52: GENERIC# apply 0 ( args fn -- maltype newenv/f )
            ^
No word named “GENERIC#” found in current vocabulary search path
(U) Quotation: [ c-to-factor
Testing cons function
TEST: '(cons 1 (list))' -> ['',(1)]
Exception: IOError(5, 'Input/output error')
Output before exception:
]
    Word: c-to-factor
(U) Quotation: [ [ (get-catchstack) push ] dip call => (get-catchstack) pop* ]
(O) Word: command-line-startup
(O) Word: run-script
(O) Word: run-file
(O) Word: parse-file
(O) Word: parse-stream
(O) Word: parse-fresh
(O) Word: (parse-lines)
(O) Word: (parse-until)
(O) Word: parse-until-step
(O) Word: no-word
(O) Word: throw-restarts
(O) Method: M\ object throw
(U) Quotation: [
        OBJ-CURRENT-THREAD special-object error-thread set-global
        current-continuation => error-continuation set-global
        [ original-error set-global ] [ rethrow ] bi
    ]
(cons 1 (list))

make: *** [Makefile:238: test^factor^step7] Error 1
```

(similarly for steps 8, 9, A)

dubek (changes requested):
Thanks @mrjbq7 for your contribution to Mal.

I ran into several problems when testing this PR locally, see comments below:

mrjbq7:
Hi @dubek, thank you for the review.

I believe I have fixed all of those issues, but the "Build and Test / linux" Github Action is still using old Factor, which is why it is erroring.

dubek (approved):
Thanks @mrjbq7 , now everything works for me locally (with your updated Dockerfile).

In order for it to work correctly, we need to wait for @kanaka to build and push the `kanaka/mal-test-factor` docker image from your new Dockerfile, and then re-run the "Build and Test" job.

asarhaddon:
Hello. Just for the record, this duplicates some parts of #588

mrjbq7:
The Dockerfile in #588 looks pretty good, the only thing it's missing from this is ``ENV HOME /mal``.

I'd be happy to be added as a maintainer of this if it's useful.

I am one of the Factor core developers.

impls/factor/Dockerfile
Gam3356: Okay!

PR #606, opened 2022-02-01
Intrusive changes in the make implementation

asarhaddon:
Please find details are in the commit header.


dubek:
Have you noticed this PR includes changes to make but also to ruby.2, vb, python.2? Is that intentional?

asarhaddon:
Not exactly. This work is based on #592, but only needs the Make parts. I have cherry-picked the interesting commits. Thanks.

asarhaddon:
The only remaining failure is `elm`, which is completely unrelated (#608 ).

PR #607, opened 2022-02-01
mal: error reporting, procedural env_set, native macro?

asarhaddon:
Hello.
This merge request contains 3 commits implementing three ideas, and a backport from #592.  The suggestions are interconnected, so a single merge request is convenient for now to start discussion and list implementations for which `apply` refuses macros.

### DIFF
--- a/impls/ada.2/core.adb
+++ b/impls/ada.2/core.adb
@@ -113,10 +113,10 @@ package body Core is
                return F.Builtin.all (A);
             when Kind_Builtin_With_Meta =>
                return F.Builtin_With_Meta.all.Builtin.all (A);
-            when Kind_Fn =>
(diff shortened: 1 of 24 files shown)


asarhaddon:
elm fails because the Docker image needs to be updated (#608), but Types.elm suggests that Core.elm/apply accepts macros.
Implementations for which core/apply refuses macro arguments:
erlang fsharp haskell nasm prolog sml vimscript
#592 should probably be completed first in order to avoid non-trivial rebase work.

kanaka:
Looks good. Definitely good to have (apply macro ...) working for all implementations. Regarding the mal error checking, I've been torn between adding those or keeping the mal implementation as concise as possible. There are pedagogical reasons for both approaches. But I think I prefer error checking now.
Thanks!

PR #608, opened 2022-02-07
Update elm implementation, fix keywords and metadata, merge eval

asarhaddon:
Hello.
The first commit updates the elm implementation for 0.19.1. The docker file describes my configuration but should probably be checked on an actual image.
The two following commits fix the last failing optional tests.
The two last commits should be part of #592, but required 0.19.1 for testing.


asarhaddon:
The tests were passing on my machine, so they will probably also pass in an updated Docker image.

asarhaddon:
Hello.
For some reason, the installation of the tzdata package interactively asks the geographical area, so the actual tests do not even start.  I have no idea how to debug this.

kanaka:
Yeah, seems like it's a behavior that started midstream in an Ubuntu release. The trick is this:
```
RUN DEBIAN_FRONTEND=noninteractive apt-get -y ...
```

kanaka:
Update with fixed the Dockerfile issue and a Makefile bug (probably search/replace issue). Merged these commits from my own branch so closing this.

PR #611, opened 2022-03-12
Fixes for guide text

hcs64:


kanaka:
@hcs64 Sorry for the super long response time. These are great improvements. I had to do a bit of conflict resolution since things have changed in the meantime, but this is merged now. Thanks!

PR #623, opened 2022-12-21
ran cargo clippy --fix on rust implementation to clean up the code a bit

BartMassey:
The Rust implementation currently has compiler warnings. I used the big hammer of `cargo clippy --fix`, which seems to have resolved these and a host of other Clippy linter warnings. I didn't find any tests to run beyond the examples, but everything seems to still work as intended.


kanaka:
Sorry for the super slow reply. I think all the warning has been resolved by other rust implementation updates here: 253cdf1ae5aa32e59d8d964cc2cad994f2fd8c99

PR #625, opened 2023-02-16
V: new implementation

edam:
- added new implementation in v (http://vlang.io)
- self hosts 👍

### DIFF
--- a/IMPLS.yml
+++ b/IMPLS.yml
@@ -98,6 +98,7 @@ IMPL:
   - {IMPL: sml, sml_MODE: mosml}
   - {IMPL: tcl}
   - {IMPL: ts}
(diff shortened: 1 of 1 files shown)


kanaka (changes requested):
Hello @edam sorry for the very slow reply (I've been unable to work on this project for the past couple of years but I'm am able to again). This would be great to get merged. If you're still interested I would ask that you do the following:

* The recommended process has changed slightly: the eval_ast is no longer seperate from eval. It would be good if you could update to follow the new structure.
* Make sure that the Github CI tests all pass for your implementation. The process is now more automated and doesn't require my manual intervention to build docker images (part of the reason to rebase).

Thanks!

PR #626, opened 2023-02-27
Add new Perl (perl.2) implementation

ingydotnet:
I'm working on a YAML syntax Lisp called [YAMLScript](https://github.com/yaml/yamlscript) that will work in many languages.

I realized I really needed to learn how to implement a Lisp.

I found this project and completed the adventure.

This PR is for the perl.2 base implementation.

I'll be submitting a yamlscript implementation PR soon.

PR #627, opened 2023-02-27
Fix ci.sh to build new impl images

ingydotnet:
From PR #626 (new impl).

In GHA, run local docker build if kanaka/... image not available.


kanaka:
@ingydotnet Thanks for working on this. I implemented something similar but with a bit more functionality (including caching/pushing to ghcr.io). And then today as I was going through cleaning up PRs I discovered that you had done part of this already.

PR #628, opened 2023-03-06
Javascript: fix reader, remove base argument to parseFloat

zoren:
parseFloat does not take a base parameter


kanaka:
Sure enough! Not sure why I thought it did.

PR #630, opened 2023-03-31
Second Lua implementation (Lua.2)

cheminDeFer:
Major difference from first Lua implementation is tokenisation is done manually instead of using regular expressions.


kanaka (changes requested):
Hello @cheminDeFer sorry for the very slow reply (I've been unable to work on this project for the past couple of years but I'm am now able to again). If you're still interested in getting this merged upstream I would ask that you do the following:

* The recommended process has changed slightly: the eval_ast is no longer seperate from eval and macroexpand is not longer a standalone function. It would be good if you could update to follow the new structure.
* Add your implementation to `IMPLS.yml` and make sure that the Github Actions CI tests all pass for your implementation.

PR #632, opened 2023-04-21
Ruby fixes

asarhaddon:
Hello.
Here are two trivial test fixes for ruby.
'do' is also broken, but #592 fixes this as a side effect and I do not want to create rebase work.


asarhaddon:
Hello.
The image is probably too old for the pattern matching constructs.  They work with ruby 3.1.2p20.

kanaka:
I update the Dockerfile on a local branch and verified this works. Merged on that branch.

PR #635, opened 2023-06-07
Adding notes to run this in 2023

Luv2C0d3:
Trial and error got me to use lts/erbium as the last lts release where this code would pass all tests.


kanaka:
I think JS implementation should work with latest version of node. Also, the Dockerfile is now updated to use Node v18 (from Ubuntu 24.04).

PR #639, opened 2023-07-06
New LaTeX3 implementation

asarhaddon:
It requires two cosmetic changes to existing tests.

### DIFF
--- a/IMPLS.yml
+++ b/IMPLS.yml
@@ -48,6 +48,7 @@ IMPL:
   - {IMPL: js}
   - {IMPL: julia}
   - {IMPL: kotlin}
+  - {IMPL: latex3, NO_PERF: 1, NO_SELF_HOST: 1, SLOW: 1}
   - {IMPL: livescript}
   - {IMPL: logo}
   - {IMPL: lua}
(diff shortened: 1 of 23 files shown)


kanaka (changes requested):
Apart from the time-ms test change (comment inline), looks really good. You'll also need to add it to `IMPLS.yml` for CI to pick it up.

kanaka (changes requested):
Looks like it failed during image build due to the tzdata interactivity issue. This should fix it:

```
-RUN apt-get -y install texlive-latex-base
+RUN DEBIAN_FRONTEND=noninteractive apt-get -y install texlive-latex-base
```

asarhaddon:
I cannot remember exactly, but I spent at least a week in cumulated time.
I was expecting a challenge. LaTeX3 is an effort to make TeX/LaTeX as consistent as possible, despite its original design (Knuth did explicitly not want TeX to be Turing-complete).  The recommend way to implement LaTeX extensions is to use the embedded LUA interpreter.
That said, the LaTeX3 documentation is wonderful and clearly describes its purpose and limitations.
I cannot reproduce the issue locally. On my machine the `texlive-latex-base` package is at version 2024.20240706-1, while the version in the build logs is 2019-*. I will try to require ubuntu:24.04 instead of 20.04 in the Dockerfile.

kanaka:
Guess the latex in 20.04 was just too old.

Issue #646, opened 2023-09-22
Floats don't work in the python implementation.


vipenzo:
In the python implementation float numbers are not accepted by the reader.
There is a bug in reader.py at line 34 (read_atom function)

It is now:
   elif re.match(float_re, token): return int(token)
should be:
   elif re.match(float_re, token): return float(token)

jero98772:
Hello , solve your problem , idk if my pull request will be accepted but here is how i solve it:

https://github.com/kanaka/mal/pull/654

i just change the line 34 to

 elif re.match(float_re, token): return float(token)

kanaka:
This should be resolved now. Thanks for the catch!

PR #654, opened 2024-06-26
solve issue Floats don't work in the python implementation. #646

jero98772:
the error was because it is not converted to float change the line 34
from
    elif re.match(float_re, token): return **_int_**(token)
to 
    elif re.match(float_re, token): return **_float_**(token)


kanaka:
Thanks for catching that!

PR #661, opened 2024-08-07
Wasm: merge EVAL_AST into EVAL

asarhaddon:
Also prevent macros from mutating functions and implement DEBUG-EVAL.


kanaka:
I've pushed some changes to remove lucet (no longer supported, merged into wasmtime) and updated the docker image. If you rebase you should at least get past the image building issues.

PR #664, opened 2024-08-22
Lua: merge evalAST and fix remaining tests

asarhaddon:


kanaka (changes requested):
Code changes look great. I think the build failure is just that the `lua-5.3` package needs to be `lua5.3`. You might be aware already but if you have docker on your system you can run `make "docker-build^lua"` to test the docker image build.

Issue #662, opened 2024-08-08
Fix self-hosted test failures


kanaka:
It's been a while since I've done a comprehensive self-host test run. I've add a input variable to workflow so that self-hosted mode can be added for manually triggered workflows. The CI run is at:  https://github.com/kanaka/mal/actions/runs/10309343464/job/28538806985

Some of these implementations might just need a `NO_SELF_HOST` setting (such as `xslt`), but a lot of the failures look like bugs rather than timeouts or memory exhaustion.

Once all these have been fixed or `NO_SELF_HOST` is added, we should make self-hosted test step enabled by default (it only adds a few minutes overall to the full CI workflow). 

Here is the the full list failures from self-hosted testing (that don't happen for normal non-self-hosted tests):

- [x] awk
- [x] c.2
- [x] cpp
- [x] common-lisp
- [x] dart
- [x] java
- [ ] jq
- [x] kotlin
- [ ] nasm
- [x] ps
- [x] powershell
- [x] python.2
- [x] sml
- [x] vala
- [ ] xslt
- [x] swift5

kanaka:
awk fixed by https://github.com/kanaka/mal/commit/e62567fb8f0105f674e865302115780dced3618d and https://github.com/kanaka/mal/commit/d1afe8ddb549dbfea313bf856648b22fb9ef856d

kanaka:
python.2 fixed in https://github.com/kanaka/mal/commit/0d32585d6b64866b52c4c3a8b1249c7bf7abfb42

kanaka:
Added common-lisp. Now that the underlying common-lisp issue with hash-maps is fixed, it reveals a new self-hosted failure (step4 issue with how nil is handled).

kanaka:
More recent CI run (with awk and python.2 fixed), is here: https://github.com/kanaka/mal/actions/runs/10393189805/job/28780104927

Some of the failures have some common patterns:
* `c.2`, `cpp`, `dart`, `swift5` all fail in step4 with the same type of error. Appears that the result of the list command is being evaluated again when it shouldn't be. For example:
  ```
  $ make DOCKERIZE=1 MAL_IMPL=cpp repl^step4
  mal-user> (list 1)
  Uncaught exception: "1" is not applicable
  ```
*` java` and `kotlin `appear to have a similar symptom:
  ```
  $ make DOCKERIZE=1 MAL_IMPL=kotlin repl^step4
  mal-user> (do (prn 101) 7)
  101
  101
  101
  101
[...]

PR #666, opened 2024-08-24
Fix VBScript implementation self-hosting fails to start issue

OldLiu001:
By checking the [logs](https://github.com/kanaka/mal/issues/662#issuecomment-2305841001) provided by @kanaka , I realize that the problem is caused by the different path specifications in `win` and `linux`.

There was no problem before because normal tests used relative paths, while self-hosted tests used absolute paths.


kanaka:
Thanks! Merged.

PR #668, opened 2024-08-26
jq: Fix functions with empty body

dubek:
jq's `find_free_references` failed on empty lists; add handling for that case.

Add relevant test in step4.

Also remove stray `_debug` call in `nth` implementation


kanaka:
The elisp and powershell failures are know failures until https://github.com/kanaka/mal/pull/669 is merged (nearly completed). So I went ahead and merged this.

Issue #657, opened 2024-08-05
Remaining impls to combine eval-ast/macroexpand into eval


kanaka:
Most implementations have been converted via https://github.com/kanaka/mal/pull/592. The following are the remaining implementations to convert at some point:

- [ ] basic
- [ ] chuck
- [ ] crystal
- [ ] fantom
- [x] hy
- [ ] io
- [ ] janet
- [ ] jq
- [ ] livescript
- [x] lua
- [ ] matlab
- [x] objc
- [ ] plpgsql
- [ ] plsql
- [x] powershell
- [x] purs/purescript (in progress: https://github.com/kanaka/mal/pull/631)
- [ ] rexx
- [x] rust
- [ ] scala
- [ ] swift
- [ ] swift3
- [ ] swift4
- [ ] vala
- [x] wasm
- [ ] xslt
- [x] zig (will need to resolve https://github.com/kanaka/mal/pull/591 first)

kanaka:
Rust complete: https://github.com/kanaka/mal/pull/659

kanaka:
wasm completed with https://github.com/kanaka/mal/pull/661

asarhaddon:
What are the remaining issues with awk bbc-basic elixir erlang forth vhdl ? Commit 03389277 was supposed to include them.
Also, I suggest to add a commet in the zig line refering to #591.

kanaka:
@asarhaddon You're right, I did a grep for eval_ast and macroexpand and didn't look closely enough at the code or review your commit again when making this list. I've added a comment to zig.

kanaka:
purs/purescript fixed in https://github.com/kanaka/mal/pull/631

kanaka:
Powershell updated in https://github.com/kanaka/mal/pull/669

kanaka:
objc updated in https://github.com/kanaka/mal/pull/671

kanaka:
zig problem and eval_ast merge completed here: https://github.com/kanaka/mal/pull/591

kanaka:
Hy implemented here: https://github.com/kanaka/mal/pull/683

kanaka:
rexx completed: https://github.com/kanaka/mal/pull/686

kanaka:
I've added a PR to remove Swift 2 ("swift") if the original author doesn't want to get it build/testable in CI: https://github.com/kanaka/mal/pull/688

kanaka:
crystal completed: https://github.com/kanaka/mal/pull/689

[...]

PR #671, opened 2024-08-27
objc: merge eval_ast and fix some tests

asarhaddon:


asarhaddon:
This helps with #657 (by the way, lua is fixed too).

PR #673, opened 2024-08-27
mal: implement DBGEVAL without assuming that a: is a valid symbol

asarhaddon:
This does not seem worth a test, but can currently break the self-hosting.

### DIFF
--- a/impls/mal/step2_eval.mal
+++ b/impls/mal/step2_eval.mal
@@ -8,7 +8,7 @@
 
 ;; eval
 (def! EVAL (fn* [ast env]
-  ;; (do (prn 'EVAL: ast))
+  ;; (do (prn "EVAL:" ast))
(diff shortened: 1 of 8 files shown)


asarhaddon:
Sorry, I am cleaning the backlog in semi-random order.
A test for `a:` has been included meanwhile.
Do you want me to downgrade its severity? With this merged, `a:` is nice but not required for self-hosting.

kanaka:
I don't have a strong opinion. I do tend to lean towards having optional things be marked optional in the tests. That way implementors don't get hung up if those optional tests are hard to do in their target language. In this case it's probably not going to trip up implementations. Although I could be wrong. When I was implementing bash it had a bunch of these types of things that should have been simple but ended up triggering weird behaviors (like "*, "!", etc)

PR #675, opened 2024-08-28
New test for closures

asarhaddon:
I have found this in an old mess of experimentations with ada.2.  Seems simple and worth testing.

### DIFF
--- a/impls/ada.2/Dockerfile
+++ b/impls/ada.2/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:24.04
 MAINTAINER Joel Martin <github@martintribe.org>
 
 ##########################################################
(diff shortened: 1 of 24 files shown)


kanaka:
Yes, that's a really good test. It also broke nim which is unexpected.

kanaka:
I fixed nim (4c7b1d6af04931e478e848f56c843c095aa018a6). It was just step3 and 4 that had the problem which was what I guessed given that later env handling and self-hosting worked fine.

asarhaddon:
I have fixed ada.2 (disabling a misguided optimization attempt by a younger me :-1: ), added trivial changes to nim (that I was investigating too), and added a similar but more simple test.
The second test triggers an input/output error in.... zig.

kanaka:
Oh, I never actually merged it. Merged now.

Issue #679, opened 2024-09-19
latex3 time-ms seems to be overflowing now


kanaka:
In an unrelated CI workflow, latex3 is failing and it looks like the problem is that the time-ms value has gotten too high:
https://github.com/kanaka/mal/actions/runs/10947930429/job/30398541755?pr=675

kanaka:
@asarhaddon Want to take a look at this?

kanaka:
Fixed by: https://github.com/kanaka/mal/pull/681

PR #681, opened 2024-09-19
latex3: fix one more time-ms overflow

asarhaddon:


asarhaddon:
I forgot to mention that this is merge request fixes #679.

PR #688, opened 2024-10-09
swift: Remove swift 2 (no longer testable).

kanaka:
This removes the Swift 2 ("swift") implementation from the repo :-(

We haven't had the Swift 2 implementation running in CI for a while (ever since the transition from Travis CI to GHA) because there are no Github Actions runners that support it. In addition, the implementation generates the incremental step files from a single source file which makes the primary source file and the resulting build process fairly complicated compared to other implementations.

We are in the process of updating all implementations to use a simpler eval process (no separate `eval_ast`) and much simplified macro handling. The non-standard layout of this implementation along with the inability to build/test this implementation is why I've created this PR to remove it.

@keith-rollin, as the original author, I would like to give you the opportunity to weigh in. If you are able to do the work to get this building/testable in GHA/CI and also simplify the build process (i.e. split the steps into true original steps), then I would be willing to keep this implementation in the repo (at least until Swift 2 binaries are no longer available for download). I'll keep this open for a week or two so that you have the opportunity to weigh in. Let me know what you think.

---
Notes on my own attempt to get this to build in a docker container:

There are still packages for Swift 2 here: https://www.swift.org/download/ (under "Older Release"). I've attempted to build a Linux docker image using 2.2.1 based on Ubuntu 14.04. I copied the current Dockerfile image for swift3 and updated it to use ubuntu:14.04 and download 2.2.1 binary image.

I had to tweak the Makefile like this to get things attempting to build at all:
```Makefile
UNAME := $(shell uname)

ifeq ($(UNAME), Darwin)
    SWIFT   := $(shell DEVELOPER_DIR="$(DEV_DIR)" xcrun --find swiftc 2>/dev/null)
    SDKROOT := $(shell DEVELOPER_DIR="$(DEV_DIR)" xcrun --show-sdk-path 2>/dev/null)
else ifeq ($(UNAME), Linux)
    SWIFT   := $(shell which swiftc 2>/dev/null)
    SDKROOT := /
endif
```

However, even with those changes, the build fails with:
```
stepA_mal.swift:10:8: error: no such module 'Foundation'
import Foundation
       ^
```


keith-rollin:
Sure, it makes complete sense to get rid of this version. I don't think it provides any benefit. I'm amazed you've kept it around this long. :-)

kanaka:
@keith-rollin okay, done! It was one of the older implementations so I had some nostalgia for it TBH. I did add it to the "Others / Alternate implementations" section with a link to the version before this PR. Ready to do another implementation almost a decade later? Fortran is ripe for the picking :-D

PR #693, opened 2024-10-13
swift5: merge eval_ast, fix metadata, use Debian compiler

asarhaddon:
This implementation is completed, although not listed yet in #657 :-)


asarhaddon:
I wonder why Debian packages `swiftlang-dev` and Ubuntu does not.
I suggest to `echo 'deb https://deb.debian.org/debian testing main' >> /etc/apt/sources.list` before `apt update`.

kanaka:
@asarhaddon I was able to get the file to build and everything to compile and tests to pass using the following adjustment to the Dockerfile:
```
diff --git a/impls/swift5/Dockerfile b/impls/swift5/Dockerfile
index 6a8f6f0f..9112426f 100644
--- a/impls/swift5/Dockerfile
+++ b/impls/swift5/Dockerfile
@@ -6,7 +6,6 @@ MAINTAINER Joel Martin <github@martintribe.org>
 # implementations

-RUN echo 'deb https://deb.debian.org/debian trixie main' > /etc/apt/sources.list.d/swift-for-mal
 RUN apt-get -y update

 # Required for running tests
@@ -20,5 +19,8 @@ WORKDIR /mal

+RUN apt-get -y install curl
+RUN curl -s https://swiftlang.xyz/install.sh | bash
 RUN DEBIAN_FRONTEND=noninteractive apt-get -y install \
-    libc-dev swiftlang-dev
+    libc-dev swiftlang
```

However, note that this is actually swiftc version 6. We could probably just rename this implementation to Swift 6.

kanaka:
For the rpython implementation, I think there is at least a missing comma between "nil" and "false" in the DEBUG-EVAL conditional.

kanaka:
I think modifying the .travis.yml file is triggering the full build of all the implementations. I had honestly forgotten that we still had that file. We should just remove it. For now though, I would just revert it so that builds are just for swift6 and we can remove the file separately later.

kanaka:
BTW, I think you can probably fix the swift6 failure by setting HOME to /mal in the Dockerfile so that it has a place that it can write cache files to.

PR #696, opened 2024-10-14
rpython: complete the eval_ast merge, add DEBUG-EVAL

asarhaddon:


kanaka:
@asarhaddon One thing I noticed with this change is that it's 30-40% slower on the perf3 benchmark than it was before (it's notable because rpython was the clear leader before on perf3 and now it's second behind Java). It appears to be specifically the DEBUG-EVAL check. Just something interesting to note.

asarhaddon:
The cause is probably that the DEBUG-EVAL string is wrapped into a MAL symbol on each evaluation, then unwrapped repeatedly in each outer environment.  The same issue affects a lot of implementations.
I have fixed this in rpython because this should not hurt, and will confirm or infirm the diagnostic.

PR #697, opened 2024-10-14
vala: merge eval_ast into eval

asarhaddon:


dubek:
I think the commit message mentions `rpython` but should be `vala`.

PR #699, opened 2024-10-16
io: merge eval_ast into eval, add DEBUG-EVAL

asarhaddon:
The self-hosted tests fail, for apparently unrelated reasons (the first readline fails in step0). Were they failing before?

### DIFF
--- a/impls/io/step2_eval.io
+++ b/impls/io/step2_eval.io
@@ -19,6 +19,9 @@ eval_ast := method(ast, env,
 )
 
 EVAL := method(ast, env,
+
+    //  ("EVAL: " .. PRINT(ast)) println
+
     if(ast type != "MalList", return(eval_ast(ast, env)))
     if(ast isEmpty, return ast)
     el := eval_ast(ast, env)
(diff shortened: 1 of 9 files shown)


kanaka:
@asarhaddon @dubek The update to the version of io in the Dockerfile appears to have completed self-hosted tests successfully: https://github.com/kanaka/mal/actions/runs/11327008274/job/31497159120 The io repo hasn't changed from the time when @dubek did the update to when the pushed image would have been pushed (although it would have been a different instance of the build but of the same sources).

Maybe it's some sort of temperamental GC related issue? When I build and run locally with it works.

asarhaddon:
Sorry, I have no clue and miss motivation to investigate this issue.

asarhaddon:
A single comma was causing a crash reported in an unrelated module...
And now that this is solved, a self-hosting issue.
I have no clue for now, but I suggest to move this enigm from #699 to #662.

kanaka:
@asarhaddon sounds like a reasonable plan. I've added io to #662

PR #700, opened 2024-10-21
jq: Add some design/implementation documentation

alimpfard:
re https://github.com/kanaka/mal/pull/685#issuecomment-2397019088, adds a design/impl doc.

@kanaka @asarhaddon: please let me know if I missed anything or if there's anything you'd like me to go into more detail about.

### DIFF
--- a/impls/jq/docs/impl-notes.md
+++ b/impls/jq/docs/impl-notes.md
@@ -1,0 +1,58 @@
+# General Implementation Notes
+
+This document contains notes on the jq implementation, describing the deviations from the MAL specification and implementation details where necessary.
+
+## Main Deviations per Step
+
+### Step 0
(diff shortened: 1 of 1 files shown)


asarhaddon:
Thanks. This would have helped for sure.
My comment from two weeks ago in #685 gives examples of questions I had in mind when working on jq.
In order to ease possible future changes project-wide, it would be great to reduce the diff between steps, even if this implies backporting environment changes from later steps.  Applying the same idea in each language is already a repetitive task when the steps are almost identical.
The situation is quite similar for xslt.

alimpfard:
Hmm, the initial implementation actually had one environment impl, I split it up because @kanaka asked for the atoms to be split off (this is on me, putting the atoms in the Env is the easy solution because all the code that doesn't care about it can just pass it along)

I'll add a similar document for xslt, but I think it follows the same basic principles as it has the same constraints

PR #703, opened 2024-10-22
vbs: merge {macro,quasiquote}expand into EVAL, add DEBUG-EVAL

asarhaddon:
Eval_ast was already reserved to lists.

env.vbs: rewrite env.get with a string argument and without exception because this is convenient for DEBUG-EVAL.

types.vbs: after a macro expansion, evaluate with TCO.

### DIFF
--- a/impls/vbs/env.vbs
+++ b/impls/vbs/env.vbs
@@ -41,8 +41,7 @@ Class Environment
 			If TypeName(objOuter) <> "Nothing" Then
 				Set varRet = objOuter.Find(varKey)
 			Else
-				Err.Raise vbObjectError, _
-					"Environment", "'" + varKey + "' not found"
+				Set varRet = Nothing
 			End If
 		End If
 
@@ -55,7 +54,11 @@ Class Environment
 		If objEnv Is objSelf Then
 			Set varRet = objBinds(varKey)
 		Else
+		  If TypeName(objEnv) <> "Nothing" Then
 			Set varRet = objEnv.Get(varKey)
+		  Else
+			Set varRet = Nothing
+		  End If
 		End If
 		
 		Set [Get] = varRet
--- a/impls/vbs/env.vbs
+++ b/impls/vbs/env.vbs
@@ -1,66 +1,41 @@
 Option Explicit
 
 Function NewEnv(objOuter)
-	Dim varRet
-	Set varRet = New Environment
-	Set varRet.Self = varRet
-	Set varRet.Outer = objOuter
-	Set NewEnv = varRet
+	Set NewEnv = New Environment
+	Set NewEnv.Outer = objOuter
 End Function
 
 Class Environment
-	Private objOuter, objSelf
-	Private objBinds
+	Public objOuter
+	Public objBinds
 	Private Sub Class_Initialize()
 		Set objBinds = CreateObject("Scripting.Dictionary")
 		Set objOuter = Nothing
-		Set objSelf = Nothing
 	End Sub
 	
 	Public Property Set Outer(objEnv)
 		Set objOuter = objEnv
 	End Property
 
-	Public Property Get Outer()
-		Set Outer = objOuter
-	End Property
-
-	Public Property Set Self(objEnv)
-		Set objSelf = objEnv
-	End Property
-	
 	Public Sub Add(varKey, varValue)
 		Set objBinds.Item(varKey) = varValue
 	End Sub
 
-	Public Function Find(varKey)
-		Dim varRet
-		If objBinds.Exists(varKey) Then
-			Set varRet = objSelf
-		Else
-			If TypeName(objOuter) <> "Nothing" Then
-				Set varRet = objOuter.Find(varKey)
-			Else
-				Set varRet = Nothing
-			End If
-		End If
-
-		Set Find = varRet
-	End Function
-	
 	Public Function [Get](varKey)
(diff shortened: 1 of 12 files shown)


asarhaddon:
Hello.
I would like a little help with the VBS implementation. I cannot test locally, and the CI takes several minutes for a sometimes cryptic error message.

For example, the current head outputs `Object required`, see https://github.com/kanaka/mal/actions/runs/11477089378/job/31938448096?pr=703.
@kanaka or @OldLiu001, can you give me a more interesting error message and/or explanations about this issue?

By the way, the following `env.vbs` arguably improves readability, and probably improve performances (the documentation explicitly recommends against recursion, and there are a lot of environment traversals, especially with DEBUG-EVAL), but I would appreciate if someone with VBA available works on the debugging phase.
```
Option Explicit
                         
Function NewEnv(objOuter)
        NewEnv = New Environment
        NewEnv.Outer = objOuter
End Function
                 
Class Environment                 
        Private objBinds, objOuter
                                      
        Private Sub Class_Initialize()                             
                Set objBinds = CreateObject("Scripting.Dictionary")
                ' objOuter Is Nothing
        End Sub
                                         
        Public Property Set Outer(objEnv)
                Set objOuter = objEnv
        End Property

        Public Sub Add(varKey, varValue)            
                Set objBinds.Item(varKey) = varValue
        End Sub
                                     
        Public Function [Get](varKey)
                Dim objEnv = Me
                Do Until objEnv.objBinds.Exists(varKey)
                        objEnv = objEnv.objOuter            
                        If TypeName(objEnv) = "Nothing" Then
                                ' [Get] Is Nothing
                                Exit Function
                        End If
                Loop                               
                Set [Get] = objEnv.objBinds(varKey)
        End Function
End Class
```

OldLiu001:
@asarhaddon In the VBS language, `Set` is required for Object assignment, and other types (Booleans, strings, numbers, etc.) can be directly assigned (if other types use `Set` assignment, an error will be reported).
```
Set obj = New RegExp
Set obj = Nothing
bool = True
bool = False
num = 3.14
str = "String"
```
To be honest, I also think this rule is a little strange, and Microsoft's later language, VB.Net, has removed this restriction.
['Let' and 'Set' assignment statements are no longer supported](https://learn.microsoft.com/en-us/dotnet/visual-basic/misc/bc30807)

Regarding your proposal to replace recursion with iteration, I think it is a good modification, especially in a language like VBS that does not have TCO, and can also improve efficiency.

asarhaddon:
Thanks.  That may be obvious for daily users, but not from the online docs to a beginner.

The tests now pass, and take 75 minutes (60 before DEBUG-EVAL).  We will see if the env.get optimization helps.

impls/vbs/env.vbs
```diff
@@ -3,61 +3,36 @@ Option Explicit
...
 	Public Function [Get](varKey)
 		Dim objEnv, varRet
-		Set objEnv = Find(varKey)
-		If objEnv Is objSelf Then
-			Set varRet = objBinds(varKey.Value)
-		Else
-			Set varRet = objEnv.Get(varKey)
-		End If
-		
+		Set objEnv = Me
```
OldLiu001, line 23: VBS does not have the `Me` keyword.
```suggestion
		Set objEnv = objSelf
```

impls/vbs/env.vbs
```diff
@@ -3,61 +3,36 @@ Option Explicit
 Function NewEnv(objOuter)
 	Dim varRet
 	Set varRet = New Environment
-	Set varRet.Self = varRet
```
OldLiu001, line 6: Seems to be the only way to get the object itself in VBS.

OldLiu001:
You're welcome.
It seems that you are having some minor trouble with grammar, I have made some comments.

impls/vbs/env.vbs
OldLiu001: Wait a minute, I did a simple test and it seems that the `Me` keyword works, but I never saw it in the manual.

impls/vbs/env.vbs
OldLiu001: Well, `Me` keyword exist (although it's not mentioned in the manual), the problem should not be here.

OldLiu001:
@asarhaddon Try this:
```
Class Environment
	Public objOuter
	Public objBinds
	Private Sub Class_Initialize()
		Set objBinds = CreateObject("Scripting.Dictionary")
		Set objOuter = Nothing
	End Sub
	
	Public Property Set Outer(objEnv)
		Set objOuter = objEnv
	End Property

	Public Sub Add(varKey, varValue)
		Set objBinds.Item(varKey) = varValue
	End Sub

	Public Function [Get](varKey)
		Dim objEnv, varRet
		Set objEnv = Me
		Do
			If objEnv.objBinds.Exists(varKey) Then
				Set varRet = objEnv.objBinds(varKey)
				Exit Do
			End If
			Set objEnv = objEnv.objOuter
			If TypeName(objEnv) = "Nothing" Then
				Set varRet = Nothing
				Exit Do
			End If
		Loop

		Set [Get] = varRet
	End Function
End Class
```

asarhaddon:
Done! 
Removing the recursion only gets 5 minutes. Good, but that does not even compensate the introduction of DEBUG-EVAL.

kanaka:
Wow, that is very slow.

PR #705, opened 2024-10-24
plpgsql: merge macroexpand into EVAL, add DEBUG-EVAL

asarhaddon:


asarhaddon:
This merge request also adds a test that `^2` is not interpreted as a symbol. The issue is also tested in stepA, but the reader macro should be tested before the metadata system.

As far as I know, the last optional failure was already present (`throw` always converts its argument to a string). I have not managed to fix it.

By the way, someone with more SQL skills could probably simplify the Dockerfile and/or merge `wrap.sh` into `run`.
I have managed to check the syntax as a normal user with the following commands (actually running the tests require dblink, which is not packaged).
The installation of the postgresql Debian package creates the postgres user (details in `/usr/share/doc/postgresql-common/README.Debian.gz`).
```
apt install -y postgresql postgresql-contrib
service postgresql start
sudo -u postgres createuser -DRS travis
sudo -u postgres createdb -O travis mal
```
Then the `travis` user can `psql -dmal`, then `\cd` to the directory and `\i` the .sql files.

PR #711, opened 2025-04-09
Add Hare implementation

einsiedlerspiel:
mal implemented in pure hare https://harelang.org/ using only its standard library

~~There's no GC, I try to clean up any  non persistent allocations but really for the most part I'm just leaking all the memory.~~

~~Metadata implementation is incomplete and only implemented for user defined functions and hashmaps, for everything else I used native types, adding meta data would probably require completely reworking that.~~

~~There's lots of things to clean up and add checks for, but I won't get to that any time soon so I thought I'll just send the PR now, before I'll never do it.~~

Had lot's of fun with the challenge, thanks for creating it!

Pull request requirements:

- [x] Commits are well written and well organized.
- [x] Commits for a specific implementation should be prefixed with
  the implementation name.
- [x] Github Actions CI passes all checks (including self-host)

Additional requirements if you are adding a new implementation (see [FAQ](../docs/FAQ.md#add_implementation) for details):

- [x] Follow incremental structure (no common eval code)
- [x] Add `impls/<IMPL>/Dockerfile`
- [x] Add `impls/<IMPL>/Makefile`
- [x] Update `IMPLS.yml`
- [x] Update `Makefile.impls`
- [x] Update `README.md`


einsiedlerspiel:
Possible remaining TODOs: 

- GC during evaluation
- some performance tweaks

the big slowdown (in the perf test)  is not the GC itself (doesn't kick in during script evaluation rn), but that I'm now keeping track of allocated symbols in a dynamic array instead of a hashmap, commenting out the iteration in `make_symbol` gives almost all of the previous performance back.

kanaka:
Looks good. Merged! I've updated the README counts. Hare bumps the implementation language count to 89. 🎉

PR #717, opened 2025-07-21
c.2: various improvememts, fix new tests

asarhaddon:
* Define MalType as a pointer type, instead of an alias for a struct. It is now possible to declare it constant.
* Use macros to avoid lots of code duplication.
* Avoid redundant tests when possible.
* Avoid implicit signed/unsigned conversions.
* Use the const modifier whenever possible.
* Move private forward declarations from .h to .c, and forward declarations on the top of the .c.

Makefile:
  enable more warnings avoid recompiling each .o for each step fix .PHONY usage

core.c:
  Link mal_list/vector/hash_map directly to make_*. Move mal_readline from stepA (reallocate with GC).
  + - * /: exactly two arguments. =: without memory allocation, stop comparing functions and closures. Stop assuming that nil contains a valid list(first/rest/get). throw: the argument should never be an exception. vec: refuse maps. seq: fix allocation of single character-strings.

core.h:
  Represent the namespace as a static array instead of a linked list.

list:
* Avoid reversals, either by modifying the last link or simply by inserting elements in the right order,
* Avoid void*, the benefit of genericity here is small compared to the benefits of type checking.
* Store a pointer to nil as default metadata, instead of NULL.

env:
* Implement with binary trees.  This is easy for a mutable structure, and helps a lot the performance of accessing REPL keys.
* move repeated calls to env_set to steps.

maps:
* Check type of map keys.
* Remove the confusing in-place mutations.

printer:
  Replace most string allocations with idiomatic printf-style specifiers.

reader:
  Rewrite without intermediate tokenizer. Propagate more errors. Merge list/vector/hasmap handling.

run:
  does not need bash.

steps:
 * reduce diff between steps.
 * Add an env_apply helper, reducing duplication.
 * Merge regularise_parameters into eval_fnstar (propagating errors was a burden compared to the benefit).
 * Use the same function for eval_vector, eval_hashmap and apply arguments. Return an error directly, not a list containing an error.
 * Rewrite quasiquote, fixing some misplaced unquotes.
 * Clarify return value of TCO helpers.
 * Report errors during startup.

types:
* Try to merge type selection and value extraction, for concision and safety.
* In the union, use one name per type (mal_string/keyword/symbol were confused in several places anyway).
* Represent atoms as a double pointer, and macros with a type.
* Prefer enums to macros.
* Represent function parameters as a static array instead of a linked list.

### DIFF
--- a/impls/c.2/Makefile
+++ b/impls/c.2/Makefile
@@ -1,46 +1,10 @@
 CC = gcc
 
-CFLAGS = -std=c99 -g -Wall
+# Optimization is not required but enables some more warnings.
+CFLAGS = -std=c99 -g -Wall -Wextra -O1
 
-LIBS = -ledit -lgc
-FFI_LIBS = -ldl -lffi
-
-SRC = reader.c printer.c types.c env.c core.c
-HEADERS = reader.h printer.h types.h env.h core.h
-
-LIB_DIR = ./libs
-LIB_LIST_H = $(LIB_DIR)/linked_list/linked_list.h
-LIB_LIST_SRC = $(LIB_DIR)/linked_list/linked_list.c
-
-LIB_MAP_H = $(LIB_DIR)/hashmap/hashmap.h
-LIB_MAP_SRC = $(LIB_DIR)/hashmap/hashmap.c
-
-LIBS_H = $(LIB_LIST_H) $(LIB_MAP_H)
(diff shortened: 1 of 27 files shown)


asarhaddon:
Reducing the allocations has no visible effect on performances. On the other hand, the change of implementation for the REPL environment multiplies the performance score by 4.

PR #723, opened 2025-08-08
c.2: hashmaps, vectors, hide MAL struct, reduce duplication

asarhaddon:
Fix a few issues in 62d1ed3c (it seems one cannot extend a merged PR).
Out of curiosity, are the performance tests results public?
On my machine, this is ten times faster than before 62d1ed3c.

As a side note, env_get searches for DEBUG-EVAL in all parent environments until REPL during each evaluation, so map search is now probably the limiting factor for speed in most implementations.  The guide should probably mention that binary trees, even without effort to keep them balanced, are in this case (one large structure and lots of very small ones) both faster and easyer to implement than the currently suggested  linear representation. The linear representation is only convenient until the guide requires `dissoc`.

Makefile:
  Compute the build flags with pkg-config.
  Make more settings available on the command line.
  Enable the -fanalyzer warning (and close the issue found in slurp).
core:
  Simplify the argument parsing and return value constructor in ".".
env:
  Implement with hashmaps. Allow printing by DEBUG-EVAL.
error (new):
  Format most error messages consistently.
  When an error is detected, show all arguments instead of only the
  remaining ones (this was not informative when the rest was empty).
  Restore the previous error prefix (the MAL function, not the C function).
  Use a global variables instead of a type, simplifying functions
  returning non-MAL types.
hashmap (rewritten, moved to ./):
  Replace binary trees with hashed maps.
linked_list (moved to ./):
  Return a signed count in order to spare a later conversion.
printer:
  Allow a width specifier (for DEBUG-EVAL).
  Print MAL type sets (for bad_type) and hashmaps (for DEBUG-EVAL).
  Shorten name for functions (else REPL fills the screen).
  Move commands outside of assert (they didn’t survive NDEBUG).
  Inverse the meaning of " " so that list *are* spaced by default.
readline (new):
  Avoid duplication in steps and core (core was incorrect).
steps:
  Properly report reader errors (this was dealt with by the null case in EVAL,
  but not documented and removed by 62d1ed3c).
  Stop preventing TCO when if returns nil (the benefit is dubious).
  Add a quasiquote_folder used by lists and vectors.
  Stop calling quasiquote twice on each list/vector element (error in 62d1ed3c).
types:
  Hide the struct into the .c file.
  Compute the hash of each string at creation.
  Allocate common symbols once at startup.
  Remove metadata from most types.
  Use the same "pattern-matching" function to test the type and extract the contents.
  Add metadata as an optional argument to some constructor.
vectors (new):
  Provide an iterator common with lists.

global
  Move equality from core to types, use it in steps and hashmaps.

  In closures, store the raw fn* arguments (The &more syntax without
  space is lost but could be restored if this seems important).


kanaka:
@asarhaddon Nice!

The numbers are published here: https://kanaka.github.io/mal/graph/
Caveat: it's really out of date data. I need to integrate updating that data into the Github Actions workflow. The data is extracted from Github Actions results, but it's probably not comparable any more (because the underlying VMs have likely changed substantially since then).

PR #725, opened 2025-08-18
rust: various improvements

asarhaddon:
The second commit contains the actual changes and describes them in detail.
I have also reduced the diff between step files (as far as rustfmt allows).


kanaka:
Love to see this: 
image

PR #730, opened 2025-10-19
Use /usr/bin/env bash shebangs instead of /bin/bash

Dondorio:
This has better compatibility and lets you run the impls on nixos


kanaka:
Thanks @Dondorio!

PR #732, opened 2026-02-12
Add another elixir implementation

chrooti:
I've done a second elixir implementation and I thought about submitting it. It is incrementally better than the existing solution IMO but not by a lot. I'll leave you to judge whether this is merge worthy or not (at least I had fun!).

Differences:
- doesn't use Agents or external processes to keep the env state, but uses a fully functional approach (okay, except for exceptions)
- in some places (e.g. the quasiquote code) my implementatiion is arguably cleaner
- doesn't use escripts (not a pro or a con, although there is no big, elixir-embedding executable built when you run this, it just runs off the already compiled BEAM bytecode files)
- fully type checked (as much as dialyzer allows, but typing annotations help humans too)

Cons:
- leaks memory due to the way I implemented nested envs, a separate GC pass is needed which I haven't implemented.

I'm not sure if the single commit format is good or not, I'm looking for feedback on that.

Anyway, thanks for the challenge, it was not my first approach with lisp, but if it were it would have been much more fun :)

Pull request requirements:

- [x] Commits are well written and well organized.
- [x] Commits for a specific implementation should be prefixed with
  the implementation name.
- [x] Github Actions CI passes all checks (including self-host)

Additional requirements if you are adding a new implementation (see [FAQ](../docs/FAQ.md#add_implementation) for details):

- [x] Follow incremental structure (no common eval code)
- [x] Add `impls/<IMPL>/Dockerfile`
- [x] Add `impls/<IMPL>/Makefile`
- [x] Update `IMPLS.yml`
- [x] Update `Makefile.impls`
- [x] Update `README.md`

### DIFF
--- a/IMPLS.yml
+++ b/IMPLS.yml
@@ -22,6 +22,7 @@ IMPL:
   - {IMPL: dart}
   - {IMPL: elisp}
   - {IMPL: elixir}
+  - {IMPL: elixir.2}
   - {IMPL: elm}
   - {IMPL: erlang, NO_SELF_HOST: 1}  # step4 silent exit on "(DO 3)"
   - {IMPL: es6}
--- a/Makefile.impls
+++ b/Makefile.impls
@@ -33,7 +33,7 @@ wasm_MODE = wasmtime
 #
 
 IMPLS = ada ada.2 awk bash basic bbc-basic c c.2 chuck clojure coffee common-lisp cpp crystal cs d dart \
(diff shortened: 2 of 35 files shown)


kanaka:
@chrooti Thanks for the alternate implementation. I just started a new job so I haven't been able to review this. I'll try and review in the next few days.

chrooti:
best of luck for your new job then!

Issue #709, opened 2025-02-26
Memory leaks in Rust implementation


OgGhostJelly:
Circular references using `Rc` in Rust causes memory leaks! An environment can reference a function and a function can reference an environment, creating a circular reference. Possible fixes for the issue could be:
- Cloning the outer environment when a function is created instead of keeping a reference to it, this could causes issues with multi-threading e.g if the environment gets mutated the functions own copy of the environment won't be changed. 
- Using a `Weak` reference instead of `Rc`, this could cause complications with function currying and referencing environments that have been dropped.
- Implementing a garbage collector.

I have already added the cloning outer reference method into [my own mal rust implementation](https://github.com/OgGhostJelly/mal).

Or since the impls are mainly there for demonstration purposes it might not be worth the added complexity of fixing this issue.

asarhaddon:
Cloning the outer environment does not implement the desired behaviour, see #677.

Weak references could maybe fix variants of `(let* [f (fn* [] nil)] nil)` but probably not variants of `(let* [a (atom nil) l (list a)]  (reset! a l))`.

asarhaddon:
Merge request #742 fixes this issue for rust.
The guide should definitively warn implementors against these two non-solutions #739.

PR #735, opened 2026-04-23
Simplify the run scripts

asarhaddon:
Hello.

The idea was to use /bin/sh where possible. On Debian/Ubuntu, this is way faster than bash. Also, /usr/bin/env foo is not more portable than /usr/bin/foo, it just fails elsewhere.
A number of unrelated and small suggestions germed on the way. I can make distinct merge requests for the want you find useful and cannot cherrypick.
Rust fails, probably because of an outdated Docker image. The build succeeds on my machine.
Qbasic fails, although the changes in the directory are minimal, and safe for cbmbasic.
I failed to understand the issue.
Perl failed with a cryptic error, fixed by adding a --mal command line option disabling readline.
Purs fails with the same error than Perl.  I have given up applying the same fix. Any change I introduce in src/Readline.js results in a new error.
Please tell me which changes are woth working on.  I would also appreciate help about qbasic and purs.


asarhaddon:
Rust fails because of an old docker image.
Plpgsql fails or suceeds randomly, so the issue is most probably unrelated with this merge request.
Changes in Makefiles have moved to #738.
Qbasic and purs require bash, I do not know why.  The evident work-around is to keep bash for them.

asarhaddon:
The rust failure is fixed by #709.

PR #738, opened 2026-07-20
Add steps.mk and simplify some implementation Makefiles

asarhaddon:
Hello.
This merge request suggests a `steps.mk` helper for the implementation Makefiles. Some parts of #735 require a rebase when/if the automated tests pass.

Here is the rest of the commit message:

In the process, recommend steps.mk for compiled language, and a minimal Makefile for interpreted ones.

steps.mk avoids the duplication of the step names in each Makefile. It provides a consistent and explicit naming scheme.
BINS             ->  step0A:=.exe
OBJS             ->  step0A:%=obj/.o
EARLY_STEPS_SRC  ->  step03:=.js
LATE_STEPS       ->  step4A

It allows some patterns to carry an explicit list of desired matches. step%.o: step%.c  ->  (step0A:=.o): %.o: %.c

Unrelated changes.

Replace several obsolete (lastword (steps)) constructs with (stepA).

ada: replace -p mkdir option with a Make order prerequisite
ada: replace force with .PHONY
ada: use -o compile option instea of renaming the object
erlang: replace a template macro with a loop on targets
go: replace most of a template with a loop on targets
hare: makefile -> Makefile
haxe, scheme, sml: group the recipes by _MODE for readability


asarhaddon:
The rust failure is fixed by #742 
Plpgsql complains, although it is not affected by this commit, and the test was passing in the previous forced push.

PR #742, opened 2026-08-27
Fix memory leaks in objpascal and rust

asarhaddon:
Hello.
These commits fix some memory leaks detected with valgrind.


asarhaddon:
rust requires the latest computer (with or without this change). The Docker image is already Ubuntu 25.04.
Objpascal is a mystery.  Both readline and gc changes, independently, pass normal and self-hosted tests but crash the three perf tests. Any clue is welcome.

PR #743, opened 2026-08-28
update github actions because github is deprecating node20

asarhaddon:


asarhaddon:
The rust failure is fixed by #742.

